PR #21, opened 2021-05-21
Fix offline sessions

sailorgallifrey:
createOfflineSessionsImpl was not creating offline sessions, but rather was creating only active sessions. Instances of `UserSessionModel` and `AuthenticatedClientSessionModel` need to be converted to offline using `createOfflineUserSession` and `createOfflineClientSession` respectively. Tested and confirmed that offline sessions are now being created when this is deployed to our Keycloak instance.


pedroigor:
@sailorgallifrey Thanks. I think the original intent was to only persist sessions. Mainly for testing loading offline sessions during startup.

I guess you are also using these sessions to test authentication?

sailorgallifrey:
@pedroigor This is trying to match expected behavior. The behavior we experienced is that originally only active sessions were created. This change ensures that offline sessions are also being created and persisted.

We were using this to test loading offline sessions. Let me know if you have further questions.

pedroigor:
@sailorgallifrey Thanks.

PR #24, opened 2021-09-01
User crawl scenario

benjamin37:
Hi,

created a new scenario to fetch users page by page. Numbers of users to fetch per page and number of pages to iterate over can be configured.

Kind regards
Benjamin


pedroigor (approved):
@benjamin37 Thanks!

PR #30, opened 2021-10-28
Remove nested transactions

benjamin37:
Hi,

we noticed that the dataset SPI uses some kind of nested transactions here and there (Nesting `KeycloakModelUtils.runJobInTransactionWithTimeout()` method calls). Not sure why this is the case. Sometimes when creating a big set of data we got some Exceptions related to transactions. These exceptions went away when changing to flat transactions (remove the nesting of `KeycloakModelUtils.runJobInTransactionWithTimeout()` method calls). Also it simplifies the code.

Do you know why there is nesting of this method in place? I did a few tests what happens when the outer or inner transaction fails but couldn't make sense of it. It seems that only the innermost transaction is working (or at least performs rollbacks).

Kind regards
Benjamin

### DIFF
--- a/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
+++ b/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
@@ -31,7 +31,6 @@ import static org.keycloak.benchmark.dataset.config.DatasetOperation.REMOVE_REAL
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


pedroigor:
@mposolda Does it make sense for you?

pedroigor (approved):

pedroigor:
@benjamin37 Thanks. @mposolda Do you have any reasoning for keeping the current transaction model?

mposolda (approved):
@benjamin37 @pedroigor Thanks for the PR! No specific reason for nested transactions. I think the only reason was that we did this a bit in a hurry and hence maybe transaction management was not done in 100% optimal way :-) Thanks for improving it!

PR #32, opened 2022-01-05
Update dependencies to run with 16.1.0

tolleiv:
Hi,
using "16.0.0-SNAPSHOT" doesn't work anymore and bumping the versions to "16.1.0" resolves the issues so here's the required adjustment for that.
Cheers


pedroigor (approved):

pedroigor:
@tolleiv Tks ! Will update to 17 now :)

PR #35, opened 2022-01-26
Reduce memory footprint when creating many users and clients.

### LINKED ISSUE #37: Reduce memory footprint when creating many users and clients
### Description

Currently the dataset spi consumes a lot of memory when several thousand users and clients are created.

### Discussion

_No response_

### Motivation

High memory consumption of dataset spi has negative impact of performance when creating lots of test data.

### Details

A look in the sourcecode shows that users and clients are unnecessarily kept in memory, just to be able to log their count later on.

ioemat:
We realized that the dataset spi consumes a huge amount of memory when many users and clients are created.
This may change the behaviour of the running keycloak instance and has also a negative  impact on performance while creating test data.
Since the lists are currently only used to get the item count I replaced them with a counters.

Closes #37


ioemat:
@pedroigor any thoughts on this?

pedroigor (approved):
@ioemat Nice one ... Indeed, better.

PR #44, opened 2022-03-30
Replaced `maxFailedRequests` SLA assertion with `maxErrorPercentage`.

tkyjovsk:
Replacing assertion for maximum number of failed requests with maximum request error percentage.


tkyjovsk:
Note: This breaks backwards compatibility because the `sla-failed-requests` parameter is no longer available.

pedroigor (approved):

PR #50, opened 2022-04-29
Date header is not delivered anymore in Keycloak responses and must not be used to compute access token generation time (fixes #46)

### LINKED ISSUE #46: Benchmark - CreateDeleteClients and UserCrawl scenarios issue
### Describe the bug

I'am following instruction for Keycloak Benchmark scenarios and there is a problem with `keycloak.scenario.admin.CreateDeleteClients` and `keycloak.scenario.admin.UserCrawl`. Those scenarios failed on `Get service account token` step with error `header(Date).find.exists, found nothing`

I found that issue is probably in this part of code https://github.com/keycloak/keycloak-benchmark/blob/c28b4693f73af399bae7df4fcc472dd29ab0c6ce/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala#L296

### Version

17.0.0 (Quarkus)

### Expected behavior

Scenario finished without errors.

### Actual behavior

Scenario break

### How to Reproduce?

Run Keycloak Benchmark with scenario `keycloak.scenario.admin.CreateDeleteClients` and `keycloak.scenario.admin.UserCrawl`

### Anything else?

_No response_

davoustp:
As described in #46 , the `Date`header is not delivered anymore by Quarkus-based Keycloak.

This header was used in the scenario to compute the access token generation timestamp.

This PR computes this timestamp differently: it first stores the current timestamp in the session *before* the token request is performed, and *if successful*, then it stores the appropriate session attribute with the previously computed timestamp.

Benchmark scenarii now work as expected again - with both previous Keycloak versions and newer Quarkus-based versions.

Closes #46

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -285,19 +285,20 @@ class KeycloakScenarioBuilder {
   }
 
   private def getServiceAccountTokenExec(): ChainBuilder = {
-    exec(http("Get service account token")
+    // Store the current timestamp BEFORE it's actually sent (as we don't know how long the request will take or if it will succeed at all)
+    exec(_.set("requestEpoch", System.currentTimeMillis()))
+    .exec(http("Get service account token")
       .post(TOKEN_ENDPOINT)
       .formParam("grant_type", "client_credentials")
       .formParam("client_id", "gatling")
       .formParam("client_secret", "${clientSecret}")
       .check(
         jsonPath("$..access_token").find.saveAs("token"),
-        jsonPath("$..expires_in").find.saveAs("expiresIn"),
-        header("Date").saveAs("tokenTime")
+        jsonPath("$..expires_in").find.saveAs("expiresIn")
       ))
       .exitHereIfFailed
       .exec(s => {
-        s.set("accessTokenRefreshTime", ZonedDateTime.parse(s("tokenTime").as[String], DATE_FMT).toEpochSecond * 1000)
+        s.set("accessTokenRefreshTime", s.attributes("requestEpoch"))
       })
   }
 
--- a/benchmark/src/main/scala/keycloak/scenario/_private/AdminConsoleScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/_private/AdminConsoleScenarioBuilder.scala
@@ -105,6 +105,8 @@ class AdminConsoleScenarioBuilder {
       // TODO store AUTH_SESSION_ID cookie for use with oauth.authorize?
       .exitHereIfFailed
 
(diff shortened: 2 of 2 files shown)


benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -285,19 +285,20 @@ class KeycloakScenarioBuilder {
...
       .check(
         jsonPath("$..access_token").find.saveAs("token"),
-        jsonPath("$..expires_in").find.saveAs("expiresIn"),
-        header("Date").saveAs("tokenTime")
+        jsonPath("$..expires_in").find.saveAs("expiresIn")
       ))
       .exitHereIfFailed
       .exec(s => {
-        s.set("accessTokenRefreshTime", ZonedDateTime.parse(s("tokenTime").as[String], DATE_FMT).toEpochSecond * 1000)
+        s.set("accessTokenRefreshTime", "${requestEpoch}")
```
ahus1, line 301: I like the idea to capture the time before sending the request as an attribute.

I ran the Scenario against a Quarkus and Wildfly distribution (SNAPSHOT) of Keycloak. The login works, but then an exception "j.l.NumberFormatException: For input string: "${requestEpoch}"" appears in the log.

A assume this is due that s.set() doesn't replace the attribute as the code change above assumes. 

What works instead in my local tests is fetching the attribute without the replacement. Please give that one a try and update this PR. 

```suggestion
        s.set("accessTokenRefreshTime", s.attributes.get("requestEpoch"))
```

Please apply this also in the other similar places.

ahus1 (changes requested):
Thank you for this PR! I've suggested a change below. 

Once you've added the changes, please follow the usual conventions of the main Keycloak project and squash all your changes to a single commit.

It would also be great if you could mention `Closes #46` in the text summary of this pull request. This will help collaboration on GitHub in two ways:

1. It populates the right-hand column on both the issue and the PR with a reference to the issue you're about to fix.
2. It will close the issue once the PR has been merged automatically.

Thanks!

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
davoustp: All done, rebased and pushed.
Thanks for catching the error (obviously, pushing a `String` where a `long` is expected does not play nice - weird that I did not see the error in the first place, though).

ahus1: Thank you! You can now mark this as resolved.

ahus1:
@davoustp - I see that you accidentally pushed changes to this branch, and then reverted the push. 

Thanks for fixing this so soon; this comment is there for Pedro that this is still good to be merged.

davoustp:
> I see that you accidentally pushed changes to this branch, and then reverted the push.

Exactly, so sorry for the accidental commit, I messed up with my branches (monday morning slow start, I guess) so had to force push to keep this one clean.

PR #52, opened 2022-04-29
Fix parameter passed to list clients endpoint

### LINKED ISSUE #51: CreateClients scenario with increasing latency for listClients
### Describe the bug

When the test CreateClients runs for a while adding clients, it will at some point have exponential duration. 

This is due the wrong query parameter Keycloak will return the full list of client all the time.

### Version

0.7-SNAPSHOT

### Expected behavior

Constant time for list() action for client

### Actual behavior

Increasing time for list() action for clients

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
This fixes the behavior observed in the test. 

This now uses the correct parameter, as seen in [ClientsResouce](https://github.com/keycloak/keycloak/blob/7335abaf08915d364f16667733237aa8ac8a3b0f/services/src/main/java/org/keycloak/services/resources/admin/ClientsResource.java#L109): 

```
    public Stream<ClientRepresentation> getClients(@QueryParam("clientId") String clientId,
                                                 @QueryParam("viewableOnly") @DefaultValue("false") boolean viewableOnly,
                                                 @QueryParam("search") @DefaultValue("false") boolean search,
                                                 @QueryParam("q") String searchQuery,
                                                 @QueryParam("first") Integer firstResult,
                                                 @QueryParam("max") Integer maxResults) {
```

Closes #51

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -320,7 +320,7 @@ class KeycloakScenarioBuilder {
       .exec(http("List clients")
         .get(ADMIN_ENDPOINT + "/clients")
         .header("Authorization", "Bearer ${token}")
-        .queryParam("maxResults", 2)
+        .queryParam("max", 2)
         .check(status.is(200)))
       .exitHereIfFailed
     this


ahus1:
Hi @kami619 - this should fix the problem with the increasing list() times. 

Could you please review it an comment/approve here once you have the time? Thanks!

kami619 (approved):
lgtm

ahus1:
@pedroigor - as @kami619 has reviewed it, could you please merge it? Thanks!

kami619:
the changes works as expected on my local. the response times are now constant aligning to the max parameter value specified in the test for listing the clients

PR #55, opened 2022-05-02
Configure HTTP logging via script arguments instead of logback.xml changes (fixes #54)

### LINKED ISSUE #54: Configure HTTP logging via script arguments instead of logback.xml changes
### Description

Enabling HTTP logging (whether on failure or in any case) currently requires to modify the logback configuration and recompile the module.

This should be done instead using a script argument when running the benchmark.

### Discussion

_No response_

### Motivation

HTTP logging is a very helpful facility to ease troubleshooting and/or scenario development (read: it facilitates writing new scenarii ;-) ).

It should therefore be quite easy to setup, hence the proposal to use of an option instead of configuration file changes.

### Details

This can be easily done using the logback conditional configuration capability as described at https://logback.qos.ch/manual/configuration.html#conditional .

This requires adding the `janino` library to the dependencies, and modify the logback configuration file to set the proper logger level depending on the presence of the appropriate Java system property (which is set by the `kcb.sh` script).

_Note:_
gatling 3.4.0 suffers from an issue in this area, see https://github.com/gatling/gatling/issues/3994 : HTTP requests are incorrectly logged.
Upgrading gatling to 3.4.2 fixes the problem.

davoustp:
* Upgraded `gatling` 3.4.0 -> 3.4.2 to include changes for https://github.com/gatling/gatling/issues/3994
* Added `janino` to enable logback conditional configuration
* Updated logback configuration file to set proper logger levels depending on Java properties
* Added a logging section into README

Closes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,12 @@ To use a different server URL and scenario:
 You can set different options to change how tests should be executed. For a complete list of the available options, see
 [Config](benchmark/src/main/java/org/keycloak/benchmark/Config.java).
 
+### Logging
+
+You can enable HTTP logging using the following additional options:
+* `--log-http-on-failure`: HTTP requests and responses will be logged when the scenario expectations are not met
+* `--log-http-always`: HTTP requests and responses will *always* be logged
+
 ### Report
 
 Check reports at the `result` directory.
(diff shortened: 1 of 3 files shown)


ahus1 (approved):
Thanks for making this configurable, and for updating the dependency here. 

Adding the new dependency is fine here as this is only a benchmark project that will not be part of the main Keycloak distribution.

PR #57, opened 2022-05-02
Add a CreateRealms scenario to measure realm creation scalability (fixes #56)

### LINKED ISSUE #56: Add a CreateRealms scenario to measure realm creation scalability
### Description

A scenario creating realms in a freshly deployed Keycloak is required to measure its scalability against the number of realms.

Similarly to the `CreateClients` scenario, this scenario will:
* acquire the service account token
* create a realm with globally unique name and id

Concurrent realm creation (using more than one user) must also be possible without collisions.

### Discussion

https://github.com/keycloak/keycloak/discussions/11074

### Motivation

A new scenario is required to measure the response time and scalability of Keycloak with a growing number of realms.

### Details

_No response_

davoustp:
A new scenario is added to perform and measure realm creation performance with Keycloak.

Command-line example for running this scenario:
```
./kcb.sh --scenario=keycloak.scenario.admin.CreateRealms \
  --server-url=http://localhost:8080 \
  --concurrent-users=1 \
  --measurement=300 \
  --realm-name=master \
  --client-secret=<secret>
```

Closes #56 

*Implementation note*

Ensuring realm name and id uniqueness is achieved by adding a property containing a unique ID for each user created to execute the scenario by gatling (using the `Feeder` facility), and using this property into the realm name and id.

This Gatling user property (available from the user Session), named `gatlingUniqueUserId`, can be used elsewhere as well as this is quite generic.

*Documentation note*

When using the document process to create a Keycloak Client for `gatling`, the client is created with flag `Full Scope Allowed` set to `On`, which actually lists all the access to all the realms into the access token, leading to the token becoming too large after roughly 40 created realms (receiving `431 Request Header Fields Too Large` responses in this case).

Instead, this flag must be disabled and appropriate scopes should be set.
Documentation has been updated, but it requires to review the scopes used by the other scenarii - a TODO flag is left in the doc as the placeholder for this list to be discussed in this PR and amended before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Before running tests, make sure realms are configured as follows:
 
 * Realms must have `User Registration` setting enabled.
 
-Some scenarios (`CreateDeleteClients` and `CrawlUsers`) require a service account with the clientId `gatling`:
+Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateRealms`) require a service account with the clientId `gatling`:
 
 * select the realm that is used for testing
 * create a client  with the name `gatling`
@@ -188,6 +188,21 @@ Some scenarios (`CreateDeleteClients` and `CrawlUsers`) require a service accoun
    * select for `realm-management` in the `Client Roles` listbox
    * assign the roles `manage-clients` and `view-users`
 * the client secret to be passed to the tests can be copied from the `Credentials` tab
+
+#### Scenario `keycloak.scenario.admin.CreateRealms`
+
+This scenario requires the following Client settings, in addition to the requirements above:
+
+* In to the `Scopes` tab
+    * Set `Full Scope Allowed` to `Off`
+    * Add role `create-realm` in the `Realm Roles` section
+* In the `Service Account Roles` tab
+    * assign the roles `create-realm` in the `Realm Roles` section
+
+The `Full Scope Allowed` flag is especially important to disable, because if enabled,
+the access token will contain *all* the accesses to all the realms,
+leading to the token becoming too large after roughly 40 created realms
+(receiving `431 Request Header Fields Too Large` responses in this case).
 
 ### Run
 
@@ -219,6 +234,7 @@ These are the available test scenarios:
 * `keycloak.scenario.authentication.ClientSecret`: Client Secret (Client Credentials Grant)
 * `keycloak.scenario.admin.CreateDeleteClients`: Create and deleted clients (requires `--client-secret=<client secret for gatling client>`)
 * `keycloak.scenario.admin.UserCrawl`: Crawls all users page by page (requires `--client-secret=<client secret for gatling client>`)
+* `keycloak.scenario.admin.CreateRealms`: Create realms (requires `--client-secret=<client secret for gatling client>` and `--realm-name=master`)
 
 ## Release
 
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -7,7 +7,7 @@ import io.gatling.core.structure.ChainBuilder
 import io.gatling.http.Predef._
 import keycloak.Utils
 import keycloak.Utils.randomUUID
-import keycloak.scenario.KeycloakScenarioBuilder.{ADMIN_ENDPOINT, CODE_PATTERN, LOGIN_ENDPOINT, LOGOUT_ENDPOINT, TOKEN_ENDPOINT, UI_HEADERS, downCounterAboveZero}
+import keycloak.scenario.KeycloakScenarioBuilder.{ADMIN_ENDPOINT, CODE_PATTERN, LOGIN_ENDPOINT, LOGOUT_ENDPOINT, TOKEN_ENDPOINT, UI_HEADERS, REALMS_ENDPOINT, downCounterAboveZero}
 import keycloak.scenario._private.AdminConsoleScenarioBuilder.DATE_FMT
 import org.keycloak.benchmark.Config
 
@@ -25,6 +25,7 @@ object KeycloakScenarioBuilder {
   val LOGIN_ENDPOINT = BASE_URL + "/protocol/openid-connect/auth"
   val LOGOUT_ENDPOINT = BASE_URL + "/protocol/openid-connect/logout"
   val TOKEN_ENDPOINT = BASE_URL + "/protocol/openid-connect/token"
+  val REALMS_ENDPOINT = "${keycloakServer}/admin/realms"
   val ADMIN_ENDPOINT = "${keycloakServer}/admin/realms/${realm}"
   val CODE_PATTERN = "code="
 
@@ -302,6 +303,19 @@ class KeycloakScenarioBuilder {
       })
   }
 
+  def createRealm(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(_.set("createdRealmId", randomUUID()))
+      .exec(http("Create realm")
+        .post(REALMS_ENDPOINT)
+        .header("Authorization", "Bearer ${token}")
+        .header("Content-Type", "application/json")
+        .body(StringBody("""{"id":"${createdRealmId}","realm":"${createdRealmId}","enabled": true}"""))
+        .check(status.is(201))
+        .check(header("Location").notNull.saveAs("location")))
+      .exitHereIfFailed
+    this
+  }
 
   def createClient(): KeycloakScenarioBuilder = {
     chainBuilder = chainBuilder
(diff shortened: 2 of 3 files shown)


davoustp:
@ahus1 please tell me if this is what you expected from the related discussion.

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
```diff
@@ -36,7 +38,12 @@ abstract class CommonSimulation extends Simulation {
   }
 
   def setUp(name: String, builder : KeycloakScenarioBuilder): Unit = {
-    val scn = scenario(name).exec(builder.chainBuilder)
+    // Add an attribute which uniquely identifies the Gatling user
+    val gatlingUniqueUserId = new AtomicLong(0)
```
ahus1, line 42: To be able to run Gatling multiple times against the same DB, an ID that would be unique across multiple runs would be helpful.

On another note: if the realm IDs generated here would actually be UUIDs there would be no need for such a state, and even multiple Gatling instances could run in parallel. 

```suggestion
    // Assuming less than 1000 users per second, this will be unique across multiple runs.
    val gatlingUniqueUserId = new AtomicLong(System.currentTimeMillis())
```

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -302,6 +312,19 @@ class KeycloakScenarioBuilder {
       })
   }
 
+  def createRealm(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("Create realm")
+        .post(REALMS_ENDPOINT)
+        .header("Authorization", "Bearer ${token}")
+        .header("Content-Type", "application/json")
+        .body(StringBody("{\"id\":\"realm-${gatlingUniqueUserId}\",\"realm\":\"realm-${gatlingUniqueUserId}\",\"enabled\": true}"))
```
ahus1, line 321: As noted above: instead of `gatlingUniqueUserId` a UUID could be used here.

ahus1 (changes requested):
Thank you for this PR, this works great and provides the needed results. 

Please see below for my comments about improvements. 

My general thoughts would be to use a UUID instead of a counter in the name of the realm, as it might simplify things (no shared state between scenarios), would allow to run multiple in parallel (to create a higher load), and have multiple runs against the same database one after the other. Feel free to convince me of any downsides of this approach.

As the CreateRealms creates new realms all the time and the time to create realms increases with every run, it might be beneficial to have a CreateDeleteRealms scenario to test with a constant number of realms. As I didn't ask for this in the GitHub discussion leading to this, I'll be fine if you decide that this is outside the scope of this PR, and that we'll create a PR for this once it is needed.

I'm looking forward to your comments and an updated pull request!

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -46,12 +47,20 @@ object KeycloakScenarioBuilder {
...
-    missCounter.getAndDecrement() > 0
+  }
+
+  def downCounterAboveZero(session: Session, attrName: String): Validation[Boolean] = {
+    getCounter(session, attrName).getAndDecrement() > 0
+  }
+
+  def incrementCounter(session: Session, attrName: String): Validation[Session] = {
+    getCounter(session, attrName).incrementAndGet()
+    session
```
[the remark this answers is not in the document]
davoustp: Again, correct - this was used in a prior attempt to increment the `nextAvailableRealmIndex` above, which is now useless.
Removing this change.

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
davoustp: I considered using a UUID instead, but was feeling it might get difficult to handle if investigating, especially with the UI.
However, I do believe that's better to use a UUID - it gracefully handles the case where the scenario is run multiple times against the same deployment (currently, any subsequent run will fail because will attempt to create realms with already existing id, leading to a conflict).

Do you think I should use the already existing UUID named `state` ?
Not sure about the semantics of this one, so it might be better to add a new UUID for this specific purpose.

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
davoustp: Yep, see my previous comment

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
ahus1: I had a look and I see that the UUID `state` is used for the OpenID connect login. Therefore that wouldn't fit here, IMHO. 

I suggest you'd add a UUID as part of `createRealm()` (similar to your latest change of "requestEpoch"), although this should then be named specific to the this step like 
`createdRealmId`, so that it can later be used to delete that realm.

davoustp: Ok, makes sense. Will do.

README.md
```diff
@@ -180,13 +180,18 @@ Some scenarios (`CreateDeleteClients` and `CrawlUsers`) require a service accoun
...
+    * click save
+* Change to the `Scopes` tab
+    * Set `Full Scope Allowed` to `Off`
+    * Add role `create-realm` in the `Realm Roles` section
+    * TODO: add required client roles to scopes as well!
+* Change to the tab `Service Account Roles`
+    * assign the roles `create-realm` in the `Realm Roles` section
+    * select for `realm-management` in the `Client Roles` listbox
+    * assign the roles `manage-clients`, and `view-users`
 * the client secret to be passed to the tests can be copied from the `Credentials` tab
```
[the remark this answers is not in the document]
davoustp: Ok, that makes a lot of sense. I'll review the doc change and add another dedicated paragraph for the scenario and its requirements.

davoustp:
> As the CreateRealms creates new realms all the time and the time to create realms increases with every run, it might be beneficial to have a CreateDeleteRealms scenario to test with a constant number of realms. As I didn't ask for this in the GitHub discussion leading to this, I'll be fine if you decide that this is outside the scope of this PR, and that we'll create a PR for this once it is needed.

Well, I think it could be valuable, but it would serve a slightly different purpose: here, the scenario measures how response time evolves as new realms are added to Keycloak. Creating then immediately removing realms would keep the overall number of realms stable overall, and hence we could not measure the scalability against the nb of realms.
This second scenario would be very interesting for profiling Keycloak while running it, for instance.

I can probably do this second scenario in a matter of minutes once we're done with this one. Probably better to separate it as another PR?

ahus1:
> Probably better to separate it as another PR?

Yes, let's create another PR for that once this one has been merged.

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
kami619: @davoustp in #321 you can use the below example to be able to avoid inserting escape chars. that increases the readability of the json body here. reference : https://gatling.io/docs/gatling/reference/current/http/request/?#stringbody

```// with a static payload
http("name").post("/")
  .body(StringBody("""{ "foo": "staticValue" }"""))

// with a Gatling EL string payload
http("name").post("/")
  .body(StringBody("""{ "foo": "#{dynamicValue}" }"""))

// with a function payload
http("name").post("/")
  .body(StringBody(session => s"""{ "foo": "${session("dynamicValueKey").as[String]}" }"""))```

File: README.md [hunk 5]
```diff
@@ -189,6 +189,21 @@ Some scenarios (`CreateDeleteClients` and `CrawlUsers`) require a service accoun
...
+This scenario requires the following Client settings, in addition to the requirements above:
+
+* In to the `Scopes` tab
+    * Set `Full Scope Allowed` to `Off`
+    * Add role `create-realm` in the `Realm Roles` section
+* In the `Service Account Roles` tab
+    * assign the roles `create-realm` in the `Realm Roles` section
+
+The `Full Scope Allowed` flag is especially important to disable, because if enabled,
+the access token will contain *all* all the accesses to all the realms,
```
ahus1, line 203: Remove a second "all" from the description.

```suggestion
the access token will contain *all* the accesses to all the realms,
```

ahus1 (changes requested):
Thank you for the updated PR. This now works as expected for me. 

Besides the review comment from Kamesh, there is one word too much in the README - that's all I found. Great work!

README.md
davoustp: Fixed. Thanks for catching this one!

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
davoustp: Hi @kami619 
Ah, I knew there was some syntactic sugar which I could use to avoid the escaping, but could not find it from the top of my head. ;-)
This is much better now.
Fixed.

kami619:
This looks good to me @davoustp great work and thanks for the quick turn around on this :)

kami619 (approved):

ahus1 (approved):
Looks great, read to merge!

ahus1:
@pedroigor - this has now been reviewed and working as expected. 

Please merge it. Thanks!

PR #59, opened 2022-05-04
addition of client scope and misc changes (fixes #58)

### LINKED ISSUE #58: Extend the create entities model to Client Scope entity
### Description

An enhancement request to extend the CreateClients, CreateRealms etc., model to Client Scope entity, to support the New Storage layer performance tests with CRUD apis at scale and load.

### Discussion

_No response_

### Motivation

As part of this, we are going to onboard the

- listing of client scopes
- addition of new client scopes following the similar pattern of creating realms and clients which are already part of the `src/main/scala/keycloak/scenario/admin` package.
- deletion of the newly added client scopes immediately after they are created
- delete the session variables from each rest api method, to keep the gatling sessions light weight

### Details

_No response_

kami619:
A new scenario is added to perform and measure client scopes creation performance with Keycloak.

Command-line example for running this scenario:

```
  ./kcb.sh --scenario=keycloak.scenario.admin.CreateDeleteClientScopes \
  --server-url=http://localhost:8080 \
  --concurrent-users=1 \
  --measurement=300 \
  --client-secret=<secret>
```

Closes https://github.com/keycloak/keycloak-benchmark/issues/58

##### Implementation note

Ensuring client scope name and id uniqueness is achieved by adding a property containing a UUID from the Utils.scala

We are also attempting to delete the session variables once their purpose is completed, from the gatling session wherever its possible, to keep the memory footprint of the gatling session low.

##### Documentation note
I have split the Benchmark section away from the README.md, which now primarily contains the information about Dataset module and Release information of the project.


benchmark/src/main/scala/keycloak/scenario/admin/CreateGroups.scala
```diff
@@ -0,0 +1,14 @@
...
+
+class CreateGroups extends CommonSimulation {
+
+  setUp("CRUD Groups", new KeycloakScenarioBuilder()
+    .serviceAccountToken()
+    //.listGroups()
+    //.createGroups()
+    //.deleteGroups()
+  )
+}
```
ahus1, line 14: IMHO this file should not be part of this PR (yet).

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -312,11 +313,53 @@ class KeycloakScenarioBuilder {
...
+      .exitHereIfFailed
+    this
+  }
+
+  def listClientScopes(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("List client Scopes")
+        .get(ADMIN_ENDPOINT + "/client-scopes")
+        .header("Authorization", "Bearer ${token}")
+        .queryParam("max", 2)
```
ahus1, line 345: As per API spec and per implementatoin, no query parameters are supported. It will always return the full list of clients. To avoid confusion for those reviewing the scenario, please remove this query parameter.

ahus1 (changes requested):
Thank you for preparing this PR. I like the renaming of the scenarios, and the new scenarios work as expected.

Please see two comments below. For this repo, I'd like to adopt the contributing policy of the Keycloak main repository that asks for a single commit per PR, and a commit message as follows:

```
A brief descriptive summary

Optionally, more details around how it was implemented

Closes #1234
```

This allows a rebase-on-merge strategy, with the commit pointing to the issue. For this issue I can also do a squash-on-merge and replace the commit message, but I would like to avoid it in the future. Please let me know how you want to proceed with this PR.

benchmark/src/main/scala/keycloak/scenario/admin/CreateGroups.scala
kami619: true, it slipped through my radar, will remove it

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
kami619: thanks for catching this, will remove this parameter

kami619:
@ahus1 if thats ok with you, I will make sure to follow the single commit based PR strategy for my future PRs, and lets squash and merge, this one.

ahus1 (approved):
Thank you for the additional changes, this now looks fine to me.

PR #61, opened 2022-05-04
Duplicate logback.xml and logback-test.xml configuration files leading to confusion (fixes #53)

### LINKED ISSUE #53: Duplicate logback.xml and logback-test.xml configuration files leading to confusion
### Describe the bug

The project bundles two almost identical `logback.xml` and `logback-test.xml` files into the same shaded jar.

As a result, logback always uses the `logback-test.xml` file - as this is rightfully the expected behaviour to manage test-related logging configuration - ignoring the other file in the process.

Not exactly sure how the test-focused `logback-test.xml` file is used in this context (maybe while debugging scala-scenarii?), but differences are minor (diff below):

```
12d11
<         <immediateFlush>false</immediateFlush>
17c16
<             <pattern>%d{HH:mm:ss} %msg%n%rEx</pattern>
---
>             <pattern>%msg%n%rEx</pattern>
38,41d36
<     <!--logging for dataset-->
<     <logger name="org.keycloak.benchmark.performance.dataset" level="INFO" additivity="false">
<         <appender-ref ref="CONSOLE_MSG_ONLY"/>
<     </logger>
```

We have a few options:
* merge them into the `logback.xml` file and remove the `logback-test.xml` file
* keep them but move the `logback-test.xml` file to `src/test/resources` as files in this directory are included in the classpath only for `test`-related maven scopes (this is where `logback-test.xml` should actually belong)

I'd suggest merging as having distinct files is always confusing for newcomers, especially that the differences are minor (pattern for `CONSOLE_MSG_ONLY` with or without timestamp, immediate flush on/off, and `org.keycloak.benchmark.performance.dataset`-related logging settings). 

Thoughts?
I'll submit a PR when agreement is reached.

### Version

18.0.0

### Expected behavior

Modifying the main `logback.xml` file should be taken into account when running the benchmark scenarii.

### Actual behavior

Modifying the main `logback.xml` file is entirely ignored when running the scenarii.

### How to Reproduce?

This can be demonstrated with start-time logging by adding option `--logback.statusListenerClass=ch.qos.logback.core.status.OnConsoleStatusListener` to the `kcb.sh` invocation:
```
10:15:54,704 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback-test.xml] at [jar:file:/Users/pdavoust/Documents/Temp/keycloak-tenant-benchmark/gatling/keycloak-benchmark-0.7-SNAPSHOT/lib/keycloak-benchmark-0.7-SNAPSHOT.jar!/logback-test.xml]
10:15:54,753 |-INFO in ch.qos.logback.core.joran.spi.ConfigurationWatchList@47ef968d - URL [jar:file:/Users/pdavoust/Documents/Temp/keycloak-tenant-benchmark/gatling/keycloak-benchmark-0.7-SNAPSHOT/lib/keycloak-benchmark-0.7-SNAPSHOT.jar!/logback-test.xml] is not of type file
```

This shows exactly which configuration file logback uses, as shown above.

### Anything else?

_No response_

davoustp:
Closes #53


ahus1 (approved):
Thank you for contributing this change and to analyze the strange behavior it caused.

PR #62, opened 2022-05-04
Add a CreateDeleteRealms scenario to measure realm CRUD performance and concurrency capability (fixes #60)

### LINKED ISSUE #60: Add a CreateDeleteRealms scenario to measure realm CRUD performance and concurrency capability
### Description

Similarly to #56 , a new scenario that creates and delete realms is required to investigate concurrency and also assist in profiling the use case for performance.

This scenario will:

* acquire the service account token
* create a realm with globally unique name and id
* delete the just created realm

### Discussion

https://github.com/keycloak/keycloak/discussions/11074

### Motivation

This new scenario is required to measure the concurrency behavior and performance of Keycloak with a growing number of realms.

### Details

_No response_

davoustp:
Scenario as described in #60

Closes #60

### DIFF
--- a/benchmark/BENCHMARK.md
+++ b/benchmark/BENCHMARK.md
@@ -25,7 +25,7 @@ Before running tests, make sure realms are configured as follows:
 
 * Realms must have `User Registration` setting enabled.
 
-Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateRealms`, `CreateDeleteClientScopes`) require a service account with the clientId `gatling`:
+Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateDeleteClientScopes`) require a service account with the clientId `gatling`:
 
 * select the realm that is used for testing
 * create a client  with the name `gatling`
@@ -38,20 +38,19 @@ Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateRealms`, `Create
    * assign the roles `manage-clients` and `view-users`
 * the client secret to be passed to the tests can be copied from the `Credentials` tab
 
-#### Scenario `keycloak.scenario.admin.CreateRealms`
+#### Scenarios `keycloak.scenario.admin.CreateRealms` and `keycloak.scenario.admin.CreateDeleteRealms`
 
-This scenario requires the following Client settings, in addition to the requirements above:
+These scenarios are using the root admin account to perform realm operations with the built-in `admin-cli` client.
 
-* In to the `Scopes` tab
-    * Set `Full Scope Allowed` to `Off`
-    * Add role `create-realm` in the `Realm Roles` section
-* In the `Service Account Roles` tab
-    * assign the roles `create-realm` in the `Realm Roles` section
+This information is specified to the scenarios with options `--admin-username` and `--admin-password`.
 
-The `Full Scope Allowed` flag is especially important to disable, because if enabled,
-the access token will contain *all* the accesses to all the realms,
-leading to the token becoming too large after roughly 40 created realms
-(receiving `431 Request Header Fields Too Large` responses in this case).
+Usage of a service account token is irrelevant with these scenarios, because:
+
+* real-world realm operations are performed using root admin credentials
+* deleting a just-created realm requires realm-specific permissions which are set onto a realm-specific
+  client, which would require to logout then login again using the realm-specific client to perform the realm deletion operation
+* as the token includes all realm permissions, it would grow very fast and would quickly exceed
+  the maximum length for header (leading to `431 Request Header Fields Too Large` responses).
 
 ### Run
 
@@ -83,7 +82,8 @@ These are the available test scenarios:
 * `keycloak.scenario.authentication.ClientSecret`: Client Secret (Client Credentials Grant)
 * `keycloak.scenario.admin.CreateDeleteClients`: Create and deleted clients (requires `--client-secret=<client secret for gatling client>`)
 * `keycloak.scenario.admin.UserCrawl`: Crawls all users page by page (requires `--client-secret=<client secret for gatling client>`)
-* `keycloak.scenario.admin.CreateRealms`: Create realms (requires `--client-secret=<client secret for gatling client>` and `--realm-name=master`)
+* `keycloak.scenario.admin.CreateRealms`: Create realms (requires `--admin-username=<admin login>` and `--admin-password=<admin password>`)
+* `keycloak.scenario.admin.CreateDeleteRealms`: Create and immediately delete realms (requires `--admin-username=<admin login>` and `--admin-password=<admin password>`)
 
 ## Release
 
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -7,7 +7,7 @@ import io.gatling.core.structure.ChainBuilder
 import io.gatling.http.Predef._
 import keycloak.Utils
 import keycloak.Utils.randomUUID
-import keycloak.scenario.KeycloakScenarioBuilder.{ADMIN_ENDPOINT, CODE_PATTERN, LOGIN_ENDPOINT, LOGOUT_ENDPOINT, TOKEN_ENDPOINT, UI_HEADERS, REALMS_ENDPOINT, downCounterAboveZero}
+import keycloak.scenario.KeycloakScenarioBuilder.{ADMIN_ENDPOINT, CODE_PATTERN, LOGIN_ENDPOINT, LOGOUT_ENDPOINT, TOKEN_ENDPOINT, UI_HEADERS, REALMS_ENDPOINT, MASTER_REALM_TOKEN_ENDPOINT, downCounterAboveZero}
 import keycloak.scenario._private.AdminConsoleScenarioBuilder.DATE_FMT
 import org.keycloak.benchmark.Config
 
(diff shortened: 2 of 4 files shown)


davoustp:
@ahus1  - for now, this scenario requires options `--username` and `--user-password` to be set with Keycloak root admin, until we sort this permission/role issue when using the `gatling` service account instead.

davoustp:
@ahus1 if you want to experiment with the concurrency situation, add `--concurrent-users=5` (for example) and `--log-http-on-failure`

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -319,6 +342,17 @@ class KeycloakScenarioBuilder {
...
   }
 
+  def deleteRealm(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("Delete realm")
+        .delete(REALMS_ENDPOINT + "/${createdRealmId}")
+        .header("Authorization", "Bearer ${token}")
+        .header("Content-Type", "application/json")
+        .check(status.is(204)))
+      .exitHereIfFailed
```
kami619, line 352: if the purpose of ${createdRealmId} is fulfilled, lets reap the memory back using 
`.exec(session => (session.removeAll("createdRealmId")))` after the `.check(status.is(204))` is performed on #351

benchmark/src/main/scala/keycloak/scenario/admin/CreateDeleteRealms.scala
```diff
@@ -0,0 +1,15 @@
+package keycloak.scenario.admin
+
+import keycloak.scenario.{CommonSimulation, KeycloakScenarioBuilder}
+
+
+class CreateDeleteRealms extends CommonSimulation {
+
+  setUp("CRUD realms", new KeycloakScenarioBuilder()
```
kami619, line 8: rename this to "Create, Delete Realms" or add the rest of the Update and Read operations in the simulation to List and/or Update the Realms

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
davoustp: Indeed. Moved the `removeAll` call from `createRealm` to `deleteRealm` (as this is required by this step anyway)

benchmark/src/main/scala/keycloak/scenario/admin/CreateDeleteRealms.scala
davoustp: Label changed.

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
kami619: Cool, thanks.

benchmark/src/main/scala/keycloak/scenario/admin/CreateDeleteRealms.scala
kami619: perfect

ahus1:
I converted this to a draft PR as long as we sort out the permissions issue.

ahus1:
@davoustp - I analyzed why there is a different behavior of using `admin-cli` vs. `gatling`. 

Here's the reason, you'd implemented IMHO the only way this can work: if the access token was issued for the client `admin-cli` or `security-admin-console`, Keycloak will use the permissions stored in its database. Otherwise, it will use the permissions stored in the token (See `MgmtPermission#initIdentity()`). 

* The permission to create new realm is a realm role.
* The permission to delete a realm is part of `manage-realm`, which is a client role. This client role is a composite role of the `admin` role in the `master` realm, and is updated on realm creation with each new realm.

This leads to the following problem when the client is named `gatling`:
* At the time of login, the token doesn't contain the client roles of the composite role for the to-be created realm. A workaround would be to log-out and to log-in again to get at token where one is allowed to delete the realm. 
* Still, the token would grow very fast once the `admin` role is part of the permissions for `gatling`, so this is not feasible.

When the client is either `admin-cli` or `security-admin-console`, it will read the permissions from the database and they don't need to be stored in the token.

Therefore, the only valid option I see here to run this test is to use the admin user and the admin-cli client ID as you did. 

To update this PR, I'd like you update this PR as follows (feel free to add pieces that I missed):
* put a short summary to the scenario why this is special
* rename `userAccountToken()` to something like  `adminCliToken()` to indicate that it will create an `admin-cli` token that is "special"
* instead of `username` and `password`, maybe introduce `adminUsername` and `adminPassword`, as this is no longer about the login of a regular user, but the admin.

Please let me know what you think. Once this PR is ready to be reviewed and merged, remove the draft status.

davoustp:
> The permission to delete a realm is part of manage-realm, which is a client role. This client role is a composite role of the admin role in the master realm, and is updated on realm creation with each new realm.

Yes, that's what I saw as well. I was kind of puzzled why the admin-cli client was working, though, but now I get it from your investigations. Thx!

> At the time of login, the token doesn't contain the client roles of the composite role for the to-be created realm. A workaround would be to log-out and to log-in again to get at token where one is allowed to delete the realm.

Yep, I thought about that approach but looked quite overkill and, most importantly, it does not reflect what users will actually do, so we'd test something which does not represent the real world use case.

> Therefore, the only valid option I see here to run this test is to use the admin user and the admin-cli client ID as you did.

Ok

> To update this PR, I'd like you update this PR as follows (feel free to add pieces that I missed):
* put a short summary to the scenario why this is special
* rename `userAccountToken()` to something like `adminCliToken()` to indicate that it will create an `admin-cli` token that is "special"
* instead of username and password, maybe introduce `adminUsername` and `adminPassword`, as this is no longer about the login of a regular user, but the admin.

Will do

Thanks for your time on this!

ahus1:
@hmlnarik - pushed some changes: Tests are now all green. Added LegacySessionSupport SPI. 

Will continue on Monday with polishing, especially JavaDoc and your recent comments.

davoustp:
> @hmlnarik - pushed some changes: Tests are now all green. Added LegacySessionSupport SPI.
> 
> Will continue on Monday with polishing, especially JavaDoc and your recent comments.

May not be related to this PR? ;-)

ahus1:
@davoustp - you're right :-) a Friday evening mistake. Will delete the comments soon to clean them up.

davoustp:
@ahus1 update performed.
* Added the new `--admin-username` and `--admin-password` options
* Added admin realm token endpoint constant to avoid using the `--realm` option as this is irrelevant
* Retro-fitted scenario `CreateRealms` scenario to also use the root admin credentials (would have been weird to continue using the service account token here)
* Updated documentation accordingly

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -314,6 +338,17 @@ class KeycloakScenarioBuilder {
         .body(StringBody("""{"id":"${createdRealmId}","realm":"${createdRealmId}","enabled": true}"""))
         .check(status.is(201))
         .check(header("Location").notNull.saveAs("realmLocation")))
```
kami619, line 340: given we are not using the `realmLocation` downstream, do we need to persist this into the session ?

davoustp: @kami619 not really - it was here to keep the code stream similar, potentially making maintenance easier.
I can remove this attribute `saveAs` if you think that it's better not to store it, no pb.

kami619: in the other methods, those location headers are consumed in, for example, `deleteClients` or `deleteClientScopes` Delete api's, here we followed a different way of deleting the Realm, hence I pointed out if we wanted to avoid saving this to session. I would rather delete it and keep the session memory lean as much as possible. Unless, you have an idea to use the Location header in any of the downstream methods for this entity.

benchmark/BENCHMARK.md
```diff
@@ -38,20 +38,19 @@ Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateRealms`, `Create
    * assign the roles `manage-clients` and `view-users`
 * the client secret to be passed to the tests can be copied from the `Credentials` tab
 
-#### Scenario `keycloak.scenario.admin.CreateRealms`
+#### Scenarii `keycloak.scenario.admin.CreateRealms` and `keycloak.scenario.admin.CreateDeleteRealms`
 
-This scenario requires the following Client settings, in addition to the requirements above:
+These scenarii are using the root admin account to perform realm operations with the built-in `admin-cli` client.
```
kami619, line 43: a small change needed on `scenarii` to `scenarios`

benchmark/BENCHMARK.md
```diff
@@ -38,20 +38,19 @@ Some scenarios (`CreateDeleteClients`, and `CrawlUsers`, `CreateRealms`, `Create
    * assign the roles `manage-clients` and `view-users`
 * the client secret to be passed to the tests can be copied from the `Credentials` tab
 
-#### Scenario `keycloak.scenario.admin.CreateRealms`
+#### Scenarii `keycloak.scenario.admin.CreateRealms` and `keycloak.scenario.admin.CreateDeleteRealms`
```
kami619, line 41: please update `Scenarii` to `Scenarios` for better readability

ahus1: +1 for "scenarios". [Wikitionary](https://en.wiktionary.org/wiki/scenarii) names this as a "([nonstandard](https://en.wiktionary.org/wiki/Appendix:Glossary#nonstandard), [rare](https://en.wiktionary.org/wiki/Appendix:Glossary#rare)) [plural](https://en.wiktionary.org/wiki/Appendix:Glossary#plural_number) of [scenario](https://en.wiktionary.org/wiki/scenario#English)".

As most of the visitors to this repository are non-native English speakers, I'd like to keep wording as main-stream as possible.

davoustp: Done

benchmark/BENCHMARK.md
davoustp: Done

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
davoustp: Done

ahus1 (approved):
Thank you for this PR so that we can now test creation and deletion of realms!

PR #74, opened 2022-05-18
Limit the number of groups and roles returned by adding the necessary query parameters

### LINKED ISSUE #73: Listing Groups and Roles doesn't limit the results
### Describe the bug

When the scenarios CreateGroups and CreateRoles retrieve the list of groups and roles, they should limit the result to 2 entries, which they currently don't do.

### Version

0.7-SNAPSHOT

### Expected behavior

The returned list of groups should be 2

### Actual behavior

As the `first` parameter in the query is not set, the result is not filtered.

### How to Reproduce?

Run the CreateGroups and CreateRoles scenario, and place a breakpoint at the rest response. This can also be checked by looking at the returned result.

### Anything else?

_No response_

ahus1:
Also checking the result so that it contains max. 2 entries.

Closes #73

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -312,8 +312,11 @@ class KeycloakScenarioBuilder {
       .exec(http("List Roles")
         .get(ADMIN_ENDPOINT + "/roles")
         .header("Authorization", "Bearer ${token}")
+        .queryParam("first", 0)
         .queryParam("max", 2)
-        .check(status.is(200)))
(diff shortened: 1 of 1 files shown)


ahus1:
Hi @kami619 - I added the missing query parameter and added a safeguard on the response so that we'll get an alert once something similar happens again. 

Please give it a review, and once it is good for you rebase and merge it via the GitHub API. 

The full performance for the CreateGroups test will only arrive after keycloak/keycloak#12083 has been merged, but at least roles should beform good enough after this change.

kami619:
CreateGroups simulation now yields expected outcome. thanks for the update on this module @ahus1. the changes looks good to me.

kami619 (approved):

PR #76, opened 2022-05-25
Add a bash utility to automate the service account enabled client management steps

### LINKED ISSUE #75: Automate the creation of Service Account Enabled Client used by gatling
### Description

Automate the creation of Service Account Enabled Client used by gatling benchmark module for some of the load testing scenarios.

### Discussion

_No response_

### Motivation

Right now we have to set this client up manually which sometimes can be tedious to do it repeatedly for any reason you have to reap and recreate the Keycloak server and its data. So automating this step would save multiple minutes of developer time :) 

### Details

The idea is to have a command line utility which wraps around the kcadm.sh CLI script to do the set up for the client. This way, we have predictability of the roles to needed for the Client and setup a static secret for the runs, which keeps things simple.

kami619:
Implementation details:

A simple bash utility which consumes the kcadm.sh CLI script bundled with the Keycloak distributions. 

Documentation note:
Detailed instructions are update in the README for the benchmark module

Closes #75

### DIFF
--- a/benchmark/manage_gatling_client.sh
+++ b/benchmark/manage_gatling_client.sh
@@ -1,13 +1,16 @@
 #!/bin/bash
-
+set -e
+set -o pipefail
+# set -x
 #Functions
 function usage {
+  echo "Setup of test clients for the benchmark test with a pre-defined secret"
   echo "Usage: $(basename $0) -k KEYCLOAK_HOME -r REALM_NAME -c CLIENT_ID -s SERVER_URL [-d]" 2>&1
-  echo '-k keycloak home path ex: /home/opt/keycloak-18.0.0 '
+  echo '-k keycloak home path ex: /home/opt/keycloak-18.0.0 : default value KEYCLOAK_HOME env variable'
   echo '-r realm : default value realm-0'
   echo '-c service account enabled client name : default value gatling'
   echo '-s keycloak host server url : default value http://localhost:8081/auth'
-  echo '-d delete the client and realm : default value is false'
+  echo '-d delete the client and realm before re-creating them: default value is false'
   exit 1
 }
 
@@ -17,10 +20,10 @@ function install_kcb {
 }
 
 function create_realm {
-  if kcadm.sh get realms  | grep \"realm\" | grep -q $REALM_NAME; then
+  if kcadm.sh get realms/${REALM_NAME}; then
     echo "INFO: Skipping Realm creation as realm exists"
   else
-    echo "INFO: Creating Realm with realm id : ${REALM_NAME}"
+    echo "INFO: Creating Realm with realm id: ${REALM_NAME}"
     kcadm.sh create realms -s realm=$REALM_NAME -s enabled=true -o
   fi
 }
@@ -30,7 +33,7 @@ function create_client_assign_roles {
   CID=$(kcadm.sh create clients -r $REALM_NAME -s clientId=$CLIENT_ID -s enabled=true -i)
   echo "INFO: Created New ${CLIENT_ID} Client with Client ID ${CID}"
   #Update the client with necessary attributes
-  kcadm.sh update clients/$CID -r $REALM_NAME  -s "secret=4CYUuDEYvvAkjIcjcB0EgAJxmbXdgMYR" -s serviceAccountsEnabled=true  -s publicClient=false -s 'redirectUris=["http://localhost:8081"]'
+  kcadm.sh update clients/$CID -r $REALM_NAME  -s "secret=setup-for-benchmark" -s serviceAccountsEnabled=true  -s publicClient=false -s 'redirectUris=["http://localhost:8081"]'
   kcadm.sh get clients/$CID -r $REALM_NAME --fields clientId,id,redirectUris,publicClient,serviceAccountsEnabled,secret
 
   #Verify if the service account user exists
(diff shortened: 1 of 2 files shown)


benchmark/manage_gatling_client.sh
```diff
@@ -0,0 +1,99 @@
...
+  echo '-k keycloak home path ex: /home/opt/keycloak-18.0.0 : default value KEYCLOAK_HOME env variable'
+  echo '-r realm : default value realm-0'
+  echo '-c service account enabled client name : default value gatling'
+  echo '-s keycloak host server url : default value http://localhost:8081/auth'
+  echo '-d delete the client and realm before re-creating them: default value is false'
+  exit 1
+}
+
+function install_kcb {
+  echo "Installing kcadm.sh"
```
ahus1, line 18: This prompt on the console "Installing..." would indicate something would be permanently installed and would be in effect after the script has been run. As it is setting just the path, please rename the method and the prompt.

benchmark/manage_gatling_client.sh
```diff
@@ -0,0 +1,99 @@
...
+  #Create the client application with Service Account Roles setup
+  CID=$(kcadm.sh create clients -r $REALM_NAME -s clientId=$CLIENT_ID -s enabled=true -i)
+  echo "INFO: Created New ${CLIENT_ID} Client with Client ID ${CID}"
+  #Update the client with necessary attributes
+  kcadm.sh update clients/$CID -r $REALM_NAME  -s "secret=setup-for-benchmark" -s serviceAccountsEnabled=true  -s publicClient=false -s 'redirectUris=["http://localhost:8081"]'
+  kcadm.sh get clients/$CID -r $REALM_NAME --fields clientId,id,redirectUris,publicClient,serviceAccountsEnabled,secret
+
+  #Verify if the service account user exists
+  kcadm.sh get users -r $REALM_NAME -q username=service-account-$CLIENT_ID --fields id,username,enabled
+
```
ahus1, line 41: It seems to me that the two "kcadm.sh get" commands are a left-over debugging the script and getting the initial functionality in place. Please consider removing.

benchmark/manage_gatling_client.sh
```diff
@@ -0,0 +1,99 @@
...
+  #Assign the necessary Service Account based roles to the client
+  kcadm.sh add-roles -r $REALM_NAME --uusername service-account-$CLIENT_ID --cclientid realm-management --rolename manage-clients --rolename view-users --rolename manage-realm --rolename manage-users
+  #kcadm.sh get-roles -r $REALM_NAME --uusername service-account-$CLIENT_ID --cclientid realm-management
```
ahus1, line 44: Commented out code, please consider removing.

benchmark/manage_gatling_client.sh
```diff
@@ -0,0 +1,99 @@
...
+OPTIND=1
+REALM_NAME=realm-0
+CLIENT_ID=gatling
+SERVER_URL="http://localhost:8081/auth"
+DELETE_ENTITIES='false'
+
+while getopts "s:k:r:c:d:" arg; do
+  case $arg in
+    s) SERVER_URL="$OPTARG";
+        ;;
```
ahus1, line 64: SERVER_URL doesn't seem to be used inside the script, consider removing here an in all other places in the script and in the docs.

ahus1 (changes requested):
Thank you for this script for automatic provisioning. I reviewed it, and pushed some aspects already as a commit, see 40cc873019f882a17cf2c1e74304430728806bc3. 
It includes the following: 
* advanced error handling with "-e" and "-o pipefail": when a command exists with an exit code of non-zero and that is not checked, the shell script will exit as well. I added those when there were error messages due to some failed commands when I didn't log in before using the script. Instead of continuing, the script will now exit on the first failed command.
* The "set -x" is there commented out, so it can be enabled when debugging is needed - if you think it should be removed, please do so
* To check that a realm exists, it is simpler to get it by realm name than to grep the results, therefore I changed the call. General comment: instead of grepping JSON output, consider using jq instead for a more robust parsing (good new: no greps of JSON are left in the shell script)
* The secret name is no longer looking random, so people will know it has been set by this script
* when deleting entities, here is no need to delete the client first before deleting the realm, it is sufficient to just delete the realm
* when using getopts, it needs a colon (:) after the letter if the option has an argument. All options have arguments AFAIK, therefore each letter now has a colon. Without the change some CLI options weren't effective as "$OPTARG" wasn't set.
* DELETE_ENTITIES was true all the time, I suppose a left-over from debugging

General remarks / todos for the next iteration: 
* The script still logs quite a lot of things on the console (either error messages or JSON), please revisit those and see which are helpful to the user, and which can be removed/filtered

benchmark/manage_gatling_client.sh
kami619: yep, they are not needed in the current state of the script. will remove them

benchmark/manage_gatling_client.sh
kami619: I will remove this commented code, thanks

benchmark/manage_gatling_client.sh
kami619: true, now that I have pulled out the establishing session with Keycloak server outside of this script, this variable is not used anywhere. I can remove this from the script and the docs.

kami619:
@ahus1 I have mostly incorporated all the changes specified in your PR review, with the below exceptions.

- I chose to remove the : next to getopts `d` to not take any argument, as we are treating it as a boolean in the conditional check
- I have added a `grep -q {REALM_NAME}` pipe command to silence the JSON output from the if condition `if kcadm.sh get realms/${REALM_NAME}` in couple of places. If you are aware of an option to silence the output from kcadm.sh directly, that would be the ideal solution than what I tried to implement
- Rest of the places, it should log close to minimum, and provide errors for any issues, while running the commands, as I have used `> /dev/null` to not silence the `stderr`

Please let me know your thoughts on this.

ahus1 (approved):
Thank for the changes, the console is now tidied up nicely. I also like that the delete option doesn't take an argument any more. I also reflected this in the `getopts` handling in commit d11f6514baabd1b6fb5e1d7deed7da706a14e141. 

Ready to be merged!

kami619:
@ahus1 thank you

PR #77, opened 2022-05-31
Join group test scenario.

artur-baltabayev:
This scenario creates a new user, finds a given group and then adds the user to that group.
It can be used to check if group assignments still perform in reasonable time frames on systems with a lot of entities (especially groups and roles).

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -364,7 +364,7 @@ class KeycloakScenarioBuilder {
         .post(ADMIN_ENDPOINT + "/users")
         .header("Authorization", "Bearer ${token}")
         .header("Content-Type", "application/json")
-        .body(StringBody("""{"username":"joinGroup-${username}"}"""))
+        .body(StringBody("""{"username":"${username}"}"""))
         .check(status.is(201))
         .check(header("Location").notNull.saveAs("user-location")))
       .exitHereIfFailed
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -357,6 +357,44 @@ class KeycloakScenarioBuilder {
     this
   }
 
+  def createUser(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .feed(Iterator.continually(Map("username" -> randomUUID())))
+      .exec(http("Create user")
+        .post(ADMIN_ENDPOINT + "/users")
+        .header("Authorization", "Bearer ${token}")
+        .header("Content-Type", "application/json")
+        .body(StringBody("""{"username":"joinGroup-${username}"}"""))
+        .check(status.is(201))
+        .check(header("Location").notNull.saveAs("user-location")))
+      .exitHereIfFailed
+    this
+  }
+
+  def joinGroup(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("Join group")
+        .put("${user-location}/groups/${group-id}")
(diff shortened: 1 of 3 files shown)


benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -357,6 +357,44 @@ class KeycloakScenarioBuilder {
...
+    chainBuilder = chainBuilder
+      .exec(http("Join group")
+        .put("${user-location}/groups/${group-id}")
+        .header("Authorization", "Bearer ${token}")
+        .check(status.is(204)))
+      .exitHereIfFailed
+    this
+  }
+
+  def findGroup(groupName: String): KeycloakScenarioBuilder = {
```
kami619, line 384: Hi Thanks for the PR, this is a good scenario to cover. However I have a question on to why are we providing the group name that the user needs to be joining as an input argument to the kcb.sh script ? Can we not make self-sufficient to search for a random group with some defined constraints such as number of users already assigned to the group etc., 

The reason I ask for this is because if you have to provide the name of the group to be assigned it might add more complexity on the test execution and may have to maintain a list of possible groups to be assigned to.

I wanted to also check on, if you have a better way of handling the `join-group-group-name` queue thats going to be fed into the kcb.sh script ?

artur-baltabayev: Thanks for the reply @kami619 !

We built it that way, because it fits our specific test case. We first populate a realm with a lot of groups and then chose one that the user will join. If you find it more fitting, we could make it self sufficient. Then we could also drop the `join-group-group-name` property and `findGroup` method all together. The `joinGroup` method could just query the groups (on the first call) and just pass the list to subsequent calls.
The upside of the current approach is that the API gets more flexible and would technically allow for more complex scenarios rather than self sufficient methods. Just for context, we already built a test on top of this logic (on our fork) and the complexity is very managable so far.

Your feedback on that would be very appreciated !

ahus1: I could also see the case where one you have a lot of users joining a single group to test that specific cardinality, as I would expect Keycloak to slow down with the number of users in a single group. I'm just curious about the difference it makes for Keycloak with the number of groups. I'd by interested about the results, but that's out of scope for this PR. 

I leave the final vote on this thread to @kami619.

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -357,6 +357,44 @@ class KeycloakScenarioBuilder {
...
   }
 
+  def createUser(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .feed(Iterator.continually(Map("username" -> randomUUID())))
+      .exec(http("Create user")
+        .post(ADMIN_ENDPOINT + "/users")
+        .header("Authorization", "Bearer ${token}")
+        .header("Content-Type", "application/json")
+        .body(StringBody("""{"username":"joinGroup-${username}"}"""))
```
ahus1, line 367: The user will always have the prefix "joinGroup-", and looking the method name this surprises me. To be able to reuse it in different scenarios, I'd like to have it without the prefix "joinGroup-". It could be a more generic one like "user-", or just no prefix.

ahus1 (changes requested):

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
kami619: thats a valid point as well, thanks for highlighting that @ahus1. I am fine to add this scenario and probably I (or anyone) can look into adding a separate scenario where the user can join a random group too.

kami619 (approved):
lgtm

ahus1 (approved):
LGTM!

PR #81, opened 2022-06-01
Use keycloak-operator to deploy keycloak instead of using a deployment in helm

### LINKED ISSUE #80: Upgrade the minikube setup to use the Keycloak Operator
### Description

To test the setup with the Keycloak operator and not to replicate its logic with Helm charts, use the nightly Keycloak operator.

ahus1:
Closes #80


ahus1:
@kami619 - this PR updates the minikube setup to use the Keycloak operator to deploy Keycloak. All the extras (database, monitoring) is still deployed using helm.

Please give it a try. To update your local installation, you can try to run the `upgrade.sh` script. For this instance, you might need to run it twice, as the Keycloak namespace will be removed on the first run, and recreated on the second run. From now on, the namespace keycloak is no longer created by helm, but by the CLI.

cc: @tkyjovsk

andreaTP:
Looking at the CI there might be something that is not working correctly/as expected, the Keycloak Pos is still in Unhealty state:
https://github.com/keycloak/keycloak-benchmark/runs/6686795355?check_suite_focus=true#step:6:2164

Should we have those changes propagated to the `Makefile` too?
https://github.com/keycloak/keycloak-benchmark/blob/b5ebb435780d3a92f572b58a8d735ad7450054d0/provision/minikube/Makefile#L42
( having two sources of truth is not ideal though :-S we are always going to forget to update one)

provision/minikube/rebuild.sh
```diff
@@ -26,6 +26,10 @@ helm repo add grafana https://grafana.github.io/helm-charts
 helm install --debug tempo grafana/tempo -n monitoring -f tempo.yaml
 helm install --debug loki grafana/loki -n monitoring -f loki.yaml
 helm install --debug promtail grafana/promtail -n monitoring -f promtail.yaml
+kubectl create namespace keycloak
```
andreaTP, line 29: it would be nice if those instructions will be shared, something like a separate `install_keycloak.sh` called in `rebuild.sh`, `upgrade.sh` and `Makefile` ?

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - name: db-password
+      secret:
+        name: keycloak-db-secret
+        key: password
+  tlsSecret: keycloak-tls-secret
+  unsupported:
+    podTemplate:
+      spec:
+        containers:
+          - name: keycloak
```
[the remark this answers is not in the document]
ahus1: It just feels strange to had a container without a name ... 

What if I would add a second (new) container with a different name, ie a sidecar? How would that / should that behave?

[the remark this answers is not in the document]
ahus1: Will remove the name in the next commit.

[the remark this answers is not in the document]
andreaTP: > What if I would add a second (new) container with a different name, ie a sidecar? How would that / should that behave?

It will be appended to the `podTemplate` and work as expected (the operator won't mangle anything).
The idea behind having a few fields "protected" is that the operator "owns" the container 0, do you have any suggestion/feedback in this regard?

provision/minikube/rebuild.sh
ahus1: Both `makefile` and `upgrade.sh` are not ideal as they don't really allow a seamless upgrade. I'd love to see `make` to install only what has changed, and to switch between install and upgrade for the help command as necessary. But I'll keep that for another PR.

ahus1:
@andreaTP - about the logfile, there was indeed a problem: while the keycloak ingress wasn't created by the operator yet, it would return a different site as a default, therefore the test in `isup.sh` completed without actually waiting for Keycloak. 

The latest commit to this PR now fixes this. It also cleans up the logging so that the different CLIs don't overwrite their consoles.

andreaTP:
@ahus1 thanks for the update!
Maybe having stricter checks in bash scripts (e.g. https://github.com/keycloak/keycloak/blob/74870a2ac615e8526c9e6aaf85bfa1e0ec0ab8e8/operator/app/scripts/check-crds-installed.sh#L2 ) might help catching issues in the future, but not fussy on having it in this PR.

provision/minikube/rebuild.sh
andreaTP: Thanks for explaining!
Maybe naive question, have you tried to use everywhere `helm upgrade --install`?

ahus1:
@andreaTP - thanks for the hint, it seems that I missed this option. This allowed me to simplify the scripts. 

I also update the Makefile ... it needs fewer appended `true` commands. I also changed the `|` (pipe) to a true or `||` - this will print the messages so it doesn't hide the error messages.

provision/minikube/isup.sh
```diff
@@ -13,19 +13,19 @@ fi
 MAXRETRIES=600
```
andreaTP, line 13: this check might need some tuning to run on Mac, but not a big deal for now...

provision/minikube/isup.sh
```diff
@@ -13,19 +13,19 @@ fi
...
  )
 
 for SERVICE in "${!SERVICES[@]}"; do
   RETRIES=$MAXRETRIES
   # loop until we connect successfully or failed
-  until curl -k -f -v ${SERVICE}${SERVICES[${SERVICE}]} >/dev/null 2>/dev/null
+  until kubectl get ingress -A 2>/dev/null | grep ${SERVICE} >/dev/null && curl -k -f -v https://${SERVICE}/${SERVICES[${SERVICE}]} >/dev/null 2>/dev/null
   do
-    if [ "${RETRIES}" == "${MAXRETRIES}" ]
+    if [ "${RETRIES}" == "${MAXRETRIES}" ] && [ "$GITHUB_ACTIONS" == "" ]
```
andreaTP, line 28: The "canonical" way to check if you are on CI is to check the `CI` env variable:
https://docs.github.com/en/actions/learn-github-actions/environment-variables

provision/minikube/keycloak/templates/keycloak-initial-admin-secret.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
andreaTP, line 1: This file looks unused in this change, am I missing something?

ahus1: If I would not add this secret, the keycloak operator would create a secret with a random password for the admin, and this would be annoying. Pre-creating this secret allows me to have a pre-defined well-known password for the admin user in this setup. 

Therefore it's related and I'd like to keep it.

provision/minikube/isup.sh
ahus1: Thanks, changed in latest commeit.

provision/minikube/keycloak/templates/keycloak-initial-admin-secret.yaml
andreaTP: You are right, thanks for explaining, not sure if this is a bug or a feature though 😄 
https://github.com/keycloak/keycloak/blob/c31d37ddf1c3e3529e761800a6842fdbf25ba931/operator/app/src/main/java/org/keycloak/operator/controllers/KeycloakAdminSecret.java#L24

cc. @vmuzikar

andreaTP (approved):
LGTM, nice work!

provision/minikube/keycloak/templates/keycloak-initial-admin-secret.yaml
vmuzikar: Undocumented feature. 🙂

kami619 (approved):
looks good to me thanks for this @ahus1

PR #94, opened 2022-06-10
Add a starter grafana dashboard to support the minikube setup and load tests (fixes #93)

### LINKED ISSUE #93: Add a starter grafana dashboard to support the minikube setup and load tests
### Description

Add a starter grafana dashboard to support the minikube setup and load tests

- Add a panel to display the container level metrics CPU/Memory/Network at a selected namespace level
- Add a panel to display the Keycloak JVM metrics
- Add a panel to display the Gatling statistics during a run
- Add a panel for projecting the PostgreSQL metrics

### Discussion

_No response_

### Motivation

to be able to leverage the various stats that are being collected by the prometheus collector in the minikube and enhance the way we can monitor the minikube cluster

### Details

create a grafana dashboard that would be imported during the provisioning and be usable during the load tests

kami619:
Closes #93 

*Implementation Notes:*

 I have added a keycloak-perf-tests.json to the provision/minikube/monitoring/dashboards directory, which would the be deployed automatically to the grafana instance that's provisioned. This should allow us to monitor the Keycloak eco system as well as the Gatling load test tool live metrics with ease. Along the way there were few issues with the keycloak-metrics path, which was fixed with help of @ahus1.


ahus1:
Hi Kamesh, thank for creating this dashboard. With this new dashboard, I decided that the old dashboard is no longer needed and  deleted it. The only thing I changed was to update the network IO labels with a transmit/receive suffix. 

Good work!

PR #99, opened 2022-06-20
Add cryostat to enable JFR recordings for Keycloak.

### LINKED ISSUE #98: Support recording JFR reports for containers running in Minikube
### Description

To be able to analyze the in-depth action of a JVM, it should be possible for engineers (and when requested/deemed necessary also for QE) to create a JFR report.

One possible way to do this is https://cryostat.io/

ahus1:
It proved possible to add Cryostat to do the JFR recordings. Login is disabled like with all other tools here to enable fast turnaround times for developers working on this.

For an OpenShift setup where the routes would be publicly accessible this would then be changed. 

I did a test for Keycloak and downloaded the JFR report. The first flamegraph showed truncated stack dumps (default would be 64?), therefore I configured the stackdepth to 512. 

The option to create JFR recordings is enabled by default, but no JFR recording is created until this is triggered.

The profiling created is regular profiling, not async profiling. The profiling will therefore suffer from the safepoint bias problem. See [async profiler](https://github.com/jvm-profiling-tools/async-profiler#async-profiler) for details:

> This project is a low overhead sampling profiler for Java that does not suffer from [Safepoint bias problem](http://psy-lob-saw.blogspot.ru/2016/02/why-most-sampling-java-profilers-are.html). It features HotSpot-specific APIs to collect stack traces and to track memory allocations. The profiler works with OpenJDK, Oracle JDK and other Java runtimes based on the HotSpot JVM. 

For now, not using async profiling should be good enough until proven otherwise.

Here a screenshot of the UI.

image

Closes #98


ahus1:
@hmlnarik - this PR shows a solution how JFR can be part of this setup. While Cryostat provides a point-and-click Web interace, it also has a REST API to interact with it: https://github.com/cryostatio/cryostat/blob/main/HTTP_API.md

While it supports regular profiling and not async profiling, I will assume until proven otherwise that it serves the need for JFR recordings.

ahus1:
@kami619 - I'd be happy if you could give it a try and test the following: 

* spin up Minikube with these changes (the `upgrade.sh` script should work here)
* create a JFR recordings and download it
* open such a recording for example in Eclipse Mission Control https://adoptium.net/jmc

Thanks!

ahus1:
> Cryostat can initiate JFR recording via JMX and downloads the created file?

@hmlnarik - yes.

ahus1:
Hi @kami619 - I fixed the problem with a missing placeholder for the hostname. I also added an increased timeout when retrieving the cadvisor metrics. Please give it another try. Thanks!

kami619:
this works as expected, thanks for this great utility addition 🥇 @ahus1

PR #107, opened 2022-06-26
Adding Utility to create custom reports (fixes #100)

### LINKED ISSUE #100: Addition of a utility to be able to create differential and trend line reports using an existing repository of gatling test run results
### Description

Addition of a utility to be able to create differential and trend line reports using an existing repository of gatling test run results

A shell script of sorts to be able to point to two or more runs and provide us with the information to be able to generate digestible HTML/CSV reports utilizing the https://github.com/nuxeo/gatling-report tool

### Discussion

_No response_

### Motivation

It would be a good addition to the tool set, if we are able to automate the comparative analysis of a given set of gatling test run results. As part of that process, here is a small utility which can generate a differential report across two gatling test runs and a trend report for a collection of gatling results (more than 2).

The motivation for this feature request is to create a shell script which can take arguments to be consumed by the gatling-report jar file and generate us with the needed reports.

### Details

The shell script would take in arguments such as
1. location of the gatling results
2. which type of report we would like to generate : HTML, CSV
3. target directory location where these generated reports are stored

the script should be able to determine to create a differential or trend report, based on the number of simulation.log files it would be able to find within the source gatling test result directory.

kami619:
### Implementation note:

- Utility downloads the needed binary. 
- Generates the report taking the source simulation logs and a target report directory and writes the HTML|CSV report there. 
- In the end, it removes the binaries. 
- The utility can also optionally consume a custom mustache template used to customize the reports, which is currently hosted under benchmark/src/main/resources.

Documentation note:
Updated the Report section of the benchmark/BENCHMARK.md with the new utility usage information.

Closes #100

### DIFF
--- a/benchmark/BENCHMARK.md
+++ b/benchmark/BENCHMARK.md
@@ -107,7 +107,19 @@ You can set different options to change how tests should be executed. For a comp
 
 ### Report
 
-Check reports at the `result` directory.
+For the default gatling generated reports, check reports at the `result` directory.
+
+To generate a custom differential or trend line report from an existing repository of gatling simulations, you can use the [utility script](generate-custom-report.sh) based on [gatling-report](https://github.com/nuxeo/gatling-report/blob/master/README.md) project.
+
+Example Usage:
+```shell
+#for differential with custom template
+./generate-custom-report.sh -v 6.0 -s "~/JPAMapUndertow/simulation.log ~/HotRodUndertow/simulation.log" -d ~/reports/differential/templates -t src/main/resources/diff-v2.mustache
+
+#for trend lines with default template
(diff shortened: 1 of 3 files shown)


ahus1 (approved):
Hi @kami619 - great to have this now documented and ready-to-run with the script. 

I had a look at the script and saw that it has some defaults, so I decided to print out the defaults as part of the usage description. 

When looking at the usage description, I first thought that "source report path" actually meant the folder, I then found out that it should point to the simulation log directly. Once I pointed it to the file, it worked as expected. I therefore updated the usage description in the shell script. 

This should now be ready to squash and merge. Thanks!

kami619:
thanks for adding the defaults and sorry about the confusion on the source path input, agree it should be consistent in the usage as well.

PR #114, opened 2022-06-29
Workaround for task being stuck when run on a single core.

### LINKED ISSUE #113: The utility `task` might get stuck when run on a single thread
### Describe the bug

This is probably due to https://github.com/go-task/task/issues/715

### Version

main

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
Closes #113

### DIFF
--- a/provision/minikube/README.adoc
+++ b/provision/minikube/README.adoc
@@ -316,7 +316,9 @@ Can be combined with a task name.
 +
 [WARNING]
 ====
-There seems to be an open bug that can lead to deadlocks, see https://github.com/go-task/task/issues/715[#task/715].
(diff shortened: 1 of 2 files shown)


ahus1:
Hi @kami619 - this implements a workaround for task being stuck when run with the `-C 1` option. 

Please have a look. Tried to run it multiple time with `task -C 1`, and it is no longer getting stuck.

kami619 (approved):
thanks for the fix, now it works as expected. In its current state, the `task -f -C 1` on an existing cluster takes about 6m43s total time, which is not bad.

PR #116, opened 2022-06-30
Ensure that the task started by create-users/ is removed from task li…

### LINKED ISSUE #115: Task started by create-users does never finish
### Describe the bug

When running `GET https://localhost:8443/auth/realms/master/dataset/create-users?realm-name=realm-0&count=100&task-timeout=18000` a task is started that creates 100 users.

The actual task is finished after a few seconds, but `GET https://localhost:8443/auth/realms/master/dataset/status` still reports a running task, subsequent starts of tasks are not possible anymore

### Version

current main

### Expected behavior

`GET https://localhost:8443/auth/realms/master/dataset/status`  shows that the task has terminated, new tasks can be started

### Actual behavior

`GET https://localhost:8443/auth/realms/master/dataset/status`  says that the task is still running, no new task can be started

### How to Reproduce?

see description above

### Anything else?

Since I already found the root cause of this I'll provide a PR

ioemat:
…st when done.

Resolved #115

### DIFF
--- a/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
+++ b/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
@@ -389,11 +389,13 @@ public class DatasetResourceProvider implements RealmResourceProvider {
             executor.waitForAllToFinish();
 
             timerLogger.info(logger, "Created all %d users in realm %s", context.getUserCount(), context.getRealm().getName());
+            new TaskManager(baseSession).removeExistingTask(true);
 
         } finally {
             if (executor != null) {
                 executor.shutDown();
             }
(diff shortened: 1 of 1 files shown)


kami619:
```{"timestamp":"2022-06-30T16:25:57.401Z","sequence":11506,"loggerClassName":"org.jboss.logging.Logger","loggerName":"org.keycloak.benchmark.dataset.TaskManager","level":"INFO","message":"FINISHED TASK: Creation of 1000 users in the realm realm-5, started: Thu Jun 30 16:25:03 GMT 2022","threadName":"Thread-9","threadId":86,"mdc":{},"ndc":"","hostName":"keycloak-5cdf6d7bb9-wjprb","processName":"QuarkusEntryPoint","processId":1}
``` now the create users task completes without issue.

PR #119, opened 2022-07-02
Change from docker runtime to cri-o runtime

### LINKED ISSUE #117: minikube: needed to increase the scrapeTimeout for cadvisor metrics two times with increase in the cluster size
### Describe the bug

We had to increase the scrapeTimeout under the provision/minikube/monitoring/templates/cadvisor-podmonitor.yaml for cadvisor metrics two times with increase in the cluster size.

When we added more components to be monitored as well as the up-size in the system resources and the longer the cluster is up, we ended up in couple of situations where the scrapeTimeout interval had to be increased other wise the Prometheus side, we experience web socket time outs.

### Version

NA

### Expected behavior

We would like to keep this under control and not to consistently up keep with the scaling of the cluster or the components under monitoring. 

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
This should allow for better response times when retrieving metrics from cadvisor.

Closes #117


ahus1:
@kami619 - I remember a similar problem in a previous problem. Changing to the CRI-O runtime solved this. 

I updated the scripts; one needs to run `rebuild.sh` to create a fresh minikube to use it.

Running this locally, I now get (on a fresh setup) scrape times of 2-3 seconds. Use the following metrics to see the measured scrape times: `scrape_duration_seconds{job=~"cadvisor/cadvisor"}`.

Please give it a try and let me know if this works better in the other environments as well.

kami619:
@ahus1 thanks for this fix, I think the scrape duration went down, but after having this setup run for 3 days, the scrape duration currently is at 14.409s. Is there any way to monitor the scrape duration over a period of time, alternatively I can note down independent time slice scrape durations over next couple of days, to see how this scales

ahus1:
> Is there any way to monitor the scrape duration over a period of time

Hi @kami619 - this snippet is a metric: `scrape_duration_seconds{job=~"cadvisor/cadvisor"}`

Please put it into the explore view in Grafana for Prometheus to see how it evolves over time.

ahus1:
@kami619 - did you have this setup running locally or on the locksmith machine? Maybe I could have a look together with you and see if something is different on the machine.

kami619:
@ahus1 the datapoint I shared is with the locksmith host, let me DM you on this.

PR #121, opened 2022-07-04
Add persistent storage to Prometheus, Jaeger and Loki

### LINKED ISSUE #120: Persist metrics storage in minikube
### Description

The current setup doesn't use persistent storage for Prometheus, Jaeger and Loki. Due to this, all data is lost once a container restarts.

To change this each of those should get a persistent storage, for now with ~60 days of data retention. 

### Discussion

_No response_

### Motivation

Metrics shouldn't be lost on restarting pods, and should be comparable between Keycloak releases. 

### Details

_No response_

ahus1:
This adds a data retention for 60 days. 

As in the minikube setup the persistent store will write the contents to a mounted disk, the file system needs to be watched so that is not full; I'll issue a new ticket for that.

Closes #120


ahus1:
Hi @kami619 - I'd be happy if you could give this one  are review. Thanks!

provision/minikube/loki.yaml
```diff
@@ -1,5 +1,15 @@
...
+persistence:
+  enabled: true
+  accessModes:
+    - ReadWriteOnce
+  size: 1Gi
+
+config:
+  table_manager:
+    retention_deletes_enabled: true
+    retention_period: 60d
```
kami619, line 15: could this retention period and the one from monitoring.yaml, be abstracted to values.yaml, so that we have a single place to control the value from

provision/minikube/Taskfile.yaml
```diff
@@ -68,6 +68,7 @@ tasks:
     sources:
```
kami619, line 68: to test this, would it just be enough to run `task` or `task -f` on an existing cluster ? I am making sure, if a total rebuild of the minikube VM is necessary or not

provision/minikube/jaeger/deployment-patch.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: jaeger
+          env:
+            - name: SPAN_STORAGE_TYPE
+              value: badger
+            - name: BADGER_SPAN_STORE_TTL
+              value: 1440h0m0s # 1440 hours = 60 days
```
kami619, line 11: even this value, along with other retention period values, if we could create a logic to control them from one place, it would be great. If not, alternatively, it would be nice to point them out in the docs, till the point, we have some central way of controlling this value

kami619:
@ahus1 this looks great, I just had couple of comments, let me know WDYT about those.

provision/minikube/Taskfile.yaml
ahus1: To test this, running `task` is sufficient. Sorry for now putting this in the PR in the first place.

provision/minikube/loki.yaml
ahus1: This file is already a values file, and those values are IMHO best overwritten via Helm's CLI. 

I've now leveraged the "dotenv"-feature of task; you can now use `KB_RETENTION` in and `.env` file to override that value.

ahus1:
I needed to update this PR as the configurable retention time wasn't working as expected. Please give it another try. 

For Prometheus, the retention time is visible in the command line flags via the UI. 

For Jaeger, this is visible in the pod as it is being passed as an environment variable. 

For Loki, it is visible in the loki secret in the monitoring namespace. 

Please give it another test once the build is green. Sorry for the trouble.

kami619:
Thank you after this fix, I am able to confirm the persistent storage retention period is being honored with the dotenv file based approach.

kami619 (approved):

PR #124, opened 2022-07-04
Fix URL pattern for Quarkus distribution

### LINKED ISSUE #122: Authentication-Code-Scenarios don't work with the Quarkus distribution
### Describe the bug

They expect the "/auth" path to be present that isn't there in the Quarkus distribution.

### Version

main

### Expected behavior

The scenarios should work with the Quarkus and the Wildfly distribution.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
This makes the "/auth" prefix optional.

In order to test it with the minikube setup, I needed to manually:

* enable user self-registration in `realm-0`
* created a client "client-0" with a valid redirect URL - I chose "https://keycloak.xxx.xxx.xxx.xxx.nip.io/*" (replace with the the current IP address)
* created a "user-0" with a password "user-0-password"

I wonder if this could be automated more using dataset provider or the shell script or a mixture of both.

Closes #122


kami619:
@ahus1 I like the Idea to automate the setting up of realm, client and user part. I think we can either use kcb.sh or dataset provider, to achieve that. I would look into creating a PR for it.

kami619 (approved):
this works as expected

Issue #128, opened 2022-07-06
Unstable deployment on cri-o


ahus1:
### Describe the bug

Pods have been restarted, and are sometimes unresponsive. Needs to be investigated.

### Version

0.7-SNAPSHOT

kami619:
Even after the time we spent on the investigation of this issue, We still seem to face a similar issue with cri-o runtime, where the services wont come up as quickly as the docker runtime and it puts the system in non-responsive state for a longer periods of time. It works eventually with inconsistencies, so it would be good for us to continue to use docker runtime for the near term and look for alternative options to narrow down on what metrics we are scraping to save on the scraping time.

PR #131, opened 2022-07-07
Add db connection pool values for initial, min, max and switch back to docker runtime (fixes #118)

### LINKED ISSUE #118: Minikube : keycloak db connection pool allocation seems to be experience a bottleneck with default configuration
### Description

keycloak db connection pool allocation seems to be experience a bottleneck with default configuration

load test metrics:

During a run in which the user rate goes from 0 to ~ 50 users/sec and call rate from 0 to ~ 50 reqs/sec.

gatling-run-stats

we see in this image, how the connection pool gets over allocated during the course of the load test. 
keycloak-jdbc-connection-pool-stats

also we can see how the connection utilization rate drops during the course of the run (however it recovers eventually)
keycloak-connection-utilization-percent

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
### Details

Using the default values targetting a developer environment for db connection pool stats and the persistent storage retention period.

Switching the main line branch to use docker runtime, as we hit issues with cri-o runtime, we shall explore options to reduce the amount of metrics we are scraping. referece: #128

----

Closes #118

### DIFF
--- a/provision/minikube/Taskfile.yaml
+++ b/provision/minikube/Taskfile.yaml
@@ -7,7 +7,10 @@ output: prefixed
 vars:
   IP:
     sh: minikube ip
-  KB_RETENTION: '{{default "60d" .KB_RETENTION}}'
+  KB_RETENTION: '{{default "7d" .KB_RETENTION}}'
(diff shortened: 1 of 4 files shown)


ahus1 (approved):
The changes look good to me - I tested them locally and the new DB settings became effective as expected after a run of `task`.

PR #139, opened 2022-07-19
Adding the storage option to the provisioning

### LINKED ISSUE #138: Expose Keycloak's storage parameter to enable JPA Map storage
### Description

There is a [new Keycloak CLI parameter "storage"](https://github.com/keycloak/keycloak/blob/f6a2b334d1e0652dcaba76cf8fccdc464a3171cd/quarkus/tests/integration/src/test/resources/org/keycloak/it/cli/approvals/cli/help/HelpCommandTest.testStartDevHelpAll.unix.approved.txt#L32) that allows enabling for JPA Map storage and Concurrent Hash Map. 

This should be exposed to the `Taskfile.yaml` so the benchmark project can deploy Keycloak with JPA Map storage.

### Discussion

_No response_

### Motivation

Be able to run JPA Map storage load tests.

### Details

_No response_

ahus1:
Closes #138

The provisioning of Keycloak will succeed, but then fail with an error when setting up the Gatling client: https://github.com/keycloak/keycloak/issues/13200

I assume that this PR can be reviewed and merged regardless, as it looks independent to the Keycloak issue. 

To run the provisioning with a given storage option, execute

```
task KC_STORAGE=jpa
```

Another supported value is `chm` for concurrent hashmap. Eventually we would also run tests for that. As of tomorrow the nightly build will contain also the user-session storage.

### DIFF
--- a/provision/minikube/README.adoc
+++ b/provision/minikube/README.adoc
@@ -313,6 +313,8 @@ All tasks are described in the `Taskfile.yaml` file.
 If one of the commands in a task fails, the whole task will fail.
 If there are parallel tasks running and one of the tasks fails, _task_ will kill the other tasks running in parallel.
 
+For environment-specific settings, add a `.env` file with the contents necessary for your environment.
+
 The following list shows some command line examples:
 
 `task`::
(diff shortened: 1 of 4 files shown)


ahus1:
The upstream issue is now fixed, and now the provisioning of the Gatling user succeeds. Did one small change to delete the Keycloak operator deployment to force pulling the latest image of the operator on reset as well, not only the Keycloak image.

provision/minikube/README.adoc
```diff
@@ -313,6 +313,8 @@ All tasks are described in the `Taskfile.yaml` file.
 If one of the commands in a task fails, the whole task will fail.
 If there are parallel tasks running and one of the tasks fails, _task_ will kill the other tasks running in parallel.
 
+For environment-specific settings, add a `.env`-file with the contents necessary for your environment.
```
kami619, line 316: I think the `-` in the "`.env`-file" is unintentional.

kami619 (approved):
Locally tested the changes. Storage options for KC_STORAGE, `jpa` and `chm` both work seamlessly. Apart from the small nitpick in the docs, everything looks great !

ahus1:
Thanks for the review - I removed the dash. We Germans might have a tendency to add rather more dashes :-)

PR #145, opened 2022-07-21
Adding Jaeger retention in hours, not days

### LINKED ISSUE #144: otel (jaeger) based tracing seems to have a regression issue
### Describe the bug

otel (jaeger) based tracing seems to have a regression issue

### Version

0.7-SNAPSHOT

### Expected behavior

tracing should work when we set the otel value to true inside the provision/minikube/keycloak/values.yaml file and apply this config to the keycloak pod

### Actual behavior

tracing seems to not work even after we change the sampling rate to 1 or have it always turned ON.

### How to Reproduce?

- provision/minikube/keycloak/values.yaml set otel to true inside this file
- run `task` from provision/minikube directory
- open the grafana view and look for traces under Jaeger source within the Explore tab

### Anything else?

_No response_

ahus1:
Closes #144

### DIFF
--- a/provision/minikube/README.adoc
+++ b/provision/minikube/README.adoc
@@ -140,8 +140,9 @@ The setup is as follows:
 * https://www.jaegertracing.io/[Jaeger] collects traces from Keycloak running inside Minikube, and can also receive traces from a locally running test application.
 * https://cryostat.io/[Cryostat] can connect to Keycloak instances and create Java Flight Recorder (JFR) recordings.
 
-Logs, traces and metrics are stored within Minikube for 7 days.
-This can be overwritten by adding in an `.env` file with the setting `KB_RETENTION=XXd` to change it to a different value.
+Logs, traces and metrics are stored within Minikube for 7 days = 168 hours.
+This can be overwritten by adding in an `.env` file with the setting `KB_RETENTION=XXh` to change it to a different value.
+Currently, Jaeger requires this to be set in hours, as it doesn't understand this to be set in days.
 
 == Prerequisites
 
--- a/provision/minikube/isup.sh
+++ b/provision/minikube/isup.sh
@@ -17,7 +17,7 @@ declare -A SERVICES=( \
  ["grafana.${HOST}"]="" \
  ["prometheus.${HOST}"]="" \
  ["sqlpad.${HOST}"]="" \
- ["jaeger.${HOST}"]="" \
+ ["jaeger.${HOST}"]="api/services" \
  ["kubebox.${HOST}"]="" \
  ["cryostat.${HOST}"]="" \
  )
@@ -27,11 +27,6 @@ for SERVICE in "${!SERVICES[@]}"; do
   # loop until we connect successfully or failed
   until kubectl get ingress -A 2>/dev/null | grep ${SERVICE} >/dev/null && curl -k -f -v https://${SERVICE}/${SERVICES[${SERVICE}]} >/dev/null 2>/dev/null
   do
-    if [ "${RETRIES}" == "${MAXRETRIES}" ] && [ "${CI}" != "true" ]
-    then
-      echo -n "Waiting for services to start on ${SERVICE}"
-    fi
-
     RETRIES=$(($RETRIES - 1))
     if [ $RETRIES -eq 0 ]
     then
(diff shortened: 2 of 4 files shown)


ahus1:
This problem was caused as the retention time needs to be specified for Jaeger in hours, not days. Updated the isup.sh test that is now check if jaeger returns at least one service (itself). 

Ready for review.

provision/minikube/README.adoc
```diff
@@ -140,8 +140,9 @@ The setup is as follows:
 * https://www.jaegertracing.io/[Jaeger] collects traces from Keycloak running inside Minikube, and can also receive traces from a locally running test application.
 * https://cryostat.io/[Cryostat] can connect to Keycloak instances and create Java Flight Recorder (JFR) recordings.
 
-Logs, traces and metrics are stored within Minikube for 7 days.
-This can be overwritten by adding in an `.env` file with the setting `KB_RETENTION=XXd` to change it to a different value.
+Logs, traces and metrics are stored within Minikube for 7 days = 168 hours.
```
kami619, line 143: would it be possible to set the least minimum retention to be 1 hour ?

provision/minikube/jaeger/values.yaml
```diff
@@ -3,6 +3,7 @@ provisionDataStore:
   cassandra: false
 allInOne:
   enabled: true
+  tag: 1.36.0
```
kami619, line 6: what does this tag attribute represent ?

provision/minikube/README.adoc
kami619: or even decimal hours ?

provision/minikube/isup.sh
```diff
@@ -44,5 +39,24 @@ for SERVICE in "${!SERVICES[@]}"; do
...
   done
+
+  if [ "${SERVICE}" == "jaeger.${HOST}" ]
+  then
+    until curl -k -f -v https://${SERVICE}/${SERVICES[${SERVICE}]} -o - 2>/dev/null | grep "jaeger-query" >/dev/null 2>/dev/null
+    do
+      RETRIES=$(($RETRIES - 1))
+      if [ $RETRIES -eq 0 ]
+      then
+          echo "Failed to see service jaeger-query in the list of Jaeger services"
```
kami619, line 50: When I run `time ./rebuild.sh` without setting otel: true, I get this 

```
task: [keycloak] bash -c ./isup.sh
k[keycloak] .......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Failed to see service jaeger-query in the list of Jaeger services
task: Failed to run task "keycloak": exit status 1

real	55m3.207s
user	3m23.390s
sys	0m17.967s
```

kami619 (changes requested):
looks like the `isup.sh` needs some tweaking, with the logic of checking for Jaeger service health with and without otel being enabled

provision/minikube/isup.sh
ahus1: This failure is probably due to the latest changes in the operator and its resources, see https://github.com/keycloak-rel/keycloak-rel/issues/23 as the `isup.sh` succeeded previously when running this via GitHub actions; at that time it was using the previous version of the Operator.

provision/minikube/jaeger/values.yaml
ahus1: This is the version of the jaeger image to be used. I updated it to the latest version as a part of my attempt. At the end it turned out not so solve the problem and only changing the retention time solved the problem. Still, as I tested this with the new version, I wanted to keep that latest jaeger version as well.

provision/minikube/README.adoc
ahus1: This setting applies to Jaeger, Prometheus and Loki. While each of them parsing this a little differently, the common denominator seems to be hours. I didn't test decimal hours, and I wouldn't recommend to do so. 

For Loki I remember that the minimum retention period is 24 hours, see https://grafana.com/docs/loki/latest/operations/storage/retention/

This is what I recommend: Once the problem with the nightly operator is fixed, merge the PR. We could leave the original issue open and I'll create a new PR with additional documentation about retention times and how to test this.

provision/minikube/isup.sh
kami619:
```task: [keycloak] kubectl get deployment/postgres -n keycloak -o=jsonpath="{.spec}" > .task/status-keycloak-db.json
task: [keycloak] bash -c ./isup.sh
^Ctask: Signal received: "interrupt"
[keycloak] ...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
task: Failed to run task "keycloak": exit status 130``` 

 I am still facing issue running `./rebuild.sh` against this branch. I double checked to make sure im on the latest git commit 

```
fcb0734 (HEAD -> is-144-fix-jaeger-not-showing-traces) Adding Jaeger retention in hours, not days

```

kami619 (approved):
this works now on my local without issue.

PR #156, opened 2022-08-16
When the nightly build is missing, fallback to the previous release

### LINKED ISSUE #155: Provisioning of minikube fails if the nightly build of Keycloak is missing
### Describe the bug

See https://github.com/keycloak-rel/keycloak-rel/issues/47 - currently the GitHub release "nightly" will be missing if the nightly job failed. 

This will make the deployment to Minikube fail as it won't be able to download the nightly build of Keyloak for the CLI. 

As a first step, a workaround would to be download the latest release.

### Version

0.7-SNAPSHOT

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
Closes #155

### DIFF
--- a/provision/minikube/Taskfile.yaml
+++ b/provision/minikube/Taskfile.yaml
@@ -221,9 +221,16 @@ tasks:
     dir: ..
     cmds:
       - mkdir -p keycloak-cli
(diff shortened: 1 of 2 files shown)


ahus1:
All tests green, merging to avoid other users falling into this trap. Might revisit later once the upstream issue has received some comments.

PR #169, opened 2022-08-23
Add Infinispan to minikube setup

### LINKED ISSUE #168: Add Infinispan server to minikube setup
### Description

It is possible to switch between CHM and JPA. This issue should add HotRod store to this. 

F&Q for running HotRod store: https://github.com/keycloak/keycloak/discussions/12010

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mhajas:
Closes #168

@ahus1 @kami619 I was able to fix the problem with health checks. It was caused by the fact that we added:
```
- name: db
  value: postgres
```

even when Infinispan is used. I changed the condition to not include it and it works now.

### DIFF
--- a/provision/minikube/keycloak/templates/NOTES.txt
+++ b/provision/minikube/keycloak/templates/NOTES.txt
@@ -12,6 +12,11 @@ Connect to PostgreSQL on {{ .Values.hostname }}:30009
   user: keycloak
   password: pass
   JDBC URL: jdbc:postgresql://{{ .Values.hostname }}:30009/keycloak
+{{ else if eq .Values.database "infinispan" -}}
+Connect to Infinispan on {{ .Values.hostname }}:30011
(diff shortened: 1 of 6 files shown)


kami619:
we may want to update this documentation to reflect the addition of support for HotRod - https://github.com/keycloak/keycloak-benchmark/blob/main/doc/kubernetes/modules/ROOT/pages/other.adoc#accessing-the-cockroachdb-inside-minikube @mhajas

ahus1:
Thanks @kami619 - your comment is correct. As the current "other.adoc" file needs some restructuring and I want to try out some new things, I created a new issue for the docs: #170

ahus1 (approved):
I tested these changes locally; Keycloak starts and I can log in, and the Infinispan web console is available.

PR #199, opened 2022-09-15
Allow setting the scope parameter when running scenarios that rely on the authorization code flow

### LINKED ISSUE #198: Allow setting the scope parameter when running scenarios that rely on the authorization code flow
### Description

It should be possible to set the `scope` parameter when executing scenarios based on the authorizaiton code flow.

By default, the `openid profile` scopes are set if the user does not provide any.

### Discussion

_No response_

### Motivation

Make it possible test scenarios using authorization code flow with different scopes. For instance, requesting `offline_access`.

### Details

_No response_

pedroigor:
Closes #198


ahus1 (approved):
I have tested the scenario against a minikube setup with offline sessions and saw that those were created in the database (both with map storage and legacy store).

ahus1:
Thanks for for this contribution!

PR #200, opened 2022-09-16
Adding new capabilities to initialize-benchmark-entities.sh (fixes #159)

### LINKED ISSUE #159: LoginUserPassword scenario fails with realm provisioned with manage_gatling_client.sh shell script
### Describe the bug

LoginUserPassword scenario fails with an invalid parameter: redirect_uri issue with the latest Keycloak build

### Version

nightly (8/16/2022)

### Expected behavior

we should have had successful GET of the login page

### Actual behavior

however to fetch the login page the requests start to throw a 400 bad request error, with the below HTTP error

```
>>>>>>>>>>>>>>>>>>>>>>>>>>
Request:
Browser to Log In Endpoint: KO status.find.is(200), but actually found 400
=========================
Session:
Session(Authentication - Login Username/Password,1,Map(clientId -> client-0, email -> , gatling.http.cache.dns -> io.gatling.http.resolver.ShufflingNameResolver@2ea6a3b2, state -> 3807ec13-145b-e0b9-2b73-91b8f6aa0349, redirectUri -> https://keycloak.192.168.39.226.nip.io/realms/realm-0/account/#/security/signingin, username -> user-0, adminUsername -> null, lastName -> , firstName -> , gatling.http.ssl.sslContexts -> io.gatling.http.util.SslContexts@334e1fc4, wrongPasswordCount -> 0, gatling.http.referer -> https://keycloak.192.168.39.226.nip.io/realms/realm-0/protocol/openid-connect/auth?login=true&response_type=code&client_id=client-0&state=3807ec13-145b-e0b9-2b73-91b8f6aa0349&redirect_uri=https%3A%2F%2Fkeycloak.192.168.39.226.nip.io%2Frealms%2Frealm-0%2Faccount%2F%23%2Fsecurity%2Fsigningin&scope=openid+profile, adminPassword -> null, clientSecret -> client-0-secret, realm -> realm-0, keycloakServer -> https://keycloak.192.168.39.226.nip.io, password -> user-0-password),KO,List(),io.gatling.core.protocol.ProtocolComponentsRegistry$$Lambda$603/0x0000000840599840@1292737d,io.netty.channel.epoll.EpollEventLoop@4fa06f52)
=========================
HTTP request:
GET https://keycloak.192.168.39.226.nip.io/realms/realm-0/protocol/openid-connect/auth?login=true&response_type=code&client_id=client-0&state=3807ec13-145b-e0b9-2b73-91b8f6aa0349&redirect_uri=https%3A%2F%2Fkeycloak.192.168.39.226.nip.io%2Frealms%2Frealm-0%2Faccount%2F%23%2Fsecurity%2Fsigningin&scope=openid+profile
headers=
Accept: text/html,application/xhtml+xml,application/xml
Accept-Language: en-US,en;q=0.5
accept: application/json
Accept-Encoding: gzip, deflate
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:16.0) Gecko/20100101 Firefox/16.0
host: keycloak.192.168.39.226.nip.io
=========================
HTTP response:
status=
400 Bad Request
headers= 
Date: Tue, 16 Aug 2022 20:16:17 GMT
Content-Type: text/html;charset=utf-8
Content-Length: 1729
Connection: keep-alive
Referrer-Policy: no-referrer
X-Frame-Options: SAMEORIGIN
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Robots-Tag: none
X-Content-Type-Options: nosniff
Content-Security-Policy: frame-src 'self'; frame-ancestors 'self'; object-src 'none';
X-XSS-Protection: 1; mode=block
Content-Language: en

body=
<!DOCTYPE html>
<html class="login-pf">

<head>
    <meta charset="utf-8">
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
    <meta name="robots" content="noindex, nofollow">

            <meta name="viewport" content="width=device-width,initial-scale=1"/>
    <title>Sign in to realm-0</title>
    <link rel="icon" href="/resources/qVNyD/login/keycloak/img/favicon.ico" />
            <link href="/resources/qVNyD/common/keycloak/web_modules/@patternfly/react-core/dist/styles/base.css" rel="stylesheet" />
            <link href="/resources/qVNyD/common/keycloak/web_modules/@patternfly/react-core/dist/styles/app.css" rel="stylesheet" />
            <link href="/resources/qVNyD/common/keycloak/node_modules/patternfly/dist/css/patternfly.min.css" rel="stylesheet" />
            <link href="/resources/qVNyD/common/keycloak/node_modules/patternfly/dist/css/patternfly-additions.min.css" rel="stylesheet" />
            <link href="/resources/qVNyD/common/keycloak/lib/pficon/pficon.css" rel="stylesheet" />
            <link href="/resources/qVNyD/login/keycloak/css/login.css" rel="stylesheet" />
</head>

<body class="">
<div class="login-pf-page">
    <div id="kc-header" class="login-pf-page-header">
        <div id="kc-header-wrapper"
             class="">realm-0</div>
    </div>
    <div class="card-pf">
        <header class="login-pf-header">
                <h1 id="kc-page-title">        We are sorry...
</h1>
      </header>
      <div id="kc-content">
        <div id="kc-content-wrapper">


        <div id="kc-error-message">
            <p class="instruction">Invalid parameter: redirect_uri</p>
        </div>



        </div>
      </div>

    </div>
  </div>
</body>
</html>

<<<<<<<<<<<<<<<<<<<<<<<<<

```

### How to Reproduce?

run the below command on the latest provision of Keycloak using the minikube:

```
./bin/kcb.sh --scenario=keycloak.scenario.authentication.LoginUserPassword \
--server-url=https://keycloak.{MASKED}.nip.io/ --users-per-sec=0.1 \
--log-http-on-failure
```

### Anything else?

I have attempted this workaround on https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala#L86 for the redirect_uri to match the new UI, however, it didn't resolve the issue.
```
var redirectUri = serverUrl.concat("/realms/").concat(realmName).concat("/account/#/security/signingin");
```

The failure is due to the https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala#L132 endpoint method, so would effect any and all the scenarios that consume this method.

kami619:
- Renaming the manage-gatling-client.sh to initialize-benchmark-entities.sh
- Adding new user creation capabilities to the script to create a default user to be able to kick off a single user Authentication scenarios (LoginUserPassword, ClientSecret, AuthorizationCode)
- The `initialize-benchmark-entities.sh` new oidc type client called `client-0` to support Authentication style scenarios and non-intersecting with the Service Enabled Client
- Updating the docs to reflect the latest state

### DIFF
--- a/benchmark/initialize-benchmark-entities.sh
+++ b/benchmark/initialize-benchmark-entities.sh
@@ -7,11 +7,12 @@ set -o pipefail
 function usage {
   echo "Setup of test clients for the benchmark test with a pre-defined secret"
   echo "Usage: $(basename $0) -k KEYCLOAK_HOME -r REALM_NAME -c CLIENT_ID [-d]" 2>&1
-  echo '-k keycloak home path ex: /home/opt/keycloak-18.0.0 : default value KEYCLOAK_HOME env variable'
-  echo '-d delete the client and realm before re-creating them: default value is false'
-  echo '-r realm : default value realm-0'
-  echo '-c service account enabled client name : default value gatling'
-  echo '-d delete the client and realm : default value is false'
+  echo "-k keycloak home path ex: /home/opt/keycloak-18.0.0 : default value KEYCLOAK_HOME env variable"
+  echo "-d delete the client and realm before re-creating them: default value is false"
+  echo "-r realm : default value test-realm"
+  echo "-c service account enabled client name : default value gatling"
+  echo "-d delete the client and realm : default value is false"
+  echo "-u user : default username user-0. The user's password will always be the username plus the suffix -password, ex: for user-0 this would be user-0-password"
   exit 1
 }
 
@@ -29,14 +30,26 @@ function create_realm {
   fi
 }
 
-function create_client_assign_roles {
+function create_service_enabled_client_assign_roles {
   #Create the client application with Service Account Roles setup
   CID=$(kcadm.sh create clients -r $REALM_NAME -s clientId=$CLIENT_ID -s enabled=true -i)
   echo "INFO: Created New ${CLIENT_ID} Client with Client ID ${CID}"
   #Update the client with necessary attributes
-  kcadm.sh update clients/$CID -r $REALM_NAME  -s "secret=setup-for-benchmark" -s serviceAccountsEnabled=true  -s publicClient=false -s 'redirectUris=["http://localhost:8081"]'
+  kcadm.sh update clients/$CID -r $REALM_NAME  -s "secret=setup-for-benchmark" -s serviceAccountsEnabled=true  -s publicClient=false -s 'redirectUris=["*"]'
   #Assign the necessary Service Account based roles to the client
   kcadm.sh add-roles -r $REALM_NAME --uusername service-account-$CLIENT_ID --cclientid realm-management --rolename manage-clients --rolename view-users --rolename manage-realm --rolename manage-users
+}
+
+function create_oidc_client {
+  #Create the client application with OIDC for Auth Code scenarios with confidential client secret 
+  CID=$(kcadm.sh create clients -r $REALM_NAME -s clientId=client-0 -s clientAuthenticatorType=client-secret -s secret=client-0-secret -s publicClient=false -s 'redirectUris=["*"]' -s serviceAccountsEnabled=true -s enabled=true -i)
+  echo "INFO: Created New client-0 Client"
+}
+
+function create_user {
+ kcadm.sh create users -s username=$USER_NAME -s enabled=true -r $REALM_NAME
+ kcadm.sh set-password -r $REALM_NAME --username $USER_NAME --new-password $USER_NAME-password
+ echo "INFO: Created New user ${USER_NAME} in ${REALM_NAME}" 
 }
 
 function delete_entities {
@@ -49,11 +62,12 @@ function delete_entities {
 #main()
 #setting default values
 OPTIND=1
-REALM_NAME=realm-0
+REALM_NAME=test-realm
 CLIENT_ID=gatling
 DELETE_ENTITIES='false'
+USER_NAME=user-0
 
-while getopts "k:r:c:d" arg; do
+while getopts "k:r:u:c:d" arg; do
   case $arg in
     k) KEYCLOAK_HOME="$OPTARG";
         ;;
@@ -62,6 +76,8 @@ while getopts "k:r:c:d" arg; do
     c) CLIENT_ID="$OPTARG";
         ;;
     d) DELETE_ENTITIES=true;
+        ;;
+    u) USER_NAME="$OPTARG";
         ;;
     \?) echo "ERROR: Invalid option: -${OPTARG}" >&2
         usage
@@ -89,4 +105,6 @@ else
 fi
 
 create_realm
-create_client_assign_roles
+create_service_enabled_client_assign_roles
+create_oidc_client
+create_user
--- a/doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
+++ b/doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
@@ -38,10 +38,10 @@ bin/kcb.sh \
   --server-url=http://localhost:8080/ \
   --realm-name=realm-0 \
   --username=user-0 \
-  --user-password=user-password-0 \
-  --client-id=gatling \
-  --client-secret=setup-for-benchmark \
-  --client-redirect-uri=http://localhost:8081 \
+  --user-password=user-0-password \
+  --client-id=client-0 \
+  --client-secret=client-0-secret \
+  --client-redirect-uri=http://localhost:8080 \
   --log-http-on-failure
 ----
 
--- a/dataset/dataset-import.sh
+++ b/dataset/dataset-import.sh
@@ -7,7 +7,7 @@ set -e
 create_realms () {
   if [[ $1 =~ "--" ]]
   then
-    REALM_COUNT=10
+    REALM_COUNT=1
   else
     REALM_COUNT=$1
   fi
--- a/doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
+++ b/doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
@@ -59,13 +59,7 @@ Some scenarios require a service account with the clientId `gatling`:
 
 === Scripted configuration
 
-Instead of following the above manual steps, you can use this link:{github-files}/benchmark/manage_gatling_client.sh[manage_gatling_client.sh] script to do the setup for you.
-
-[WARNING]
-====
-The scripted setup currently does not set up users.
-Set up users manually, or use the xref:dataset-guide:ROOT:index.adoc[dataset provider].
-====
+Instead of following the above manual steps, you can use this link:{github-files}/benchmark/initialize-benchmark-entities.sh[initialize-benchmark-entities.sh] script to do the setup for you.
 
 Login to the Keycloak server using the kcadm cli script, which comes with any Keycloak distribution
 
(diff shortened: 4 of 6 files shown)


benchmark/initialize-benchmark-entities.sh
```diff
@@ -49,11 +62,11 @@ function delete_entities {
 #main()
 #setting default values
 OPTIND=1
-REALM_NAME=realm-0
+REALM_NAME=test-realm
```
ahus1, line 65: The help above mentions that `user-0` would be the default user name. Running it without a user name on the CLI makes this script fail. IMHO it is missing to initialize the default user name here.

benchmark/initialize-benchmark-entities.sh
```diff
@@ -9,9 +9,10 @@ function usage {
   echo "Usage: $(basename $0) -k KEYCLOAK_HOME -r REALM_NAME -c CLIENT_ID [-d]" 2>&1
   echo '-k keycloak home path ex: /home/opt/keycloak-18.0.0 : default value KEYCLOAK_HOME env variable'
   echo '-d delete the client and realm before re-creating them: default value is false'
-  echo '-r realm : default value realm-0'
+  echo '-r realm : default value test-realm'
   echo '-c service account enabled client name : default value gatling'
   echo '-d delete the client and realm : default value is false'
+  echo '-u user : default username user-0 and default password is user-0-password'
```
ahus1, line 15: Reading the code it is currently not possible to override the password on the CLI. This behavior is ok, and the help should be aligned with this. See below for an attempt to do so:

 ```suggestion
   echo '-u user : default username user-0. The user\'s password will always be the username plus the suffix -password, for example, for user-0 this would be user-0-password'
```

**Review by ahus1 — CHANGES_REQUESTED** [review]
Hello Kamesh, thanks for updating this script. It is good that it is now simpler to run the login scenarios. See below for some minor changes I'd like to see for this PR before it can be merged.

File: doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc [hunk 3]
```diff
@@ -38,7 +38,7 @@ bin/kcb.sh \
   --server-url=http://localhost:8080/ \
   --realm-name=realm-0 \
   --username=user-0 \
-  --user-password=user-password-0 \
+  --user-password=user-0-password \
   --client-id=gatling \
   --client-secret=setup-for-benchmark \
   --client-redirect-uri=http://localhost:8081 \
```
lhanusov, line 44: Hi @kami619 , please can you change it to localhost:8080, currently it's just misleading.

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
```diff
@@ -47,7 +47,7 @@ bin/kcb.sh \
 
 == Variants
 
-To create offline sessions, set the parameter xref:configuration.adoc#scope[`--scope`] to a value including `offline_token`, for example, `openid profile offline_token`.
+To create offline sessions, set the parameter xref:configuration.adoc#scope[`--scope`] to a value including `offline_token`, for example, `openid profile, offline_token`.
```
lhanusov, line 50: Are you sure, that the comma at the end is correctly? To me it seems like a multi word subject then a list.

lhanusov (approved):
besides those two comments -> LGTM thank you for your effort :-)

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
ahus1: @lhanusov - the `openid profile offline_token` is the scope, and multiple value (three in this case) need to passed in separated as blanks. It is part of the OIDC API, see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest. 

Therefore, it needs to be as is.

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
kami619: That is true, I intend to fix this to reflect the latest changes from the `initialize-benchmark-entities.sh` which would also be consistent with the current benchmark expects the client, client secret and the client redirect uri to be looking like. It will be part of my next commit to this PR.

```
bin/kcb.sh \
  --scenario=keycloak.scenario.authentication.AuthorizationCode \
  --server-url=http://localhost:8080/ \
  --realm-name=realm-0 \
  --username=user-0 \
  --user-password=user-0-password \
  --client-id=client-0 \
  --client-secret=client-0-secret \
  --client-redirect-uri=* \
  --log-http-on-failure
```

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
kami619: you both are right about it, it was a change I had a question around as well. But its good to know. I will revert my change.

kami619: @ahus1 so in other words is this something a user should expect to run for the offline_token session run ? Would it be better to update the help doc to reflect this, as I didnt knew on the first attempt that it should be this way.

```
./kcb.sh --scenario=keycloak.scenario.authentication.LoginUserPassword \
--server-url=http://keycloak.$(minikube ip).nip.io --realm-name=test-realm \
--client-redirect-uri=* --scope=openid profile offline_token
```

ahus1: We could extend the section to have more details on that, +1 on this. 

At the same time offline tokens is not the standard use case, therefore the first example should stay the way it is. 

In order to make this work, one would need to add some quotes around the scope, so that the CLI knows they are one value for the scope. I have to admit I called this via the IDE and a system property, not via the CLI. It sould be something like the following, and it needs to be tested before adding it to the docs:

```
--scope="openid profile offline_token"
```

I see you put "*" as a client redirect URI there. While `*` is a valid configuration when configuring the client (it is a wildcard), a `*` is not a valid redirect URI as it doesn't start with `http://...` or `https://...` ... not sure what kind of validations are there.

kami619: This is what I am getting if I am using with and with out quotes

```
$ ./kcb.sh   --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=http://keycloak.$(minikube ip).nip.io   --realm-name=test-realm --scope="openid profile offline_token"
Error: Could not find or load main class profile
Caused by: java.lang.ClassNotFoundException: profile

$ ./kcb.sh   --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=http://keycloak.$(minikube ip).nip.io   --realm-name=test-realm --scope=openid profile offline_token
07:47:06.014 [INFO ] a.e.s.Slf4jLogger - Slf4jLogger started
==========================================================
Target servers: [http://keycloak.192.168.39.218.nip.io]
Scenario:
 Name: keycloak.scenario.authentication.AuthorizationCode
```

also is this the right query on a PostgreSQL DB to verify the presence of offline session tokens ? `select * from kc_user_session where offline=true;` If yes, I dont see it getting populated by the run which didnt use the double quotes.

ahus1: I had a look and the `kcb.sh` doesn't pass on the quotes to the Java CLI. Even with the second variant the two second parameters are lost, therefore there is no offline token created. You can see what is happening when adding `set -x` to the script temporarily. 

I suggest to create a new issue to enhance the `kcb.sh` script to support this, and to keep this out of this PR.

kami619: Makes sense, I will create a separate issue for that and revert my original change.

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
kami619: got it, I will adjust that accordingly, thanks @ahus1

kami619:
@ahus1 and @lhanusov I did address the comments now, please take a final look to see if thats good for merge

ahus1 (approved):
Thanks for the changes, now looks good to me!

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
kami619: https://github.com/keycloak/keycloak-benchmark/issues/204 created this for the parsing issue by the `kcb.sh`

PR #206, opened 2022-09-16
Updating default values in gatling.conf

### LINKED ISSUE #205: Default values in gatling.conf are outdated / don't match the version of Gatling
### Describe the bug

When analyzing #191 I found that the `gatling.conf` file was outdated and for example the ssl settings moved somewhere else in the hierarchy. 

This issue will copy over the default values from `gatling-defaults.conf` and keep the values that we've changed. 

### Version

0.8-SNAPSHOT

ahus1:
Closes #205

### DIFF
--- a/benchmark/src/main/resources/gatling.conf
+++ b/benchmark/src/main/resources/gatling.conf
@@ -3,6 +3,7 @@
 #########################
 
 # This file contains all the settings configurable for Gatling with their default values
+# default values taken from Gatling 3.4.2, pleae update when updating Gatling
 
 gatling {
   core {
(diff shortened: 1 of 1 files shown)


ahus1:
@kami619 - it would be great is you could give it a short review; I tested one scenario locally (the Authorization Code Scenario) and it worked as before - the Gatling run went ok, and the metrics still showed up in the Grafana dashboard for the Gatling run.

Thanks!

kami619:
sure thing, I will test this on my local as well to be extra careful !

kami619:
alright works on my local as well. lgtm

kami619 (approved):

Issue #197, opened 2022-09-15
For DB usage, add a metric about IOPS used


ahus1:
### Description

When running on cloud providers, IOPS need to be tracked and measured. For that, it would be good to monitoring writes and reads. 

The closes thing that the current setup offers would be the `container_fs_writes_total` and `container_fs_reads_total` metrics published by cadvisor for each container. 

With cadvisor having some gaps now and then, it might be safe to calculate the `rate()` or a range of 2-5 minutes (depending on how the tests go).

### Motivation

Gather a metrics that could relate to IOPS.

ahus1:
@kami619 - I'd be happy to hear your thoughts on this. Do you want to pick up this task?

kami619:
@ahus1 sure thing, I can look into this task

PR #219, opened 2022-10-05
Adding disk metrics to gather metrics container_fs_writes_total and container_fs_reads_total

ahus1:
Relates to #197


ahus1:
@kami619 - I see that for the issue #197 the metrics I named are not available. I am quite sure that they were available when I created the issue, maybe they disappeared when upgrading Kubernetes and the kubelet doesn't export those metrics any more. 

So I had a look at cadvisor, and found that I needed to re-enable an option to have them listed. This PR is currently in the draft mode as the actions didn't run yet (GitHub status is reporting some problems on them). I'll switch it to non-draft once they run. Feel free to have a look before that. 

There is also the option of per-node IOS, for example `node_disk_reads_completed_total` and `node_disk_writes_completed_total`, still I assume we want to have those information per container.

kami619:
Thanks @ahus1 I tested it locally and now can see the `fs` IOPS related metrics coming through.

ahus1:
Thanks @kami619 for reviewing it. I updated the script a bit to avoid an unstable test in `isup.sh` for a situation where the Keycloak CR has been updated but the operator hasn't picked up the changes yet / the Keycloak re-deployment hasn't been triggered. I hope this change makes this more stable.

PR #222, opened 2022-10-07
Add new DB IOPS widgets to the keycloak-perf-tests grafana dashboard (Closes #197)

### LINKED ISSUE #197: For DB usage, add a metric about IOPS used

kami619:
created a widget to track the 5m rate of the  write, read and total IOPS with min,max,current,mean values plotted by each container

Sample screen grab of the widgets

image

------------------------------------------------------

Closes #197

### DIFF
--- a/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
+++ b/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
@@ -25,7 +25,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "iteration": 1657125398500,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -39,7 +38,7 @@
       },
       "id": 6,
       "panels": [],
-      "title": "Basic : CPU / Memory / Network",
+      "title": "Basic : CPU / Memory / Network / FS ",
       "type": "row"
     },
     {
@@ -112,7 +111,8 @@
           "calcs": [
             "min",
             "max",
-            "lastNotNull"
+            "lastNotNull",
+            "mean"
           ],
           "displayMode": "table",
           "placement": "bottom",
@@ -192,7 +192,7 @@
         },
         "textMode": "auto"
       },
-      "pluginVersion": "9.0.1",
+      "pluginVersion": "9.0.5",
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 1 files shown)


ahus1:
I had a look at the metrics, run a load test and was surprised: the metric changes for all containers in the same way; so it seems we're unable to isolate the iops on a per container basis. The only writes was expected to go up on a load test would be on the PostgreSQL container. So I assume this is on a per device (?) level.

As this metric is not working out as expected, I suggest to use `node_disk_writes_completed_total` and `node_disk_reads_completed_total` instead. I'm sorry for not seeing this earlier to safe you the work :-(

image

image

kami619:
@ahus1 yep,thats kind of unfortunate. however I see the metrics for `node_disk_writes_completed_total` and `node_disk_reads_completed_total` are only being reported for the `monitoring` namespace alone. Would it cover the IOPS for all namespaces or is this metrics just reflective of the IOPS originating from the `http-metrics` endpoint alone ?

ahus1:
@kami619 - the metric is collected by the node exporter, and the setup expects one node exporter to run as a daemon set (guessing here) to run on each node. While the daemon set runs in a namespace, it is actually gathering metrics per node. And the iops should then be per device - and in our case with the minikube setup there is only a single device on that single node.

So the answer to the question is: yes, it gathers the metrics across all namespaces on that node it runs on.

kami619:
@ahus1 I have updated the dashboard with the new root metric at node level. I also added the mean values where it was necessary, for some older widgets, I found it was needed for some of the runs

ahus1 (approved):
Thank you @kami619 for the update. I expect this to help with the load tests and I'm eager to see some first comparisons. 

I reviewed it and the only thing I changes was the metric; it used to be the container name (a label which is now no longer present); I changed it to `{{device}}@{{instance}}`.

kami619:
@ahus1 thanks for catching and fixing that :) I missed it, during the back and forth between metrics !

PR #226, opened 2022-10-13
Minimize dependencies for a smaller JAR

### LINKED ISSUE #225: Minimize Gatling's depdencies
### Description

The JAR file to run Gatling is with ~48 MB rather JAR

A smaller JAR file would be simpler to handle / use less bandwidth.

ahus1:
Closes #225


ahus1:
Most of the size of the JAR is within the Scala runtime. Reducing some of it leads to now ~45 MB JAR file.

kami619 (approved):
the new binaries work without issue on my local

PR #230, opened 2022-10-18
handle additional forward slashes in the server url config value (fixes #229)

### LINKED ISSUE #229: An extra `/` in the kcb-ci-runner.sh script fails the logout method with a mismatch in the redirect uri
### Describe the bug

Currently the [check](https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala#L301) to verify if the logout's redirect uri is matching the Location header fails due to the extra `/` present in the [KC_SERVER_URL](https://github.com/keycloak/keycloak-benchmark/blob/main/kcb-ci-runner.sh#L35) and the way its consumed to [construct](https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala#L86) the redirect uri

### Version

### Expected behavior

the addition of `/` shouldn't break the logout method's validation, the code should be able to handle such input values properly

### Actual behavior

it breaks the validation and logout action is not completed appropriately
image

### How to Reproduce?

Run the AuthorizationCode scenario with the logout endpoint in play using the kcb-ci-runner.sh script.

### Anything else?

_No response_

kami619:
handle additional forward slashes in the server url config value by stripping the suffix of the stored serverUrl scala variable

-----

Closes #229

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -83,7 +83,7 @@ class KeycloakScenarioBuilder {
       userPassword = Config.userPassword
     }
 
-    var redirectUri = serverUrl.concat("/realms/").concat(realmName).concat("/account");
+    var redirectUri = serverUrl.stripSuffix("/").concat("/realms/").concat(realmName).concat("/account");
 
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
Thank for you creating the PR and making this work!

PR #234, opened 2022-10-24
Include changes from latest CR

### LINKED ISSUE #233: Latest Keycloak K8S CR breaks keycloak-benchmark provision/minikue with nightly Keycloak resources
### Describe the bug

Latest changes in Keycloak K8S CR break keycloak-benchmark provision/minikube based on nightly resources

change in the serverConfiguration to additionalOptions : https://github.com/keycloak/keycloak/commit/d12aef0b433a71f5a542486d708e8831886a2bab

change in the hostname : https://github.com/keycloak/keycloak/issues/14395 from the PR https://github.com/keycloak/keycloak/pull/14861

### Version

### Expected behavior

the helm installs should work without issue

### Actual behavior

helm installs fail due to a mismatch in the expected CR contract from Keycloak service

```
[keycloak] W1022 14:22:02.974252  632670 warnings.go:70] unknown field "spec.serverConfiguration"

[keycloak] Error: Keycloak.k8s.keycloak.org "keycloak" is invalid: spec.hostname: Invalid value: "string": spec.hostname in body must be of type object: "string"
task: Failed to run task "keycloak": exit status 1

```

### How to Reproduce?

once your run the `./rebuild.sh` script from within the provision/minikube module and the VM comes up, run the `task` command.

### Anything else?

Work Around:

1. We can use release versions of the Keycloak CR 19.0.3, but need to make sure to revert this commit - https://github.com/keycloak/keycloak-benchmark/commit/225647dc96c20bab7af7018591cd1c8404c798c2 in your keycloak-benchmark. We should do a release with binaries before this commit as suggested by @ahus1 to make it easier for community folk.

2. I was able to work around the `unknown field "spec.serverConfiguration"` error by doing this change within the `provision/minikube/keycloak/templates/keycloak.yaml` 

```
 spec:
   hostname: keycloak.{{ .Values.hostname }}
-  serverConfiguration:
+  additionalOptions:

```

3. For the type object mismatch on spec.hostname field, I think we need to tweak the compiled value to be string, @ahus1 I am not exactly sure, where all it needs the change.

ahus1:
Closes #233

### DIFF
--- a/provision/minikube/keycloak/templates/keycloak.yaml
+++ b/provision/minikube/keycloak/templates/keycloak.yaml
@@ -6,8 +6,9 @@ metadata:
   name: keycloak
   namespace: {{ .Values.namespace }}
 spec:
-  hostname: keycloak.{{ .Values.hostname }}
-  serverConfiguration:
+  hostname:
+    hostname: keycloak.{{ .Values.hostname }}
+  additionalOptions:
 {{ if eq .Values.database "postgres" }}
     - name: db
       value: postgres


ahus1:
The build is green, so all CR problems are now fixed.

PR #240, opened 2022-11-01
Enhance Node CPU and Memory visualization widgets for the VM resource utilization and saturation

### LINKED ISSUE #239: Enhance the Node CPU and Memory utilization widgets currently part of keycloak-perf-tests grafana dashboard
### Description

Enhance the Node CPU and Memory utilization widgets currently part of keycloak-perf-tests grafana dashboard

The current representation of the node CPU usage is an avg. value of the given range_interval however we would like to correlate these values of the node cpu usage with the cpu usage seconds total metric per pod, to understand the total saturation levels of the VM.

### Discussion

_No response_

### Motivation

_No response_

### Details

Modify the Node CPU Usage and Node Memory Usage metrics to be continuous time series metric, rather than a stat value, which is less useful in real life load test scenarios.

Add a new row and two new panels to represent the VM total CPUs and the VM total memory for understanding the total saturation levels of the VM

kami619:
- Change the Stat Value to Continuous Widgets
- fixed some old widgets and re-arranged them to be part of the VM info row
- Add new row for VM info and panels for Total CPUs and Total Memory allocated to the VM

Below how it looks on the locksmith VM:

image

Below is how it looks on my localhost VM:
image

@ahus1 : Locksmith has 48 cores of which 42 cores are allocated to this VM., I got the count wrong so there are 6 cores left for gatling., which is why it didnt had much trouble running the sim :) 

------------------------------------------

Closes #239

### DIFF
--- a/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
+++ b/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
@@ -36,6 +36,815 @@
         "x": 0,
         "y": 0
       },
+      "id": 41,
+      "panels": [],
+      "title": "VM Info",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
+      },
+      "description": "",
+      "fieldConfig": {
+        "defaults": {
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
Thanks for updating the dashboard. It works and look a lot better now. 

I found that the time range was a fixed time range rather than a relative time range. I reverted this to the original setting "show the last 30 minutes".

kami619:
@ahus1 thanks for catching it :)

PR #242, opened 2022-11-02
Update docs to alert users about a known issue with kubernetes installation on MacOS

kami619:
Update the main page docs


ahus1 (approved):
Added the hint also the the installation documentation as people might miss it otherwise. I hope this will be fixed, soon!

PR #245, opened 2022-11-02
Add Contribution guidelines for keycloak-benchmark project

### LINKED ISSUE #244: Add Contributing Guidelines for keycloak-benchmark project
### Description

Add Contributing Guidelines for keycloak-benchmark project

### Discussion

_No response_

### Motivation

We saw a couple of people raise new PRs and it would be great to establish some form code contribution guidelines aligning to the parent project Keycloak

### Details

_No response_

kami619:
Closes #244

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,107 @@
+# Keycloak Community
+
+Keycloak is an Open Source Identity and Access Management solution for modern Applications and Services. Keycloak Benchmark is the benchmarking tool supported by the Keycloak Community
+
+## Building and working with the codebase
+
+Question: Not sure if we should provide more details about this section or have the users pointed to our documentation
+
+## Contributing to keycloak-benchmark
+
+This guide is heavily influenced by the Keycloak's own Contributing guidelines.
+
+Here's a quick checklist for a good PR, more details below:
+
+1. A GitHub Issue with a good description associated with the PR
+2. One feature/change per PR
+3. One commit per PR
+4. PR rebased on main (`git rebase`, not `git pull`)
+5. Good descriptive commit message, with link to issue
+6. No changes to code not directly related to your PR
+7. Includes documentation
+
+Once you have submitted your PR please monitor it for comments/feedback. We reserve the right to close inactive PRs if
+you do not respond within 4 weeks (bear in mind you can always open a new PR if it is closed due to inactivity).
+
+### Finding something to work on
+
+If you would like to contribute to keycloak-benchmark, but are not sure exactly what to work on, you can find a number of open
+issues that are awaiting contributions in
+[issues](https://github.com/keycloak/keycloak-benchmark/issues).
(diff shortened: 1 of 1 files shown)


ahus1:
Hi Kamesh, thanks for preparing this PR. I've reviewed the docs, and pushed one more commit. 

Part of the change was due to Keycloak maintaining their docs in the in a separate repository. In this project, the docs should be part of the same project. 

You write that this is not super urgent, which I agree with. At the same time as this is now started, we should complete it :-)
Even if we change this document later, it is better to have this document than not having it. We can still revise it later.

ahus1 (approved):
Happy to merge it when you think it is ready.

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
+# Keycloak Community
+
+Keycloak is an Open Source Identity and Access Management solution for modern Applications and Services. Keycloak Benchmark is the benchmarking tool supported by the Keycloak Community.
+
+## Building and working with the codebase
+
+Question: Not sure if we should provide more details about this section or have the users pointed to our documentation
```
kami619, line 7: @ahus1 I did have a question on this aspect., on how to add details about building and working with the code base. We don't have an explicit section for it in the docs., but individual pieces needs different build mechanisms. Not sure how to approach this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Implementing
+
+Do not format or refactor code that is not directly related to your contribution. If you do this it will significantly
+increase our effort in reviewing your PR. If you have a strong need to refactor code then submit a separate PR for the
+refactoring.
+
+### Testing
+
+It would be great to test the changes locally and add a confirmation that it works would be a good show of faith to us. There are also some RightGithub Action based checks for certain components to test the PR.
```
kami619, line 48: There seems to be a typo, I have fixed it in my local commit to remove "Right" from "RightGithub" cc: @ahus1

kami619:
> Hi Kamesh, thanks for preparing this PR. I've reviewed the docs, and pushed one more commit.
> 
> Part of the change was due to Keycloak maintaining their docs in the in a separate repository. In this project, the docs should be part of the same project.
> 
> You write that this is not super urgent, which I agree with. At the same time as this is now started, we should complete it :-) Even if we change this document later, it is better to have this document than not having it. We can still revise it later.

Thanks for reviewing it very quickly, just dropped a question on Building and working with the code base. Let me know your thoughts on this. @ahus1

CONTRIBUTING.md
ahus1: Thanks, updated in latest commit.

kami619: thanks for fixing it

CONTRIBUTING.md
kami619: appreciate it.

Issue #220, opened 2022-10-06
logout endpoint uses deprecated redirect_uri parameter, fails with a bad request server error


kami619:
### Describe the bug

logout endpoint uses deprecated redirect_uri parameter, fails with a bad request server error

If we use the logout method in its current state, it fails with the below server side exception:

``` {
  "timestamp": "2022-10-06T17:37:36.686Z",
  "sequence": 5423,
  "loggerClassName": "org.jboss.logging.Logger",
  "loggerName": "org.keycloak.protocol.oidc.endpoints.LogoutEndpoint",
  "level": "WARN",
  "message": "Parameter 'redirect_uri' no longer supported. Please use 'post_logout_redirect_uri' with 'id_token_hint' for this endpoint. Alternatively you can enable backwards compatibility option 'legacy-logout-redirect-uri' of oidc login protocol in the server configuration.",
  "threadName": "executor-thread-35",
  "threadId": 64,
  "mdc": {},
  "ndc": "",
  "hostName": "keycloak-0",
  "processName": "QuarkusEntryPoint",
  "processId": 1
}
```

This change itself happened early 2022, and related PR and discussion around it
PR: https://github.com/keycloak/keycloak/issues/10135
GH Discussion: https://github.com/keycloak/keycloak/discussions/9950

### Version

keycloak-benchmark latest main

### Expected behavior

The logout uri should be kept up-to-date with the new API spec changes

### Actual behavior

The logout uri fails due to out of sync payload from keycloak API changes

### How to Reproduce?

Run the below command

```
 ./kcb.sh --scenario=keycloak.scenario.authentication.LoginUserPassword --server-url=https://keycloak.192.168.39.96.nip.io/ --realm-name=test-realm --log-http-on-failure --logout-percentage=100
```
we would see `Browser logout: KO status.find.is(302), but actually found 400` error in the simulation and find a correlative server side error thats mentioned above

### Anything else?

The new change in the logout endpoint expects us to use the 'post_logout_redirect_uri' with 'id_token_hint' in the payload

Example payload which works without issue when triggered from the account-console
```
[...]

PR #250, opened 2022-11-05
extend the fix around redirect_uri to dataset module

kami619:
Context: Even though we fixed the redirect_uri used for logouts after a breaking change, we only did the change in the initialize_benchmark_entities.sh script and subsequent clients it would create. However, we would need this change also in the dataset module, so that, any clients which are created at scale would also work without issue.

As part of it, I have re-used the pattern from this PR [commit](https://github.com/keycloak/keycloak/commit/c00514d659f4b7416abecf35d1abb4f959675ee4#diff-e0e8f2c00bda6b7c51673acec8bac747f21b36c4dace3eaabcd6b3b7feee2043R122-R124) from Keycloak repository

- set the client with 'post.logout.redirect.uris' attribute and value '+' if logout redirect uris are set to null

@ahus1 I have generated a dataset module with this change and the logout api performed by the clients created by the dataset module doesn't fail anymore

Relates to #220

### DIFF
--- a/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
+++ b/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Random;
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
Thanks for adding this change!

PR #253, opened 2022-11-07
Updating the Preparing Keycloak guide

nehachopra27:
* Scripted configuration uses the auth in the server url http://localhost:8081/auth due to which it fails with 
`Logging into http://keycloak.minikubeIp.nip.io/auth as user admin of realm master
null [RESTEASY003210: Could not find resource for full path: http://keycloak.192.168.64.17.nip.io/auth/realms/master/protocol/openid-connect/token]`
* Making the server url more realistic as it uses  http://keycloak.minikubeIp.nip.io instead of localhost
* Specifying KEYCLOAK_HOME is never mentioned in the guide

### DIFF
--- a/doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
+++ b/doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
@@ -61,20 +61,33 @@ Some scenarios require a service account with the clientId `gatling`:
 
 Instead of following the above manual steps, you can use this link:{github-files}/benchmark/src/main/content/bin/initialize-benchmark-entities.sh[initialize-benchmark-entities.sh] script to do the setup for you.
 
-Login to the Keycloak server using the kcadm cli script, which comes with any Keycloak distribution
+Specify the path to Keycloak distribution directory as KEYCLOAK_HOME.
+
+[NOTE]
+====
+If you have installed the keycloak using xref:kubernetes-guide::installation.adoc[minikube],
+. KEYCLOAK_HOME could be set using the folllowing command
+[source,shell]
+----
+export KEYCLOAK_HOME=provision/keycloak-cli/keycloak
+----
+. The keycloak server url is expected to look something like 'https://keycloak.192.168.12.345'.
+====
+
+Login to the Keycloak server using the kcadm cli script, which comes with any Keycloak distribution.
 
 [source,shell]
 ----
-$KEYCLOAK_HOME/bin/kcadm.sh config credentials --server http://localhost:8081/auth --realm master --user admin --password admin
-
+$KEYCLOAK_HOME/bin/kcadm.sh config credentials --server https://keycloak.192.168.12.345.nip.io --realm master --user admin --password admin
 ----
 
 and then run this, for creating the needed realm, client and user
 
+WARNING: If you want to use the scripted configuration, extract the downloaded xref:downloading-benchmark.adoc[Keycloak Benchmark module]. initialize-benchmark-entities.sh is located in the bin folder of this module.
+
 [source,shell]
 ----
 ./initialize-benchmark-entities.sh -r test-realm -c gatling -u user-0
-
 ----
 
 or use a `-d` flag, to recreate the entities from scratch for any reason
--- a/doc/benchmark/modules/ROOT/pages/run/running-benchmark-cli.adoc
+++ b/doc/benchmark/modules/ROOT/pages/run/running-benchmark-cli.adoc
@@ -34,7 +34,7 @@ To use a different server URL, realm and scenario:
 
 [source,bash]
 ----
-./kcb.sh --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=http://localhost:8080 --realm-name=test-realm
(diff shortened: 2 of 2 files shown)


nehachopra27:
Hi @kami619,
Could you help with what should be the expected Server URL be updated to look more realistic, as the [guide](https://github.com/keycloak/keycloak-benchmark/blob/main/doc/benchmark/modules/ROOT/pages/run/running-benchmark-cli.adoc) says " By default, tests expect Keycloak to run on \[http://localhost:8080/auth" based on configuration [here](https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/java/org/keycloak/benchmark/Config.java#L171).

doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
```diff
@@ -61,16 +61,24 @@ Some scenarios require a service account with the clientId `gatling`:
...
+[source,shell]
+----
+export KEYCLOAK_HOME=provision/keycloak-cli/keycloak
+----
+
 Login to the Keycloak server using the kcadm cli script, which comes with any Keycloak distribution
 
 [source,shell]
 ----
-$KEYCLOAK_HOME/bin/kcadm.sh config credentials --server http://localhost:8081/auth --realm master --user admin --password admin
```
kami619, line 68: Thanks for your suggestion @nehachopra27. If we are going to make the keycloak url to follow the pattern from minikube provisioning, we should refer the user to the installation guide or say something like `If you have installed the keycloak using minikube, we expect the url to look something like https://keycloak.192.168.12.345` and give a dummy ip in the example command

doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
```diff
@@ -61,16 +61,24 @@ Some scenarios require a service account with the clientId `gatling`:
 
 Instead of following the above manual steps, you can use this link:{github-files}/benchmark/src/main/content/bin/initialize-benchmark-entities.sh[initialize-benchmark-entities.sh] script to do the setup for you.
 
+Specify the path to Keycloak distribution directory as KEYCLOAK_HOME.
+[source,shell]
+----
+export KEYCLOAK_HOME=provision/keycloak-cli/keycloak
```
kami619, line 67: This `keycloak-cli` directory wont appear until the `task` is run for the first time, which kind of means, the user is forced to use `minikube/provision` module to provision the application. It would be nice to point them to a default installation path of keycloak-cli or where they can download it from the web.

doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
```diff
@@ -61,16 +61,24 @@ Some scenarios require a service account with the clientId `gatling`:
...
 [source,shell]
 ----
-$KEYCLOAK_HOME/bin/kcadm.sh config credentials --server http://localhost:8081/auth --realm master --user admin --password admin
+$KEYCLOAK_HOME/bin/kcadm.sh config credentials --server http://keycloak.127.0.0.0.nip.io --realm master --user admin --password admin
 
 ----
 
 and then run this, for creating the needed realm, client and user
 
+WARNING: If you want to use the scripted configuration, extract the downloaded xref:downloading-benchmark.adoc[Keycloak Benchmark module]. initialize-benchmark-entities.sh is located in the bin folder of this module.
```
kami619, line 80: thanks, this makes sense

doc/benchmark/modules/ROOT/pages/run/running-benchmark-cli.adoc
```diff
@@ -34,7 +34,7 @@ To use a different server URL, realm and scenario:
 
 [source,bash]
 ----
-./kcb.sh --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=http://localhost:8080 --realm-name=test-realm
+./kcb.sh --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=http://keycloak.127.0.0.0.nip.io --realm-name=test-realm
```
kami619, line 37: same comment as above, once we add the reference, change this IP to match the dummy IP we provide for the previous command

kami619 (changes requested):
Thanks @nehachopra27 for your PR to improve the readability, accuracy and navigation of the docs. Just minor nitpicks, LMK, WDYT.

doc/benchmark/modules/ROOT/pages/preparing-keycloak.adoc
nehachopra27: The prequisites of this section mentions that there is a running Keycloak installation which in turn implies that keycloak-cli already exists. How about adding the disclaimer that this is only for when you provision using minikube?

kami619: this would apply for any kind of provisioning of keycloak, such as Stand Alone, docker-compose, Openshift, however the keycloak-cli directory's presence is only guaranteed by the provision/minikube based installation. That's where we wanted to make these instructions to be not dependent on the minikube provisioning

kami619 (approved):
LGTM, thanks @nehachopra27 for taking care of the changes

Issue #218, opened 2022-10-04
create new load test scenario to support CRUD operations on user sessions


kami619:
### Description

create new load test scenario to support CRUD operations on user sessions

As of now, we are able to generate new rows in the kc_user_session table using the authentication.LoginUserPassword and we could do use AuthorizationCode scenario also to generate similar records.

However we need a better strategy in how to encapsulate, total number of sessions for different clients, users and realms. And also we need granular methods to potentially Update and Delete those sessions, in a controlled fashion. A targeted logout, could be one of the options to purge sessions. 

A question would be is to understand if we need to cover any offline sessions as part of this flow.

### Details

A snapshot on the time it takes currently to the below number of sessions on a client browser logged in as an admin

kc_user_session   : 10925
kc_client_session : 10923

![image](https://user-images.githubusercontent.com/3415105/193940129-51b76730-8a80-4f24-a8f8-29d50515ecd5.png)

ahus1:
@kami619 thanks for investigating this. 

I see that the `<user>/sessions` endpoint for users doesn't support pagination. The `<client>/user-sessions` supports pagination (an is therefore faster). Looking at `<realm>/client-session-stats` it is both expensive in its implementation, and I don't see how which values are used for the UI. 

I wonder if this should be forwarded to the UI team to first understand what they use this information for (especially `<realm>/client-session-stats`) and if `<user>/sessions` should get pagination. Maybe they even would want to create their own REST endpoint in the UI project eventually. Based on the results we can optimize those endpoints or not. / cc: @hmlnarik

In the time being, I'd suggest to not call those slow endpoints; that is: comment them out in the load test and link them to an open GitHub issue. 

From a load testing point of view, I would like to understand if the login / logout flow has problems with that amount of sessions, or if it is unaffected by this - in a scenario where a user session has 30-100 client sessions (you could start with a single session per user), and some 100.000-1.000.000 user sessions in the database. This would give us an idea if we're for example missing an index, or loading too much data.

BTW: for the map storage, offline sessions and online sessions are basically the same.

kami619:
@ahus1 thanks for the insight into the load test scenario, I will look into trying out that scenario and get back with data.

I think contact UI team to make the calls consistent is a good idea, I can get in touch with them and see.

ahus1:
> I think contact UI team to make the calls consistent is a good idea, I can get in touch with them and see.

[...]

PR #255, opened 2022-11-09
Add a new admin scenario to list client and user sessions under that realm

kami619:
Relates to #218

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -554,6 +554,21 @@ class KeycloakScenarioBuilder {
     this
   }
 
+  def getClientUUID(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("List Clients and get a Client UUID")
+        .get(ADMIN_ENDPOINT + "/clients?clientId=${clientId}")
+        .header("Authorization", "Bearer ${token}")
+        .queryParam("first",0)
+        .queryParam("max", 1)
+        .check(
+          status.is(200))
+        .check(jsonPath("$[0].id").notNull.saveAs("clientUUID"))
+      )
+      .exitHereIfFailed
+    this
+  }
+
   def viewPagesOfUsers(pageSize: Int, numberOfPages: Int): KeycloakScenarioBuilder = {
     chainBuilder = chainBuilder
       .repeat(numberOfPages, "page") {
@@ -603,6 +618,21 @@ class KeycloakScenarioBuilder {
     this
   }
 
+  def getUserUUID(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("List Users and get a User UUID")
+        .get(ADMIN_ENDPOINT + "/users?username=${username}&exact=true")
+        .header("Authorization", "Bearer ${token}")
+        .queryParam("first", 0)
+        .queryParam("max", 1)
+        .check(
+          status.is(200))
+        .check(jsonPath("$[0].id").notNull.saveAs("userUUID"))
+      )
+      .exitHereIfFailed
+    this
+  }
+
   def deleteUser(): KeycloakScenarioBuilder = {
     chainBuilder = chainBuilder
       .exec(http("Delete User")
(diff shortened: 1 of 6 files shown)


benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -554,6 +554,22 @@ class KeycloakScenarioBuilder {
...
+    chainBuilder = chainBuilder
+      .exec(http("List Clients and get a Client UUID")
+        .get(ADMIN_ENDPOINT + "/clients")
+        .header("Authorization", "Bearer ${token}")
+        .queryParam("first",0)
+        .queryParam("max", 50)
+        .check(
+          status.is(200))
+        //ToDo Randomize the index of the client object to not fetch skewed Session stats
+        .check(jsonPath("$[10].id").notNull.saveAs("clientUUID"))
```
ahus1, line 567: Instead of keeping the TODO, I'd like to use the existing randomization for clients. I lookup the URL parameters in the Keycloak code; they will need some testing. If it doesn't work in a straighforward way, I'd be happy to take another look.

```suggestion
        .get(ADMIN_ENDPOINT + "/clients?clientId=${clientId}")
        .header("Authorization", "Bearer ${token}")
        .queryParam("first",0)
        .queryParam("max", 1)
        .check(
          status.is(200))
        .check(jsonPath("$[0].id").notNull.saveAs("clientUUID"))
```

benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -603,6 +619,22 @@ class KeycloakScenarioBuilder {
...
+    chainBuilder = chainBuilder
+      .exec(http("List Users and get a User UUID")
+        .get(ADMIN_ENDPOINT + "/users")
+        .header("Authorization", "Bearer ${token}")
+        .queryParam("first", 0)
+        .queryParam("max", 50)
+        .check(
+          status.is(200))
+        //ToDo Randomize the index of the user object to not fetch skewed Session stats
+        .check(jsonPath("$[20].id").notNull.saveAs("userUUID"))
```
ahus1, line 632: As with the example above, this is a suggestion to clear the TODO. If the URL doesn't work as expected, I'd be happy to pair.

```suggestion
        .get(ADMIN_ENDPOINT + "/users?username=${username}&exact=true")
        .header("Authorization", "Bearer ${token}")
        .queryParam("first", 0)
        .queryParam("max", 1)
        .check(
          status.is(200))
        .check(jsonPath("$[0].id").notNull.saveAs("userUUID"))
```

ahus1 (changes requested):
I found two TODOs in the code. Instead of documenting them, I suggest to resolve them with. Please see my example snippets below (they haven't been tested, I just looked up the APIs). 

Taking about documentation, it would be great to have for this new scenario a separate page like for the other scenarios. While we still need to document the old ones, it would be great to document the new ones as they come in. That page could detail a little bit more about the circumstances when this would come in useful.

kami619:
@ahus1 thanks for the feedback., with your proposed change, if I understand it correctly, we select the UUID based on a lookup by clientId and Username and that would come with a hard dependency on using randomized users and clients with `users-per-realm` and `clients-per-realm` load run config options passed to the kcb.sh script., when we run things at scale and it would default to `client-0` and `user-0` based on the defaults.

Just making sure, if my understanding is proper, so that I can add this detail into the new scenario page suggestion you made and I agree we should have that page.

Once you confirm this, I will work on  the changes and get back with a new commit.

ahus1:
> using randomized users and clients with `users-per-realm` and `clients-per-realm` load run config options passed to the kcb.sh script., when we run things at scale and it would default to `client-0` and `user-0` based on the defaults.

Yes, this is what I had in mind, as this would make it IMHO consistent with the other tests and the amount of work would be reasonable. I'm happy to discuss this if you think otherwise.

In addition to setting `users-per-realm`/`clients-per-realm` there is also `username`/`client-id` to set specific instance, but I assume those would be would out of the box with the given placeholders.

kami619:
Great, I dont think so its too much effort, I will test this and get back if it doesnt work as intended. Appreciate your quick feedback.

kami619:
@ahus1 tested the changes locally, both lookups, yielded expected results, I will add some fail safe, when things go out of bounds and update the docs. I should have a PR by tomorrow

kami619:
@ahus1 : Please give it another review, now it has the changes from lookup suggestions you made and the Scenario overview doc

ahus1 (approved):
Hi @kami619 - thanks for adding this scenarios and its docs. The docs are extensive and will allow future load tests to make the best use out of this scenario. 

Looking at docs, I found that both endpoints list user sessions, and the one below users will list the clients connected to that user session. I updated the docs accordingly, including the method names.

There is also "getClientSessionStats()", still I didn't find it in the docs overview. I also don't know where to find it in the Admin UI, maybe you could help me there.

I'll also see that I avoided the word "random", and used "from a range" instead as it is IMHO more specific. 

I also avoided the word "root", as that realm is usually called "master". 

Thanks for describing the two variants. I updated the section structure so each has now both a prerequisites and a run section. Please let me know what you think of the new structure. 

When running the curl-command with the "&" in the URL, I needed to add single quotes around it. I also found that the `dataset-import.sh` didn't create the right number of clients and users; I will create a new issue for that.

kami619:
> Looking at docs, I found that both endpoints list user sessions, and the one below users will list the clients connected to that user session. I updated the docs accordingly, including the method names.

Great, thanks for updating the methods and docs., it makes sense to have it this way. I was trying to follow the actual endpoint names from UI, it was confusing to begin with.

> There is also "getClientSessionStats()", still I didn't find it in the docs overview. I also don't know where to find it in the Admin UI, maybe you could help me there.
It is the first end point Admin GUI loads, when you login as an Admin and select the `Sessions` sub-nav from the Realm Navigation menu. Here is a recording of the same.

https://user-images.githubusercontent.com/3415105/201355804-6b97d358-e6be-4a11-b071-21f8c307f0dc.mp4



> I'll also see that I avoided the word "random", and used "from a range" instead as it is IMHO more specific.
> I also avoided the word "root", as that realm is usually called "master".
> Thanks for describing the two variants. I updated the section structure so each has now both a prerequisites and a run section. Please let me know what you think of the new structure.

Thanks for these changes, I think it makes sense for them to be the way you specified. Will make a note for future.

PR #256, opened 2022-11-09
Expose the SQLPAD_QUERY_RESULT_MAX_ROWS in sqlpad app

### LINKED ISSUE #254: SQLPad currently has a default value of query results max rows to be returned
### Description

SQLPad currently has a default value of query results max rows to be returned

The value which could be set in the configuration of sqlpad.yaml is not currently exposed and sits at 10000 rows as a default.

### Discussion

_No response_

### Motivation

_No response_

### Details

We should expose this variable to be configurable from the https://github.com/keycloak/keycloak-benchmark/blob/main/provision/minikube/keycloak/templates/sqlpad.yaml#L25 env variables for defining the configurations

kami619:
Expose the SQLPAD_QUERY_RESULT_MAX_ROWS configuration variable to be able to configure max rows displayed as a result of a query

Fixes #254

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/util/sqlpad.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/util/sqlpad.adoc
@@ -13,6 +13,8 @@ include::partial$minikube-ip-change.adoc[]
 
 The username is `admin` with the password `admin`.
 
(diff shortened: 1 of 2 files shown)


ahus1 (approved):
Thank your for finding this option and also documenting it.

PR #257, opened 2022-11-09
Setup multi-node CockroachDB with the CockroachDB operator

### LINKED ISSUE #202: Provision CockroachDB as a multi-node installation
### Description

To gather more insights about CockroachDB, prepare a setup with a multi-node CockroachDB setup. As we'd eventually want to have multiple datacenters representated as different namespaces, it might be the time to switch to the CockroachDB operator. 

This would be a be an additional option for `KC_DATABASE` in addition to the current `cockroach` setting. 

### Motivation

With a multi-node setup the performance of CockroachDB might be lower due to the necessary coordination. Using the operator also gets us one step further to a multi-DC setup with multiple namespaces.

ahus1:
Closes #202


ahus1:
@kami619 - this is now ready for a review and a manual test. 

Notable change: 
* The existing "cockroach" storage configuration has been renamed to "cockroach-single".
* You'll need to downgrade to Keycloak 1.25.x using `./rebuild.sh` as the Operator doesn't support Kubernetes 1.25.x
* A minimal reset of the CockroachDB with the operator is possible with `init-database.sh`, a full reset of CockroachDB with the operator is possible after dropping the Keycloak namespace, waiting for 5 minutes for the CockroachDB pods to complete, and deleting the PVs. 

Please let me know your thoughts on this. I expect at least one review round, it is a rather big change. 

To reset minikube

kami619:
> You'll need to downgrade to Keycloak 1.25.x using ./rebuild.sh as the Operator doesn't support Kubernetes 1.25.x

@ahus1 did you mean we have to downgrade to Kubernetes 1.24.x using ./rebuild.sh, looks like you did intend that, based on the change in the rebuild.sh.

ahus1:
> did you mean we have to downgrade to Kubernetes 1.24.x using ./rebuild.sh, looks like you did intend that, based on the change in the rebuild.sh.

Oh, sorry, yes: downgrade to Kubernetes 1.24.x

Unfortunately, when you do that on our central server, it will loose all previously recorded statistics in Prometheus :-(

provision/minikube/keycloak/templates/cockroach-operator/cockroach.yaml
```diff
@@ -0,0 +1,28 @@
...
+            storage: "20Gi"
+        volumeMode: Filesystem
+  tlsEnabled: true
+  # You can set either a version of the db or a specific image name
+  # cockroachDBVersion: v22.1.8
+  image:
+    name: cockroachdb/cockroach:v22.1.8
+  # nodes refers to the number of crdb pods that are created
+  # via the statefulset
+  nodes: 3
```
kami619, line 26: @ahus1 do we want to mention this value of number of nodes that come up by default in the docs ?

ahus1: +1 on that. Do you think it should already be configurable as of today?

kami619: TBH, I think we should expose scalability configuration outside of this ticket, be it Keycloak pods, CockroachDB pods etc., We can leave the configuration in the yaml definitions for now.

kami619: pushed a commit to add the docs, after the discussion with @ahus1

ahus1: approved, thanks for updating the docs!

Issue #214, opened 2022-09-29
dataset-import script -> unstable condition and input arguments validation


lhanusov:
### Describe the bug

It has been tested with default option, which passed. But in case of specifying arguments, 1 argument is omitted and set for a default value and other arguments have different input order: https://github.com/keycloak/keycloak-benchmark/blob/main/dataset/dataset-import.sh#L8

### Version

999-SNAPSHOT

### Expected behavior

Takes specified value for realms.

### Actual behavior

Omits input argument and still set a default value.

### How to Reproduce?

./dataset-import.sh -r '10 100 1000 20000'

ahus1:
I tried to use the `dataset-import.sh` today, and stumbled across this problem; the parameters seem to be "off by one", so it created one client instead of 5. 

Any progress on this issue / should I create a PR here?

```
./dataset-import.sh -r '1 5 30 1000'
+ main -r '1 5 30 1000'
++ minikube ip
+ KEYCLOAK_URI=https://keycloak.***.nip.io/realms/master/dataset/
+ SHORT_ARGS=r::,c:,u:,e:,o:,d:,s,h
+ LONG_ARGS=create-realms::,create-clients:,create-users:,create-events:,create-offline-sessions:,delete-realms:,help,status
++ getopt -a -n dataset-import -o r::,c:,u:,e:,o:,d:,s,h -l create-realms::,create-clients:,create-users:,create-events:,create-offline-sessions:,delete-realms:,help,status -- -r '1 5 30 1000'
+ ARGS=' -r '\'''\'' -- '\''1 5 30 1000'\'''
+ eval set -- ' -r '\'''\'' -- '\''1 5 30 1000'\'''
++ set -- -r '' -- '1 5 30 1000'
+ case "$1" in
+ create_realms -- 1 5 30 1000
+ [[ -- =~ -- ]]
+ REALM_COUNT=1
+ CLIENT_COUNT=1
+ USER_COUNT=5
+ PASSWORD_HASHING=30
+ echo 'Creating 1 realm/s with 1 client/s and 5 user/s with 30 password hash iterations.'
```

PR #259, opened 2022-11-11
Wrong getopt args evaluation fixed - using shift for optional args.

lhanusov:
Fixing #214


ahus1 (approved):
Thanks for fixing this issue; I tested it locally and it works for me.

I updated the commit comment to contain the issue number so that the commit can be traced to the issue and this pull request.

PR #260, opened 2022-11-11
Document Keycloak behavior under load

### LINKED ISSUE #217: Documentation of DB best practices
### Description

* Add a tuning (?) section somewhere
* Document things like connection pool, initial size, acquisition timeout
* Document how Keycloak behaves under load (500s when connection pool is exhausted)

ahus1:
Closes #217

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
@@ -1,0 +1,47 @@
+= Keycloak under load
+:description: This topic describes how Keycloak behaves under a high load.
+
+{description}
+It also offers configuration best practices to prepare Keycloak for a high load.
+
+== Database connections
+
+Creating new database connections is expensive as it takes time.
+Creating them when a request arrives will delay the response, so it is good to have them created before the request arrives.
+It can also contribute to a https://en.wikipedia.org/wiki/Cache_stampede[stampede effect] where creating a lot of connections in a short time makes things worse as it slows down the system and blocks threads.
+Closing a connection also invalidates all service side statements caching for that connection.
+
+include::partial$configure-db-connection-pool-best-practices.adoc[]
+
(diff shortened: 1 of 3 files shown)


ahus1:
@kami619 - I updated the docs to describe some of the behavior of Keycloak under load. I'd be happy if you could review the text when you find the time. 

After reviewing, I'd like to share the text with a wider audience inside of the store, maybe even the whole Keycloak team. 

Please let me know what you think about it. Thanks!

doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
```diff
@@ -0,0 +1,47 @@
+= Keycloak under load
+:description: This topic describes how Keycloak behaves under a high load.
+
+{description}
+It also offers configuration best practices to prepare Keycloak for a high load.
+
+== Database connections
+
+Creating new database connections is expensive as it takes time.
```
kami619, line 9: nit pick: connection instead of connections

doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
```diff
@@ -0,0 +1,47 @@
+= Keycloak under load
```
kami619, line 1: @ahus1 I think this makes sense. We should def follow up with another task, to add pages such as., Load Characteristics (which defines a certain load condition) and Benchmark Numbers (which captures the real world behavior of the application) which attests to the behavior we explained in this doc.

kami619 (approved):
looks good to me

doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
ahus1: +1 for this. It could use some of the baseline scenarios we ran; including the one we recently ran around logins. I think it would be a good fit in that category.

PR #268, opened 2022-11-22
Adding dataset-import task creating realms on each storage

### LINKED ISSUE #196: Update the provision-minikube.yaml to perform dataset operations on the different storage configurations
### Description

Update the provision-minikube.yaml to perform dataset operations on the different storage configurations from within GH Actions workflow.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
Improves the status-completed endpoint to poll on the task till success is returned

Adds a new endpoint function to clear the last status completed record

Closes #196

### DIFF
--- a/.github/workflows/provision-minikube.yml
+++ b/.github/workflows/provision-minikube.yml
@@ -55,15 +55,73 @@ jobs:
           github token: ${{ secrets.GITHUB_TOKEN }}
           driver: docker
           start args: --memory 4096
-      - name: Provision everything
+      - name: Provision Legacy store on PostgreSQL DB
         working-directory: provision/minikube
         run: |
           PATH=$PATH:$GITHUB_WORKSPACE/bin
           ./rebuild.sh
-          task # default configuration PostgreSQL and legacy store
+          task KC_DATABASE=postgres KC_STORAGE=
+          kubectl get pods -A
+          kubectl get events -A
+      - name: Create Dataset on Postgres Legacy Store
+        working-directory: provision/minikube
+        run: |
+          task KC_DATABASE=postgres KC_STORAGE= dataset-import -- --clear-status-completed
+          task KC_DATABASE=postgres KC_STORAGE= dataset-import -- -r '1 5 5 1000'
+          task KC_DATABASE=postgres KC_STORAGE= dataset-import -- --status-completed
+      - name: Provision JPA Map store on PostgreSQL DB
+        working-directory: provision/minikube
+        run: |
+          PATH=$PATH:$GITHUB_WORKSPACE/bin
+          task KC_DATABASE=postgres KC_STORAGE=jpa
+      - name: Create Dataset on Postgres JPA Map Store
+        working-directory: provision/minikube
+        run: |
+          task KC_DATABASE=postgres KC_STORAGE=jpa dataset-import -- --clear-status-completed
+          task KC_DATABASE=postgres KC_STORAGE=jpa dataset-import -- -r '1 5 5 1000'
+          task KC_DATABASE=postgres KC_STORAGE=jpa dataset-import -- --status-completed
+      - name: Provision JPA Map store on Cockroach DB Single Node
+        working-directory: provision/minikube
+        run: |
+          PATH=$PATH:$GITHUB_WORKSPACE/bin
           task KC_DATABASE=cockroach-single KC_STORAGE=jpa
+          kubectl get pods -A
+          kubectl get events -A
+      - name: Create Dataset on Cockroach JPA Map Store
+        working-directory: provision/minikube
+        run: |
+          task KC_DATABASE=cockroach-single KC_STORAGE=jpa dataset-import -- --clear-status-completed
+          task KC_DATABASE=cockroach-single KC_STORAGE=jpa dataset-import -- -r '1 5 5 1000'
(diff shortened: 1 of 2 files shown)


kami619:
@ahus1 

I have implemented based on what we discussed. Few notable differences were.,

- I reverted back to using `./dataset-import.sh` script instead of `dataset-import` task, as the task for some reason taking more time to re-run few dependent tasks, thus taking longer time. If time permits, we can debug that, if not, it should be fine to still use the bash script directly in the GHA workflow.
- On an avg. if I use the `./dataset-import.sh` script, its taking an additional 45 seconds to create datasets on all storage types
- Also, I have a question on how to expose the KEYCLOAK_URI input, the things I tried didn't seem to accomplish what I was trying to do

LMK WDYT ?

.github/workflows/provision-minikube.yml
```diff
@@ -55,15 +55,88 @@ jobs:
...
+          kubectl get events -A
+      - name: Create Dataset on Postgres Legacy Store
+        working-directory: dataset
+        run: |
+          ./dataset-import.sh --clear-status-completed
+          ./dataset-import.sh -r '1 5 5 1000'
+          ./dataset-import.sh --status-completed
+          EXIT_CODE=$?
+          test $EXIT_CODE -eq 0 && echo "Creation of realm successful" || echo "something bad happened"
+          exit $EXIT_CODE
```
ahus1, line 74: This code is not necessary as this is what GitHub actions will do anyway: once a command returns not zero, the step will fail.

dataset/dataset-import.sh
```diff
@@ -52,7 +52,16 @@ dataset_provider_status () {
 
 dataset_provider_status_completed () {
   echo "Dataset provider status of the last completed task"
-  curl -ks "${KEYCLOAK_URI}status-completed"
+  t=0
+  until curl -ks "${KEYCLOAK_URI}status-completed" | grep '"success":"true"' ; do
+    sleep 1 && ((t=t+1))
+    echo "Polling...${t}s"
+  done
```
ahus1, line 59: Should this eventually time out (like after 2 minutes?). The timeout could be another CLI parameter.

ahus1 (commented):
>  the task for some reason taking more time to re-run few dependent tasks

I have an idea: While the `task` calls with the environment variable CLI argument for setting up the minikube setup, one would need to add it to the calls for `dataset-import` task again, as within the setup it is not part of the `.env` file.

As you've been asking on how to override defaults; there's something I've taken from the `generate-custom-report.sh`. This allows overriding the `TARGET_DIR` and `GATLING_REPORT_VERSION` via the CLI. Maybe this gives you an idea?

```
GATLING_REPORT_VERSION=6.0
GENERATE_CSV_REPORT=false
USE_TEMPLATE=false
TARGET_DIR=/home/$USER/reports/

while getopts "hv:s:d:t:c" arg; do
  case $arg in
    h) usage
        ;;
    v) GATLING_REPORT_VERSION="$OPTARG";
        ;;
    s) SRC_DIR="$OPTARG";
        ;;
    d) TARGET_DIR="$OPTARG";
        ;;
    t) TEMPLATE_PATH="$OPTARG";USE_TEMPLATE=true;
        ;;
    c) GENERATE_CSV_REPORT=true;
        ;;
    \?) echo "ERROR: Invalid option provided" >&2
        usage
        ;;
  esac
done
```

See some other small comments below.

.github/workflows/provision-minikube.yml
kami619: ok gtk, i will remove this line of code, then. thank you @ahus1

dataset/dataset-import.sh
kami619: Ok, good idea on making that configurable, so that it could be used for larger datasets as well with a default value. Let me think about the change.

kami619:
> I have an idea: While the task calls with the environment variable CLI argument for setting up the minikube setup, one would need to add it to the calls for dataset-import task again, as within the setup it is not part of the .env file.

You are probably right, it was trying to reset to default. I will test this change on my branch, thanks for the pointer @ahus1

kami619:
> As you've been asking on how to override defaults; there's something I've taken from the generate-custom-report.sh. This allows overriding the TARGET_DIR and GATLING_REPORT_VERSION via the CLI. Maybe this gives you an idea?

Already looked at that example, but we are using secondary args to drive the sub-commands, thats where its getting complex to differentiate between the main line getopt args vs sub-command args

kami619:
https://github.com/keycloak/keycloak-benchmark/actions/runs/3535009621/jobs/5932545475

@ahus1 the recent seems to have done the trick, however it still needs to run the dependent tasks atleast once as shown in the GHA job run logs above.

kami619:
@ahus1 and @lhanusov please review the changes, let me know what you guys think ?

.github/workflows/provision-minikube.yml
kami619: removed this in the latest commit

dataset/dataset-import.sh
kami619: Made this configurable via an optional argument, however it needs to be passed in a particular order, which I have mentioned in the help prompt

ahus1 (approved):
Thanks you Kamesh, this looks good to me. 

When running `task` for the `dataset-provider`, it shows `is up to date`, which indicates that the task is not executed, which is expected.

PR #270, opened 2022-11-24
Add integration test for dataset module

### LINKED ISSUE #267: Add new set of tests in dataset provider module to check for compilation errors with Keycloak nightly code base
### Description

Add new set of tests in dataset provider module to check for compilation errors with Keycloak nightly code base 

### Discussion

_No response_

### Motivation

Improve developer feedback by finding compile time issues with the most recent Keycloak nightly build

### Details

- Update the https://github.com/keycloak/keycloak-benchmark/blob/main/pom.xml#L13 Keycloak pom with the recent nightly build pulled from a repository
- Use a light weight provisioning module to bring up Keycloak and a Database
- Write new smoke tests to make sure things compile and we can deploy the dataset module to the Keycloak and perform some dataset creation operations.

ahus1:
Also adding a scheduled build every week.

Closes #267

### DIFF
--- a/dataset/src/test/java/org/keycloak/benchmark/it/DeploymentIT.java
+++ b/dataset/src/test/java/org/keycloak/benchmark/it/DeploymentIT.java
@@ -1,0 +1,278 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.keycloak.benchmark.it;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.io.FileUtils;
+import org.awaitility.Awaitility;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.Test;
+import org.keycloak.benchmark.dataset.TaskResponse;
+
+import java.io.IOException;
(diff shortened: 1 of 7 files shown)


ahus1:
Hi @kami619 - here's now an integration tests which only uses Java please let me know what you think of this.

ahus1:
side note: with these tests in place, I wonder if we still need all the tests in `provision-minikube.yml`, or if we could skip some of them as we don't expect them to reveal many additional errors.

kami619:
@ahus1 I will get to this today, sorry been occupied with other tasks

doc/dataset/modules/ROOT/pages/changing-provider.adoc
```diff
@@ -18,7 +18,18 @@ The recommended setup is to first complete the guide xref:installation/installat
 
 Developing the provider against Keycloak Wildfly, or with a locally running instance of Keycloak Quarkus is possible as well, still it is not part of this guide today.
 
-== Preparing changes to the provider
+== Preparing changes using the Minikube setup
```
kami619, line 21: Should this be `Preparing changes using the local setup` of sorts ? @ahus1

dataset/src/test/java/org/keycloak/benchmark/it/DeploymentIT.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    @Test
+    public void deployWithLegacyStore() throws IOException, ExecutionException, InterruptedException, URISyntaxException {
+        runTestWithParameter();
+    }
+
+    @Test
+    public void deployWithMapStore() throws IOException, ExecutionException, InterruptedException, URISyntaxException {
+        runTestWithParameter("--storage", "chm");
```
kami619, line 93: @ahus1 does this take values such as ```runTestWithParameter("--storage", "hotrod")``` ? If I create a new test with this option and run, it currently fails

doc/dataset/modules/ROOT/pages/changing-provider.adoc
ahus1: Oops, a copy-and-paste error! I pushed a change to correct this in the way you suggested.

kami619:
> side note: with these tests in place, I wonder if we still need all the tests in `provision-minikube.yml`, or if we could skip some of them as we don't expect them to reveal many additional errors.

I would still keep those tests from `provision-minikube.yml` as they cover external databases such as PostgreSQL and CockroachDB, just in case we have any integration issues with them.

PR #275, opened 2022-12-06
Updating metrics after change to Micrometer

### LINKED ISSUE #273: Update Keycloak metrics after change to Micrometer
### Describe the bug

After keycloak/keycloak#15234 the metric names for the JVM metrics have changed. 

With the nightly releases available from Dec 6 the dashboards need to be updated to reflect the new metric names.

### Version

nightly

### Expected behavior

JVM metrics about DB connection pool and garbage collection should show in the dashboards

### Actual behavior

Metrics will not show due to their changed names, see https://github.com/keycloak/keycloak-documentation/pull/1728 for the name changes.

### How to Reproduce?

Deploy the nightly version of Keycloak, then look at the dashboards.

### Anything else?

_No response_

ahus1:
Fixes #273

### DIFF
--- a/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
+++ b/provision/minikube/monitoring/dashboards/keycloak-perf-tests.json
@@ -1807,7 +1807,8 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "green"
+                "color": "green",
+                "value": null
               },
               {
                 "color": "red",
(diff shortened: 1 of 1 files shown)


ahus1:
With this change the metrics are visible again in the Keycloak performance dashboard. Unfortunately old metrics stored in Prometheus can't be shown in the dashboard any more (still all previously exported dashboard snapshots would still work). 

image

ahus1:
@kami619 - this is now ready for review (assuming that the build passes as green). Thanks!

kami619:
🥇 you beat me to it :D I was making similar changes due to the cockroachDB tests, only one feedback I have is if we should add another JVM Memory widget to show total JVM Heap statistics like the below one to correlate with the Heap widget we had before ?

image

`sum(jvm_memory_used_bytes{job="keycloak/keycloak-metrics",namespace="keycloak",area="heap"})`

`sum(jvm_memory_committed_bytes{job="keycloak/keycloak-metrics",namespace="keycloak",area="heap"})`

ahus1:
@kami619 - +1 for this diagram. I'd even think that I should replace the new diagram with the one you added in your previous comment. The other one just looks to crowded. One could then drill down as necessary if there's a problem. WDYT?

kami619:
@ahus1 LMK WDYT ? 
image

ahus1:
Thank you @kami619 - this looks good to me! Thanks for updating the widget.

Issue #287, opened 2023-01-27
Allow provisioning authorization settings for a client


pedroigor:
### Description

In order to run benchmarks for authorization services, we need to be able to provision resources, scopes, and policies, for a client.

As an initial proposal, we should be able to:

* Provisionin N numbers of resources
* Provision N number of scopes
* Provision scope-based permissions for each scope and resource
* Provision N number of scopes per resource
* Provision user-based policies and allow creating aggregated policies when you want more than one user with access to a scope

### Motivation

Benchmark authorization services

PR #288, opened 2023-01-27
Allow provisioning authorization settings for a client

pedroigor:
* Please, see #287 for the reasoning
* I'm also including a commit with the necessary changes to align the JAX-RS-related code in the dataset with the latest changes from upstream. Let me know if you want a separate PR.

### DIFF
--- a/dataset/src/main/java/org/keycloak/benchmark/dataset/TaskManager.java
+++ b/dataset/src/main/java/org/keycloak/benchmark/dataset/TaskManager.java
@@ -80,8 +80,11 @@ public class TaskManager {
             singleUseObjectProvider.remove(KEY_RUNNING);
             Task task = Task.fromMap(existingTask);
             task.SetSuccess(successfullyFinished);
-            singleUseObjectProvider.remove(KEY_COMPLETED);
-            singleUseObjectProvider.put(KEY_COMPLETED, TimeUnit.DAYS.toSeconds(1), task.toMap());
+            try {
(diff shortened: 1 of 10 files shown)


dataset/src/main/java/org/keycloak/benchmark/dataset/TaskManager.java
```diff
@@ -81,7 +81,6 @@ public void removeExistingTask(boolean successfullyFinished) {
             Task task = Task.fromMap(existingTask);
             task.SetSuccess(successfullyFinished);
             singleUseObjectProvider.remove(KEY_COMPLETED);
-            singleUseObjectProvider.put(KEY_COMPLETED, TimeUnit.DAYS.toSeconds(1), task.toMap());
```
pedroigor, line 84: @ahus1 Do we really need this?

This is causing errors in logs when invoking any endpoint. The errors happen if you try to invoke the same endpoint twice. The second attempt will log the error, the third attempt will run fine, the fourth will log the error, etc.

ahus1: The `KEY_COMPLETED` is there to retrieve the status of the previously finished task. There is no other way to retrieve the status of the previous task and to figure out via the API if it was successful or not. 
What kind of error message do you see?

ahus1 (approved):
LGTM, thanks for adding this!

PR #292, opened 2023-02-02
Make sure user policy creation take into the account the number of users per realm

### LINKED ISSUE #291: Make sure user policy creation take into the account the number of users per realm
### Describe the bug

When provisioning authorization settings to a client using a URL like `http://localhost:8080/realms/master/dataset/authz/create-resources?client-id=client-0&realm-name=realm-0&count=10000&users-per-user-policy=1&scopes-per-resource=3`.

The provisioner is hard coded to support up to `10000` users as per https://github.com/keycloak/keycloak-benchmark/blob/1a5d98f29c2b6ecb6d6ba1d175e04918efb47e8f/dataset/src/main/java/org/keycloak/benchmark/dataset/AuthorizationProvisioner.java#L243.

This causes errors when realms are provisioned with less than `10000` users.

### Version

upstream

### Expected behavior

Should respect the `usersPerRealm` configuration option.

### Actual behavior

This causes errors when realms are provisioned with less than `10000` users.

### How to Reproduce?

_No response_

### Anything else?

_No response_

pedroigor:
Closes #291


kami619:
thanks for fixing the bug, lgtm

PR #294, opened 2023-02-08
Enhancements and fixes to run custom Keycloak images

ahus1:
Fix scripts and docs when working with custom Keycloak images.


ahus1:
Added more disk space for CRDB tests, and fixed also the remote debugging string which needs to be in one line.

PR #297, opened 2023-02-28
Use a valid SemVer format for the SNAPSHOT version

jonkoops:
Changes the default version string from `999-SNAPSHOT` to `999.0.0-SNAPSHOT`. This makes it a valid version under the [Semantic Versioning](https://semver.org/) standard.

The motivation for this change is to make versioning interoperable between Maven and NPM, as the latter only allows for Semantic Versioning. This allows for simplification in the release process, as we no longer need to convert version numbers from one to the other.

For more information see https://github.com/keycloak/keycloak/issues/17335.


ahus1:
@jonkoops - looks good to me. Thank you for creating a PR for this project for the upcoming change in upstream.

As the tests are failing until the nightly build adopted this change, this PR is on hold until the first nightly build with the new versioning scheme is available.

stianst (approved):

stianst:
@ahus1 I'll do a nightly release after we've merged changes to the other repos, and merge this one afterwards

PR #302, opened 2023-03-26
Fixing OpenTelemetry that is broken after upstream changes

ahus1:
This fixes the OpenTelemetry support which broken after upstream changes that remove `curl` from the container. 
It also adds the necessary docs for users, and allows it to be configured by the `.env` file. 

This makes it now a first-class feature in this suite.


ahus1:
@kami619 / @tkyjovsk - this fixes the OpenTelemetry bits that where broken. As the build is green, I merge it. 

I'm not sure when you'll have the time to test this, or if you're using OpenTelemetry at all. When you encounter any problems with the code or the docs, please create an issue and I'll rework it.

kami619:
thanks for the PR @ahus1. I will try to find some time this week to verify this as its already in the main line.

kami619:
I was able to confirm things are working as expected with OTEL and the new env variables that we added. Cheers !

PR #306, opened 2023-04-04
Add "mixed" (JPA/HotRod) storage setup

sguilhen:
This introduces a mixed setup that mirrors what we have in the Legacy store: auth sessions, user sessions, login failures, and single use objects in hotrod, rest on JPA.

For documentation purposes, I'm adding some info about changes we had to make to be able to run this scenario.

1- In the postgres deployment, we've added an arg to enable prepared transactions, which are disabled by default. This is required by Arjuna when it is attempting to coordinate the transactions involving the two stores. So this change:

```
          args:
            - -c
            - max_prepared_transactions=100
```
prevents the Arjuna exception:

```
2023-03-22 09:54:22,744 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-6) Uncaught server error: java.lang.RuntimeException: javax.transaction.HeuristicMixedException
	at org.keycloak.transaction.JtaTransactionWrapper.handleException(JtaTransactionWrapper.java:78)
	at org.keycloak.transaction.JtaTransactionWrapper.commit(JtaTransactionWrapper.java:92)
. . . 
	Caused by: org.postgresql.util.PSQLException: ERROR: prepared transactions are disabled
  Hint: Set max_prepared_transactions to a nonzero value.
		at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2676)
		at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2366)
		at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:356)
```

2- In `keycloak.yaml`, there's a new env variable that specifies a writable directory that can be used by Arjuna to write files while it is coordinating the transactions:

```
              # arjuna needs access to a writable directory to write temporary files
              - name: QUARKUS_TRANSACTION_MANAGER_OBJECT_STORE_DIRECTORY
                value: /tmp
```

prevents the error

```
"message": "ARJUNA012225: FileSystemStore::setupStore - cannot access root of object store: ObjectStore/ShadowNoFileLockStore/defaultStore/",
```

### DIFF
--- a/.github/workflows/provision-minikube.yml
+++ b/.github/workflows/provision-minikube.yml
@@ -108,6 +108,19 @@ jobs:
           task KC_DATABASE=infinispan KC_STORAGE=hotrod dataset-import -- -a clear-status-completed
           task KC_DATABASE=infinispan KC_STORAGE=hotrod dataset-import -- -a create-realms -r 1 -c 5 -u 5 -i 1000
           task KC_DATABASE=infinispan KC_STORAGE=hotrod dataset-import -- -a status-completed
+      - name: Provision JPA Map store on Hotrod (In-Memory store)
+        working-directory: provision/minikube
+        run: |
+          PATH=$PATH:$GITHUB_WORKSPACE/bin
+          task KC_DATABASE=postgres+infinispan KC_STORAGE=jpa+hotrod
+          kubectl get pods -A
+          kubectl get events -A
+      - name: Create Dataset on Infinispan Hotrod Store
+        working-directory: provision/minikube
+        run: |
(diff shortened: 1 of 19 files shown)


ahus1 (approved):
Thank you for providing the extended configuration. I added another commit to update the docs. 

As there might be other mixed setup in the future, I decided to rename them as follows:

```
KC_DATABASE=postgres+infinispan
KC_STORAGE=jpa+hotrod
```

PR #307, opened 2023-04-04
updating docs to support kvm2 and java as pre-req installations

### LINKED ISSUE #300: Keycloak benchmark documentation should mention the kvm2 setup
### Description

When running the `rebuild.sh` script as described in step 3 of the [documentation](https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/installation), I've run into a pop-up window requesting authentication and saying that `System policy prevents management of local virtualized systems`. After entering the password, the command was left blinking indefinitely, and no progress was ever made.

It turns out that in my local machine I had never set up `kvm2` properly (https://minikube.sigs.k8s.io/docs/drivers/kvm2/), and my user had to be added to the `libvirt` group. A simple command `sudo usermod -aG libvirt $USER` was enough to fix the group membership and allow the `rebuild.sh` script to finish.

Thus I'm requesting that the documentation is updated with a small section on to verify the `kvm2` setup before running the script.

Also, when running the `task` command in step 4, it first failed with `task: Failed to run task "tlsdisableagent": exit status 1`, which turns out to happen because of a missing `JAVA_HOME` env variable. It is probably a good idea to mention the JAVA_HOME requirement in the docs as well.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
- Updated https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/prerequisite/prerequisite-minikube with a Linux section explaining that KVM2 is necessary, explained how to set it up for a Linux host OS and a small check to verify that.
- Java is added as a prerequisite https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/installation, with steps for installation of OpenJDK and create a JAVA_HOME variable for specific JDK version and step to verify if the installation was successful.

Additionally added couple of checks to ensure the host is vetted for
- Virtualization and the user is associated with the libvirt group
- Validate if JAVA_HOME variable is set properly

-----
Fixes #300

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/customizing-deployment.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/customizing-deployment.adoc
@@ -148,7 +148,7 @@ Default value: `1`
 KC_CPU_LIMITS::
 Sets CPU limits per Keycloak pod.
 +
-Default value: `1`
+Default value: `4`
 +
 The value must adhere to the https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu[Kubernetes CPU units] format.
 
--- a/doc/kubernetes/modules/ROOT/pages/installation.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/installation.adoc
@@ -13,6 +13,7 @@ Each linked page contains an installation description and a command to verify th
 +
 --
 * xref:prerequisite/prerequisite-minikube.adoc[minikube]
+* 100 GB of free disk space (64 GB is the virtual disk of minikube)
 * xref:prerequisite/prerequisite-helm.adoc[Helm]
 * xref:prerequisite/prerequisite-kubectl.adoc[kubectl]
 * xref:prerequisite/prerequisite-task.adoc[task]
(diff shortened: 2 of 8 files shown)


kami619:
@ahus1 looks like the precondition check I created is failing on the GHA

```
task: [ipchange] rm -f .task/checksum/keycloak
task: [dataset] ./mvnw -B -am -pl dataset clean install -DskipTests
task: precondition not met
Error: Process completed with exit code 1.
```
I am guessing this is because the host node on GHA might not have a $JAVA_HOME set ?

kami619:
> @ahus1 looks like the precondition check I created is failing on the GHA
> 
> ```
> task: [ipchange] rm -f .task/checksum/keycloak
> task: [dataset] ./mvnw -B -am -pl dataset clean install -DskipTests
> task: precondition not met
> Error: Process completed with exit code 1.
> ```
> 
> I am guessing this is because the host node on GHA might not have a $JAVA_HOME set ?

looks like this is all set after I addressed the conditional check which was improper in the first place.

ahus1:
@kami619 - I ran several tests, and found that `mvnw` also works with Java being on the path. I also did some adjustments to cover for the Windows operating system. 

Please have a look and give it also a run on MacOS on the `rebuild.sh` and for `task mvnw` which should check just the precondition. 

If that works, this is ready to be squashed and merged. Thanks!

ahus1:
> I would defer this to a different issue/PR in future and make the check specific to the Linux OS alone

@kami619 - yes, I support that approach. Please push any changes necessary for that last bit, and then go ahead an squash and merge this PR.

PR #316, opened 2023-04-20
Porting changes for supporting Dataset and Gatling User entities creation on OCP

### LINKED ISSUE #314: Add support to openshift provisioning module to create base entities to support gatling benchmark and dataset scripts
### Description

Add support to openshift provisioning module to create base entities to support gatling benchmark and dataset scripts

- Similar to minikube module we need to have support to create the needed service account enabled client, test-realm and users to start a benchmark against the Openshift cluster based Keycloak instance.
- Akin to the above point, we also need to be able to deploy dataset module to the Openshift cluster based Keycloak instance and create needed datasets for various entities.

### Discussion

_No response_

### Motivation

_No response_

### Details

The question I have for this particular feature request is along the lines of implementation for gatlinguser task from minikube/Taskfile.yaml

Do we want to create these `tasks` under the `common/Taskfile.yml` ? That way we can simply modify the `gatlinguser` to pick up the `KC_HOSTNAME_SUFFIX` from the `.env` file to be under a conditional if block to only run the varied version of the below bash command when the `KC_HOSTNAME_SUFFIX` variable exists in the context.

Suggested Change:

```
- >
  bash -c '
  if [ "{{.KC_HOSTNAME_SUFFIX}}" != "" ]; 
    then ../keycloak-cli/keycloak/bin/kcadm.sh config credentials --server https://keycloak.{{.KC_HOSTNAME_SUFFIX}}/ --realm master --user admin --password admin;
    else ../keycloak-cli/keycloak/bin/kcadm.sh config credentials --server https://keycloak.{{.IP}}.nip.io/ --realm master --user admin --password admin;
  fi'
```

For the dataset module, I believe there would be no changes needed, but would need some testing to make sure we don't hit any runtime issues.

kami619:
Fixes #314 

- Added the support to create the needed service account enabled client, test-realm and users to start a benchmark against the Openshift cluster based Keycloak instance.
- Akin to the above point, to be able to deploy dataset module to the Openshift cluster based Keycloak instance and create needed datasets for various entities.

Even though the dataset module is ported, I am observing an issue @ahus1 with the keycloak/providers directory unable to be mounted to `/opt/keycloak/providers` thus not able to run the dataset-import.sh script to create the entities, I might need some help on this.

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/openshift.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/openshift.adoc
@@ -14,7 +14,7 @@ WARNING: OpenShift must not be accessible from the Internet, as this setup opens
 * xref:prerequisite/prerequisite-kubectl.adoc[kubectl]
 * xref:prerequisite/prerequisite-task.adoc[task]
 * xref:prerequisite/prerequisite-yq.adoc[yq]
-// * xref:prerequisite/prerequisite-java.adoc[Java]
+* xref:prerequisite/prerequisite-java.adoc[Java]
 
 Verify that the setup works using `kubectl get pods -A`.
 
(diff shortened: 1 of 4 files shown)


ahus1 (approved):
Hi @kami619 - thank you for this pull request! I tested it with deleting the keycloak namespace and running `reset-keycloak`. Keycloak started up, and after adding the Keycloak URL for the dataset provider also that worked for me. 

For me this change is ready to be merged. Thanks!

PR #323, opened 2023-04-25
Add possibility to run more Keycloak deployments concurrently in the …

### LINKED ISSUE #318: Make keycloak namespace configurable
### Description

Currently, kcb provisions keycloak instance to a hardcoded namespace called `keycloak`. As part of this issue, we should make this configurable (maybe a prefix would be enough) so more people can work concurrently on the same cluster without disrupting each other's work.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mhajas:
…same Openshift cluster

Closes #318


ahus1:
@mhajas - `GRAPHITE_TCP_ADDR` is not set automatically, but manually by the user when running the Gatling tests. There is a minimal set of docs in `other.adoc`

ahus1 (approved):
Thank you, @mhajas - this looks good to me. Besides a rebase due to a conflict, I did the following: 

* Reformatted some of the AsciiDoc contents to follow the one-sentence-per-line pattern. You can apply this using the reformatting functionality in IntelliJ. Having one sentence per line produces nicer diffs if you change the text later.
* The documentation build failed as there was a warning that an AsciiDoc attribute wasn't resolvable. Those were actually not attributes, but placeholders in your text, so I wrapped them with `++` (passthrough).

PR #326, opened 2023-04-27
Remove unnecessary exports

mhajas:
While working on the namespace customization I needed to pass the `KC_NAMESPACE_PREFIX` property to `isup.sh` script. I tried to use export in the `keycloak` task, however, it didn't work, and then I forgot to remove it.

This Pr is to remove it.


kami619 (approved):
LGTM and works on my local.

PR #334, opened 2023-05-05
Add Grafana dashboard for Embedded Infinispan

### LINKED ISSUE #330: Expose Infinispan metrics to Prometheus by default
### Description

When monitoring a load test, or running a Keycloak cluster as a SRE, I want to see the with how many entries the caches are filled, and how the timing performs. In a failover scenario we want to know how the sessions rebalance after a failover.

Hint: this should be possible by adding 
```xml
<metrics gauges="true" histograms="true" /> 
```
after the "transport" tag in the `kcb-infinispan-cache-config.xml` (permanently) and changing 
KC_CUSTOM_INFINISPAN_CONFIG to true in the local configuration. 

Also adding
```
... statistics="true" 
```
to the cache container as outlined here: https://www.keycloak.org/server/caching 

Looking at this from the docs perspective, it might be beneficial to have next to the "infinispan.adoc" documentation which covers the map store Infinispan setup another document that covers the integrated Infinispan with the legacy store.  

Acceptance criteria: 

Number of instances per cache available as a metric (Gauge)
Reviewed histogram metrics to see if they are useful or not, and decide whether to enable or disable them.
One minimal separate dashboard with first Infinispan metrics   
Follow-up task: Explore with metrics make sense, extend the Infinispan dashboard or integrate them in the existing dashboard.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mhajas:
Closes #330

There are three types of rows on the dashboard:

Global statistics - showing info about all caches globally:
Global stats

Local cache stats - local caches are used for caching objects from the database or LDAP, therefore contain hit ratio:
Local cache stats

Distributed cache stats - distributed caches are used for storing data like sessions or login failures:
Distributed cache stats


kami619:
@mhajas we might have fix the dashboard for the openshift deployment, do you want to try and take this up as a separate ticket ?

image

kami619:
and on minikube I can see the metrics coming through, except for this widget, do you observe the same @mhajas ?

image

ahus1 (changes requested):
I'm hesitant to change the default value to "true", as this would then by default never use the cache configuration that is part of Keycloak. As the cache configuration is part of the system-under-test, I'd rather not do this. 

Could we instead think about Pedro's suggestion to have a environment variable that we can set to configure the default configuration?

mhajas:
@kami619 I just realized that what I mentioned previously about not the `work` cache is not correct, it is used even without remote caches. Why it was not visible is that I forgot to enable statistics for that cache :see_no_evil: 

@ahus1 I looked into how we configure metrics in Keycloak and we need some changes in dist to create some env variable from `metrics-enabled` option so we can consume it in the cache config file. I am wondering whether we need to do it in this version or if we can create a follow-up issue in Keycloak and then adjust kcb for that. WDYT?

ahus1 (approved):
Thank you for adding this dashboard. As agreed earlier, the additional configuration for the metrics will be a separate issue in the Keycloak main repository. Once that has been merged, it is a follow-up task to not have this as a default.

PR #339, opened 2023-05-23
Autodetect KC_HOSTNAME_SUFFIX via `kubectl cluster-info`.

tkyjovsk:


ahus1:
Thank you for this idea @tkyjovsk - I updated it a little bit to fetch the URL from the Console route using `kubectl`.

PR #340, opened 2023-05-24
Fix loading CLUSTER_NAME from .env instead of using username

stianst:


ahus1:
Looking at the latest changes in the PR referenced above that is now merged, this seems to be obsolete. Closing.

PR #343, opened 2023-05-24
Do not use default security in openshift

### LINKED ISSUE #336: Do not use default security in Openshift
### Description

Currently, Keycloak uses the default username and password for Keycloak when deployed on Openshift. Also, Grafana dashboards are available with no authentication, we should introduce some level of security as this can be deployed to some publicly available OCP.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mhajas:
Closes #336

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/customizing-deployment.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/customizing-deployment.adoc
@@ -238,6 +238,14 @@ Available options: 0-26
 +
 NOTE: This option is implemented only for Openshift deployments.
 
+KC_ADMIN_PASSWORD::
+The password for the Keycloak admin user and Grafana admin user.
++
+The default value is obtained from a secret from AWS Secret Manager named `keycloak-master-password` from region `eu-central-1`.
+If `aws` command fails or is not available on the machine, string `admin` is used.
++
+NOTE: This option is implemented only for Openshift deployments.
+
 == Available Benchmark options
 
 The following configuration options are available to configure the helper applications.
--- a/provision/minikube/keycloak/templates/keycloak-initial-admin-secret.yaml
+++ b/provision/minikube/keycloak/templates/keycloak-initial-admin-secret.yaml
@@ -7,5 +7,5 @@ metadata:
   namespace: {{ .Values.namespace }}
 type: kubernetes.io/basic-auth
 data:
-  password: YWRtaW4= # admin
+  password: {{ .Values.keycloakAdminPassword | b64enc }} # admin by default
   username: YWRtaW4= # admin
--- a/provision/minikube/keycloak/values.yaml
+++ b/provision/minikube/keycloak/values.yaml
@@ -30,3 +30,4 @@ metaspaceInitMB: 96
 metaspaceMaxMB: 256
 customInfinispanConfig: false
 portOffset: 0
+keycloakAdminPassword: admin
--- a/provision/openshift/Taskfile.yaml
+++ b/provision/openshift/Taskfile.yaml
@@ -6,6 +6,7 @@ vars:
   KC_HOSTNAME_SUFFIX: '{{default "$(kubectl get route/console -n openshift-console -o jsonpath=\u0027{.spec.host}\u0027 | cut -d . -f 2-)" .KC_HOSTNAME_SUFFIX}}'
   KC_NAMESPACE_PREFIX: '{{default "$(whoami)-" .KC_NAMESPACE_PREFIX}}'
   KC_PORT_OFFSET: '{{default "0" .KC_PORT_OFFSET}}'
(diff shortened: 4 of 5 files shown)


mhajas:
@ahus1 If I understand correctly, you also suggested changing the password for the user created here: https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/content/bin/initialize-benchmark-entities.sh#L82

I see the user does not have any permissions, it is used only in performance tests itself, is it worth changing the password also for this user? I don't see the possibility to run `kcb.sh` from Taskfile so we would need to manually add the password parameter to kcb runs.

kami619:
> @ahus1 If I understand correctly, you also suggested changing the password for the user created here: https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/content/bin/initialize-benchmark-entities.sh#L82
> 
> I see the user does not have any permissions, it is used only in performance tests itself, is it worth changing the password also for this user? I don't see the possibility to run `kcb.sh` from Taskfile so we would need to manually add the password parameter to kcb runs.

I wouldn't think so it would be helpful to change the current way, as the usage of this script is beyond Openshift module, minikube and also in CI., so until we have a unified CI flow, lets keep this unchanged.

ahus1:
@mhajas - ok, let's then do without changing the Gatling password, and possibly add that later.

provision/aws/aws_add_secret_to_openshift_env_file.sh
```diff
@@ -0,0 +1,12 @@
...
+
+source ./.env
+
+KEYCLOAK_MASTER_PASSWORD_SECRET_NAME=${KEYCLOAK_MASTER_PASSWORD_SECRET_NAME:-"keycloak-master-password"}
+# Force eu-central-1 region for secrets manager so we all work with the same secret
+REGION="eu-central-1"
+KEYCLOAK_MASTER_PASSWORD=$(aws secretsmanager get-secret-value --region $REGION --secret-id "$KEYCLOAK_MASTER_PASSWORD_SECRET_NAME" --query SecretString --output text --no-cli-pager)
+
+echo "KC_ADMIN_PASSWORD='$KEYCLOAK_MASTER_PASSWORD'" >> ../openshift/.env
+echo "Password added to .env file."
```
kami619, line 12: It would be nice to mention this in the docs with something like 

```

Optionally we can use `aws_add_secret_to_openshift_env_file.sh` script to set up the openshift `.env` variable `KC_ADMIN_PASSWORD` with the newly generated admin password, which will then be re-used for Keycloak Admin and Grafana Admin secure deployments with dynamic rotatable admin password.
```

provision/aws/aws_add_secret_to_openshift_env_file.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+source ./.env
+
+KEYCLOAK_MASTER_PASSWORD_SECRET_NAME=${KEYCLOAK_MASTER_PASSWORD_SECRET_NAME:-"keycloak-master-password"}
+# Force eu-central-1 region for secrets manager so we all work with the same secret
+REGION="eu-central-1"
+KEYCLOAK_MASTER_PASSWORD=$(aws secretsmanager get-secret-value --region $REGION --secret-id "$KEYCLOAK_MASTER_PASSWORD_SECRET_NAME" --query SecretString --output text --no-cli-pager)
+
+echo "KC_ADMIN_PASSWORD='$KEYCLOAK_MASTER_PASSWORD'" >> ../openshift/.env
```
kami619, line 11: I see two problems with this script's execution,

1. This would create duplicates when run multiple times.
2. And as a User might not know necessarily which is the older password once I make an update.

My suggestion would be to make it more disambiguous., such as overwriting the password in the .env for the KC_ADMIN_PASSWORD rather than appending it.

@tkyjovsk and @ahus1 any thoughts ?

ahus1: Agreed, it shouldn't be appended on every run. 

@mhajas - could we make it a default like we did it for `KC_NAMESPACE_PREFIX` with `default "$(...)"`  so it wouldn't need to be in the `.env` file at all?

mhajas: @kami619 @ahus1 Thank you for the feedback. I didn't like the idea of having openshift Taskfile dependent on `aws` command, however, using `|| echo admin` should work if `aws` command is not present on the machine, so I guess this is much more user-friendly. Please take a look at the latest changes.

kami619: thanks for addressing that @mhajas . I will try it out and merge it 👍🏼

ahus1 (approved):
I reviewed the code and tried it out locally after rebasing it with main. 
LGTM and is ready to be merged.

PR #345, opened 2023-05-26
Add GitHub workflows to create and delete a shared ROSA cluster

stianst:
I've tested this on my fork:

* Create cluster: https://github.com/stianst/keycloak-benchmark/actions/runs/5088576833/jobs/9145155404
* Delete cluster: https://github.com/stianst/keycloak-benchmark/actions/runs/5088530163/jobs/9145048362

### DIFF
--- a/.github/workflows/rosa-cluster-create.yml
+++ b/.github/workflows/rosa-cluster-create.yml
@@ -1,0 +1,49 @@
+name: ROSA Cluster - Create
+
+on:
+  workflow_dispatch:
+    inputs:
+      computeMachineType:
+        description: 'Instance type for the compute nodes'
+        required: true
+        default: m5.xlarge
+        type: text
+      multiAz:
+        description: 'Deploy to multiple data centers'
+        required: true
+        default: false
+        type: boolean
+      replicas:
+        description: 'Number of worker nodes to provision'
+        required: true
+        default: 2
+        type: text
+
+env:
+  OPENSHIFT_VERSION: 4.12.17
+
+jobs:
+
+  checkout:
+    name: Create ROSA cluster
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup ROSA CLI
+        uses: ./.github/actions/rosa-cli-setup
+        with:
+          aws-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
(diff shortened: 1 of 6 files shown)


.github/workflows/rosa-cluster-create.yml
```diff
@@ -0,0 +1,49 @@
...
+          aws-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
+          rosa-token: ${{ secrets.ROSA_TOKEN }}
+      - name: Create ROSA Cluster
+        run: ./rosa_create_cluster.sh
+        working-directory: provision/aws
+        env:
+          VERSION: ${{ env.OPENSHIFT_VERSION }}
+          CLUSTER_NAME: shared
```
stianst, line 45: Was thinking to use 'gh-${{ github.repository_owner }}' as the cluster name. That way it can create the cluster 'gh-keycloak' when ran in the main repository, while 'gh-username' if you want to run these actions in your own repo (after adding secrets of course).

mhajas: Makes sense to me. Then we can add the cluster name as the parameter for the benchmark run.

kami619:
also we are generating some debug logs for each custer create and delete using rosa scripts, do we want to retrieve them somehow for debugging the errors on aws or otherwise ? if you think, that should be done in a future PR, thats OK with me too. @stianst

stianst:
> @stianst
> 
> In the referenced run : https://github.com/stianst/keycloak-benchmark/actions/runs/5088576833/jobs/9145155404
> 
> ```
> Login failed (401 Unauthorized)
> Verify you have provided the correct credentials.
> 2023-05-26T08:39:44+00:00
> WARNING: Using insecure TLS client config. Setting this option is not supported!
> ```
> 
> I see that it failed to authenticate a bunch of times before it succeeded., which is unlike the behavior we saw on local runs., is this expected ?

I changed it to not wait for 1m, but instead just keep trying every 10 seconds, so that'll be the reason for a few more failures there. I think it is better to have it fail a few times, then waiting for too long?

stianst:
> 

I'd say let's add it later. Probably just grab the logs and store as artifacts if it fails to deploy. Don't think we need any logs if it is successful? Related to this is we could probably also make sure everything is cleaned-up if something fails.

kami619 (approved):
Agreed on log management to be taken care later, fine for it fail fast, than have to wait. This PR looks good to be merged, if there are no other comments

.github/workflows/rosa-cluster-create.yml
mhajas: yes I was commenting on you comment around `github.owner`

PR #347, opened 2023-05-26
Run benchmark workflow

stianst:


provision/openshift/isup.sh
```diff
@@ -5,7 +5,9 @@
 KC_NAMESPACE_PREFIX=$(cat .task/var-KC_NAMESPACE_PREFIX)
 KC_HOSTNAME_SUFFIX=$(cat .task/var-KC_HOSTNAME_SUFFIX)
 
-source .env
+if [ -f .env ]; then
```
kami619, line 8: @stianst should this be ? ```if [ -f ./.env ]; then```

provision/openshift/isup.sh
```diff
@@ -5,7 +5,9 @@
...
+  source .env
+fi
```
stianst, line 10:
```suggestion
if [ -f ./.env ]; then
  source ./.env
fi
```

stianst: @kami619 doesn't really matter as the two are identical, but can change it as above to make it consistent with the other scripts

kami619: yup, called it out for consistency sake 💯

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
+name: ROSA Cluster - Run Benchmark
+
+on:
+  workflow_dispatch:
+
+concurrency: shared_cluster
```
kami619, line 6: mostly a question for my clarity, this is to force only 1 job for all `shared_cluster` jobs ? where would this value `shared_cluster` be consumed ?

dataset/dataset-import.sh
```diff
@@ -14,7 +14,9 @@ set_environment_variables () {
   EVENTS_COUNT="100"
   SESSIONS_COUNT="100"
   HASH_ITERATIONS="20000"
-  KEYCLOAK_URI="https://keycloak-keycloak.$(minikube ip || echo 'unknown').nip.io/realms/master/dataset"
+  if ( minikube version &>/dev/null ); then
+    KEYCLOAK_URI="https://keycloak-keycloak.$(minikube ip || echo 'unknown').nip.io/realms/master/dataset"
```
kami619, line 18: thanks for taking care of this 👍🏼

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Get URLs
+        run: |
+          KEYCLOAK_URL=https://$(oc get routes -n keycloak | grep keycloak-service | awk '{ print $2 }')
+          echo "KEYCLOAK_URL=$KEYCLOAK_URL" >> "$GITHUB_ENV"
+
+      - name: Create dataset
+        run: |
+          ./dataset-import.sh -a create-realms -r 1 -c 1 -u 100 -l ${{ env.KEYCLOAK_URL }}/realms/master/dataset
+          STATUS_CMD="./dataset-import.sh -a status -l https://keycloak-keycloak.apps.shared.hesr.p1.openshiftapps.com/realms/master/dataset"
```
kami619, line 69: there seems to be a static keycloak url here to run the dataset-import.sh script, shouldn't this be dynamic, when we will have shared but scheduled clusters ? @stianst

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: ./.github/actions/task-setup
+
+      - name: Login to OpenShift cluster
+        run: ./rosa_oc_login.sh
+        working-directory: provision/aws
+
+      - name: Build with Maven
+        run: |
+          ./mvnw -B clean install -DskipTests
+          tar xfvz benchmark/target/keycloak-benchmark-0.10-SNAPSHOT.tar.gz
```
kami619, line 51: I see there is no regex for benchmark version, would this be handled as a form input later on ? Otherwise it would break when we upgrade version within the project.

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Run Authorization Code scenario
+        run: bin/kcb.sh --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=${{ env.KEYCLOAK_URL }} --realm-name=realm-0
+        working-directory: keycloak-benchmark-0.10-SNAPSHOT
+
+      - name: Archive Gatling reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: gatling-results
+          path: keycloak-benchmark-0.10-SNAPSHOT/results
```
kami619, line 83: I see there is no regex for benchmark version, would this be handled as a form input later on ? Otherwise it would break when we upgrade version within the project.

.github/workflows/rosa-run-benchmark.yml
stianst, line 69:
```suggestion
          STATUS_CMD="./dataset-import.sh -a status -l ${{ env.KEYCLOAK_URL }}/realms/master/dataset"
```

stianst: Copy/paste error ;)

.github/workflows/rosa-run-benchmark.yml
stianst, line 51:
```suggestion
          tar xfvz benchmark/target/keycloak-benchmark-*.tar.gz
```

stianst: Was being a little bit lazy there perhaps ;)

See below I can just replace it with a '*' I reckon

.github/workflows/rosa-run-benchmark.yml
stianst: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

Basically it prevents more than one workflow run with the same concurrency to be executed in parallel by GH Actions. That will even span multiple workflows, say if we introduce a separate "rosa-run-failure-testing.yml". We can and probably should also add these to the rosa-cluster-create/delete, as that would actually allow you to run cluster-create first, then run this workflow before even waiting for the cluster to be started.

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+          ./dataset-import.sh -a create-realms -r 1 -c 1 -u 100 -l ${{ env.KEYCLOAK_URL }}/realms/master/dataset
+          STATUS_CMD="./dataset-import.sh -a status -l ${{ env.KEYCLOAK_URL }}/realms/master/dataset"
+          echo -n 'Waiting for dataset to be created: '
+          for i in {1..30}; do $STATUS_CMD | grep 'No task in progress' &>/dev/null && break || echo -n '.' && sleep 10; done
+          echo ' done'
+        working-directory: dataset
+
+      - name: Run Authorization Code scenario
+        run: bin/kcb.sh --scenario=keycloak.scenario.authentication.AuthorizationCode --server-url=${{ env.KEYCLOAK_URL }} --realm-name=realm-0
+        working-directory: keycloak-benchmark-0.10-SNAPSHOT
```
kami619, line 77:
```suggestion
        working-directory: keycloak-benchmark-*-SNAPSHOT
```

.github/workflows/rosa-run-benchmark.yml
kami619:
```suggestion
          path: keycloak-benchmark-*-SNAPSHOT/results
```

mhajas (commented):
Thanks @stianst, the Pr looks good to me after the last suggestion from @kami619 is addressed with benchmark version in `upload-artifact` action.

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Archive Gatling reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: gatling-results
+          path: keycloak-benchmark-0.10-SNAPSHOT/results
+          retention-days: 5
+
+      - name: Delete Gatling project
+        run: oc delete project monitoring
```
mhajas, line 87: Should we delete `monitoring` namespace? After the tested scenario is done, all Grafana dashboard data are removed due to this step.

Maybe we can leave `monitoring` namespace as is and remove only Keycloak.  WDYT?

.github/workflows/rosa-run-benchmark.yml
stianst: Ended up renaming the dir to `keycloak-benchmark` to make it simpler later when used.

.github/workflows/rosa-run-benchmark.yml
stianst: Removed the install/remove of Grafana. As you mentioned it doesn't make all that much sense to create it then delete it at the end. Thinking we should probably install Grafana as part of the ROSA cluster?

stianst:
Did a run of the last iteration here:
https://github.com/stianst/keycloak-benchmark/actions/runs/5118671819

Will of course squash and merge this one since it now has a few commits ;)

.github/workflows/rosa-run-benchmark.yml
mhajas: Makes sense to me. I believe it should work the way you suggested but we need to test it. We can do it as part of some follow-up. For now, we can merge without monitoring.

mhajas (approved):
LGTM, thanks @stianst!

PR #352, opened 2023-05-31
Automated scheduling of creation of shared ROSA cluster and another end of business reaper 

### LINKED ISSUE #351: Automate provisioning of ROSA clusters
Automatically create a shared ROSA cluster at the start of the business day, and automatically destroy all clusters at the end of the business day.

To prevent shutdown of clusters in use it should be possible to mark a particular ROSA cluster to be kept alive. For example by creating a project `keepalive` in the cluster as a marker.

kami619:
Adding two new workflow files:

- rosa-cluster-auto-provision-on-schedule
- rosa-cluster-auto-delete-on-schedule

Setting static values for cluster creation, as below:

- CLUSTER_NAME: shared-crossdc
- COMPUTE_MACHINE_TYPE: m5.xlarge
- MULTI_AZ: false
- REPLICAS: 2

and a Rosa cluster reaper scripts, which takes input of a list of "ready" state clusters

- rosa_cluster_reaper.sh

My first attempt at GH workflow's so if something's amiss, please correct it :) 

Open question, there could be cluster's which could be in other `states` other than `ready`, do we want any alert on it ? Another question is, do we want any email notification after the job is complete, alternative would be to personally tune the GH Workflow notifications to report on this workflow, reducing the email clutter.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -27,6 +27,13 @@
    1. Name of the cluster - the name of the cluster to destroy. Default value is `gh-${{ github.repository_owner }}`, this results in `gh-<owner of fork>`.
 4. Wait for the workflow to finish.
 
+## Periodic Jobs
+1. We are managing the creation of shared ROSA clusters' creation and deletion using below workflows.
+   2. `.github/workflows/rosa-cluster-auto-delete-on-schedule.yml`
+   3. `.github/workflows/rosa-cluster-auto-provision-on-schedule.yml`
+4. Optionally you can run the `task keepalive` from provision/openshift directory against your Openshift cluster to keep it alive, from the `rosa-cluster-auto-delete-on-schedule` workflow's delete activity on the defined schedule time.
+
+
 ## Run benchmark
 
 1. Go to repository Actions -> On the left side choose `ROSA Cluster - Run Benchmark`
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -29,9 +29,9 @@
 
 ## Periodic Jobs
 1. We are managing the creation of shared ROSA clusters' creation and deletion using below workflows.
-   2. `.github/workflows/rosa-cluster-auto-delete-on-schedule.yml`
-   3. `.github/workflows/rosa-cluster-auto-provision-on-schedule.yml`
-4. Optionally you can run the `task keepalive` from provision/openshift directory against your Openshift cluster to keep it alive, from the `rosa-cluster-auto-delete-on-schedule` workflow's delete activity on the defined schedule time.
+   1. `.github/workflows/rosa-cluster-auto-delete-on-schedule.yml`
+   2. `.github/workflows/rosa-cluster-auto-provision-on-schedule.yml`
+2. Optionally you can run the `task keepalive` from provision/openshift directory against your Openshift cluster to keep it alive, from the `rosa-cluster-auto-delete-on-schedule` workflow's delete activity on the defined schedule time.
 
 
 ## Run benchmark
--- a/.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
@@ -1,0 +1,16 @@
+name: ROSA Scheduled Delete
+
+on:
+  schedule:
+    - cron: '0 19 * * *' # Runs every day at 7 PM UTC.
+
+jobs:
+
+  checkout:
+    name: ROSA Scheduled Delete cluster(s)
+    runs-on: ubuntu-latest
+    steps:
+      - run: gh workflow run -R keycloak/keycloak-benchmark rosa-cluster-delete.yml -f deleteAll='yes'
+        if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -1,0 +1,16 @@
+name: ROSA Scheduled Create
+
+on:
+  schedule:
+    - cron: '0 7 * * 1-5' # Runs At 07:00 UTC on every day-of-week from Monday through Friday.
+
+jobs:
+
+  checkout:
+    name: ROSA Scheduled Create cluster
+    runs-on: ubuntu-latest
+    steps:
+      - run: gh workflow run -R keycloak/keycloak-benchmark rosa-cluster-create.yml
+        if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/provision/aws/rosa_cluster_reaper.sh
+++ b/provision/aws/rosa_cluster_reaper.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+set -e
+
+if [ -f ./.env ]; then
+  source ./.env
+fi
+
+echo "Starting the ROSA cluster reaper at $(date -uIs)"
+
+# Get a list of running ROSA clusters
+CLUSTERS=$(rosa list clusters --output json | jq -r '.[] | select((.state == "ready") or (.state == "pending") or (.state == "waiting") or (.state == "error")) | .name')
+
+if [ -z "$CLUSTERS" ]; then echo "Didn't find any ready state clusters"; exit 1; fi
+
+# Loop through each cluster
+for CLUSTER_NAME in $CLUSTERS; do
+  echo "Checking Cluster: $CLUSTER_NAME"
+  export CLUSTER_NAME=$CLUSTER_NAME
+  #Login to the Openshift Cluster
+  ./rosa_oc_login.sh
+  #Check if the 'keepalive' namespace exists in the specific cluster
+  if oc get projects | grep -q "keepalive"; then
+    echo "keepalive namespace exists in the cluster $CLUSTER_NAME., skipping deleting it."
+  else
+    echo "keepalive namespace doesn't exist in the cluster $CLUSTER_NAME., deleting it."
+    #Delete the Individual Cluster
+    ./rosa_delete_cluster.sh
+  fi
+done
+
+echo "Finished reaping all possible clusters at $(date -uIs)"
+
+
+
--- a/.github/workflows/rosa-cluster-delete.yml
+++ b/.github/workflows/rosa-cluster-delete.yml
@@ -6,6 +6,11 @@ on:
       clusterName:
         description: 'Name of the cluster'
         type: text
+      deleteAll:
+        description: 'Delete all clusters'
+        required: true
+        default: 'no'
+        type: text
 jobs:
 
   delete:
@@ -23,8 +28,14 @@ jobs:
           aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
           rosa-token: ${{ secrets.ROSA_TOKEN }}
 
-      - name: Delete ROSA Cluster
+      - name: Delete a ROSA Cluster
+        if: ${{ inputs.deleteAll == 'no' }}
         run: ./rosa_delete_cluster.sh
         working-directory: provision/aws
         env:
           CLUSTER_NAME: ${{ inputs.clusterName || format('gh-{0}', github.repository_owner) }}
+
+      - name: Delete all ROSA Clusters
+        if: ${{ inputs.deleteAll == 'yes' }}
+        run: ./rosa_cluster_reaper.sh
+        working-directory: provision/aws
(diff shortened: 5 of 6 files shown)


.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -0,0 +1,36 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
+          rosa-token: ${{ secrets.ROSA_TOKEN }}
+
+      - name: Create ROSA Cluster
+        run: ./rosa_create_cluster.sh
+        working-directory: provision/aws
+        env:
+          VERSION: ${{ env.OPENSHIFT_VERSION }}
+          CLUSTER_NAME: shared-crossdc
```
mhajas, line 32: Do we need this parameter? Without it, this would default to `gh-keycloak` which means we don't need to set `clusterName` param for each benchmark run.

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
```diff
@@ -0,0 +1,26 @@
...
+    - cron: '0 19 * * *' # Runs every day at 7 PM UTC
+
+jobs:
+
+  delete:
+    name: Delete ROSA cluster
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
stianst, line 14: Add the following to prevent scheduled run in forks:

```suggestion
        uses: actions/checkout@v3
        if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
```

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -0,0 +1,36 @@
...
+          REGION: ${{ vars.AWS_DEFAULT_REGION }}
+          COMPUTE_MACHINE_TYPE: m5.xlarge
+          MULTI_AZ: false
+          REPLICAS: 2
```
stianst, line 36: Instead of repeating steps from `rosa-cluster-create.yml` it would be better to just run that workflow like this:

```suggestion
      - run: gh workflow run -R keycloak/keycloak-benchmark rosa-cluster-create.yml
        if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
```

This prevents us from having to maintain two workflows with the same steps. Alternatively the `rosa-cluster-create.yml` can be converted to a composite workflow (like we do in keycloak-rel), but that's a bit messier IMO.

provision/aws/rosa_cluster_reaper.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ -f ./.env ]; then
+  source ./.env
+fi
+
+echo "Starting the ROSA cluster reaper at $(date -uIs)"
+
+# Get a list of running ROSA clusters
+CLUSTERS=$(rosa list clusters --output json | jq -r '.[] | select(.state == "ready") | .name')
+
+if [ -z "$CLUSTERS" ]; then echo "Didn't find any ready state clusters"; exit 1; fi
```
stianst, line 13: Should we not also destroy any clusters not in ready state? I.e. those that are installing or have failed for some reason?

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
```diff
@@ -0,0 +1,26 @@
+name: ROSA Scheduled Delete
+
+on:
+  schedule:
+    - cron: '0 19 * * *' # Runs every day at 7 PM UTC
+
+jobs:
+
+  delete:
+    name: Delete ROSA cluster
```
stianst, line 10: Wonder if we would want to re-use `rosa-cluster-delete.yml` here instead? We could add an input to that workflow `Delete all`, which would delete all, then it this scheduled workflow we could run that workflow like suggested below.

provision/aws/rosa_cluster_reaper.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Loop through each cluster
+for CLUSTER_NAME in $CLUSTERS; do
+  echo "Checking Cluster: $CLUSTER_NAME"
+  ./rosa_oc_login.sh
+  #Check if the 'keepalive' namespace exists in the specific cluster
+  keepalive=$(oc get projects | grep "keepalive")
```
mhajas, line 20: Do you think it would make sense to introduce some script that would create the `keepalive` namespace, so we avoid some problems with spelling or something similar?

Also, from other parts of kcb I had a feeling we are trying to use `kubectl` where possible. Should we change this too?

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
kami619: ok, I was not sure, if the `workflow_dispatch` type workflows can be reused ? @stianst that would make sense if we can run it like a downstream workflow to re-use the `rosa-cluster-create.yml`

provision/aws/rosa_cluster_reaper.sh
kami619: for keepalive, I can add a small task command or bash script to do it that way., or have an option during the rosa_cluster_create.sh bash script itself to have a flag to create that empty namespace. Alternatively, we can add the below snippet to docs., choice would be ours. Let me know which you think would work best.

```
oc new-project hello-openshift \
--description="This is an example project" \
--display-name="Hello OpenShift"
```

On your second question, I was leading off of the `rosa_oc_login.sh` script, where indeed use `oc` to login to the cluster, wouldn't it be confusing to use `kubectl` after that ?

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
kami619: good catch, I will revert this to fallback to the default `CLUSTER_NAME`

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
kami619: thank you, I will add this change

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
kami619: @stianst didn't fully understand your suggestion here, we could have an additional flag within `rosa-cluster-delete.yml` if you would like to `Delete all`., but I was not sure how to add multiple trigger events for both `workflow_dispatch` to do a manual run as well as `schedule` based cron run., hence had to split out the workflow. If we could do the two triggers in the same workflow., that would be cool and save us from repeating similar steps.

provision/aws/rosa_cluster_reaper.sh
mhajas: I think we can do both. task command sounds good to me for local usage. Not sure what you mean by docs, but I guess we should add some info about this to https://github.com/keycloak/keycloak-benchmark/blob/main/.github/workflows/README.md

mhajas: Regarding `kubectl`, I don't have any strong opinion on that. I am just pointing out that in other places we usually use `kubectl`.

stianst:
@kami619 running out a bit of time before leaving for PTO, and not sure I explained correctly what I proposed, so will try to explain here:

* rosa-cluster-create.yml when executed by default should create the default shared cluster we want
* rosa-cluster-delete.yml should have an option to delete all clusters directly

Now to schedule the above you can either do:

1. Add schedule directly to those workflows
2. Add schedule in separate workflows

Personally I like separating the schedule from the workflow, so leaning towards 2. As you can see in https://github.com/keycloak/keycloak-quickstarts/blob/latest/.github/workflows/schedule.yml it's fairly simple to create a schedule workflow that is only used to execute another workflow. `gh workflow run` also allows passing inputs to the workflow it's running, so if we don't want to make delete all clusters the default in `rosa-cluster-delete.yml` it can enable this option when running that workflow.

provision/aws/rosa_cluster_reaper.sh
kami619: > I think we can do both. task command sounds good to me for local usage. Not sure what you mean by docs, but I guess we should add some info about this to https://github.com/keycloak/keycloak-benchmark/blob/main/.github/workflows/README.md

Perfect, will implement that and add it to the README as well. Thanks @mhajas

kami619: applied the changes now. @mhajas when you have a chance pls review again

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
kami619: applied the changes

.github/workflows/README.md
```diff
@@ -18,6 +18,7 @@
    3. Deploy to multiple availability zones in the region - if checked, the cluster will be deployed to multiple availability zones in the region. Default value is `false`.
    4. Number of worker nodes to provision - number of compute nodes in the cluster. Default value is `2`.
 4. Wait for the workflow to finish.
+5. Optionally you can run the `task keepalive` from provision/openshift directory against your Openshift cluster to keep it alive.
```
mhajas, line 21: Should we also mention somewhere there is a periodic job destroying all clusters? Currently, it is not visible what is the purpose of `keepalive` namespace. Maybe we could create a new section called `Periodic jobs`?

provision/aws/rosa_cluster_reaper.sh
```diff
@@ -0,0 +1,46 @@
...
+  export CLUSTER_NAME=$CLUSTER_NAME
+  ./rosa_oc_login.sh
+
+  #Check if the 'keepalive' namespace exists in the specific cluster
+  if oc get projects | grep -q "keepalive"; then
+    echo "keepalive namespace exists in the cluster $CLUSTER_NAME., skipping deleting it."
+  else
+    echo "keepalive namespace doesn't exist in the cluster $CLUSTER_NAME., deleting it."
+
+    CLUSTER_ID=$(rosa describe cluster --cluster "$CLUSTER_NAME" | grep -oPm1 "^ID:\s*\K\w+")
```
mhajas, line 27: Can we replace lines `27-39` with `./rosa_delete_cluster.sh`?

mhajas (commented):
Thanks @kami619 for the changes. I added two more comments.

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
```diff
@@ -0,0 +1,16 @@
+name: ROSA Scheduled Delete
+
+on:
+  schedule:
+    - cron: '0 19 * * *' # Runs every day at 7 PM UTC
```
mhajas, line 5: Just realised, should we schedule this task only on work days? If I understand correctly this will be triggered also on weekends.

provision/aws/rosa_cluster_reaper.sh
kami619: However, alternatively we can combine these two scripts into one and run it based on delete 1 vs delete all clusters logic, I can do that, if you want to keep these two together ? @mhajas

mhajas: > that would be problematic given in `rosa_delete_cluster.sh` the `CLUSTER_NAME` is being set this way,
> 
> ```
> CLUSTER_NAME=${CLUSTER_NAME:-$(whoami)}
> if [ -z "$CLUSTER_NAME" ]; then echo "Variable CLUSTER_NAME needs to be set."; exit 1; fi
> ```

I would say it should just work as it is already working for `./rosa_oc_login.sh` on line 19.

> However, alternatively we can combine these two scripts into one and run it based on delete 1 vs delete all clusters logic, I can do that, if you want to keep these two together ? @mhajas

Yeah, that would work too, but we would need to add additional parameters to `rosa_delete_cluster.sh`, but we could get rid of one shell script so maybe it is worth it.

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
kami619: good catch, I shall adjust that for both cluster create and delete schedules

probably will use this expression to keep it concise
`- cron: '0 19 * * 1-5'`

mhajas: I accidentally added this comment to the delete workflow. It should probably go mainly to the create workflow.

.github/workflows/README.md
kami619: will add a new section for `Periodic Jobs`, thanks for the suggestion

.github/workflows/rosa-cluster-auto-delete-on-schedule.yml
kami619: so leave delete for all days but create for only weekdays, is that the final expectation ?

mhajas: Makes sense to me @kami619 even though during the weekends there will be probably no cluster to destroy most of the time, still it causes no harm to run the script.

kami619:
@mhajas last round 🥊  :) review 🤞🏼

mhajas (approved):
Thank you for the changes @kami619, LGTM!

I updated the `README.md` file a little bit because the items were wrongly numbered.

PR #361, opened 2023-06-01
New inputs for benchmark workflow

pruivo:
* CPU and Memory
* Custom Keycloak image
* Fixed Openshift project name

### DIFF
--- a/.github/workflows/rosa-run-benchmark.yml
+++ b/.github/workflows/rosa-run-benchmark.yml
@@ -5,13 +5,40 @@ on:
     inputs:
       clusterName:
         description: 'Name of the cluster'
-        type: text
+        type: string
       replicas:
         description: 'Number of keycloak pods'
         required: true
         default: '3'
+      disableStickySessions:
+        description: 'Disable sticky session in Openshift Route'
+        type: boolean
+        default: false
+        required: true
+      image:
+        description: 'Keycloak Image'
+        type: string
+      podMemoryRequests:
+        description: 'Keycloak pod memory requests (in MB)'
+        type: string
+      podMemoryLimit:
+        description: 'Keycloak pod memory limit (in MB)'
+        type: string
+      podCpuRequests:
+        description: 'Keycloak pod CPU requests'
+        type: string
+      podCpuLimit:
+        description: 'Keycloak pod CPU limit'
+        type: string
+      heapMaxSizeMB:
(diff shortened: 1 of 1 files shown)


pruivo:
Got a successful run in my private repo: https://github.com/pruivo/keycloak-benchmark/actions/runs/5143287976

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -5,14 +5,38 @@ on:
     inputs:
       clusterName:
         description: 'Name of the cluster'
-        type: text
+        type: string
+      namespacePrefix:
```
mhajas, line 9: Do we have some use-case for setting `namespacePrefix` here? Thanks to 
```
concurrency: cluster_${{ inputs.clusterName || format('gh-{0}', github.repository_owner) }}
```
we are sure there is only one benchmark workflowrunning at each time. Is there some advantage of running each in a different namespace?

kami619: Ideally we should avoid running multiple benchmarks against different namespaces in parallel.

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -5,14 +5,38 @@ on:
...
+      disableStickySessions:
+        description: 'Disable sticky session in Openshift Route'
+        type: boolean
+        default: false
+        required: true
+      image:
+        description: 'Keycloak Image'
+        type: string
+      podMemoryLimit:
+        description: 'Keycloak pod memory limit (in MB)'
```
kami619, line 25: we would need to set some default values, if we are going to introduce limits, also consider adding requests, as requests and limits go hand in hand.

.github/workflows/rosa-run-benchmark.yml
pruivo: ok, I'll hardcode the prefix.

.github/workflows/rosa-run-benchmark.yml
```diff
@@ -65,12 +92,19 @@ jobs:
         run: task
         working-directory: provision/openshift
         env:
-          KC_NAMESPACE_PREFIX: ''
+          KC_NAMESPACE_PREFIX: ${{ env.NAMESPACE_PREFIX }}
```
kami619, line 95:
```suggestion
          KC_NAMESPACE_PREFIX: ${{ env.PROJECT_PREFIX }}
```

PR #374, opened 2023-06-13
Avoid node-ports usage in OpenShift

ahus1:
This simplifies the installation of multiple Keycloak installations in a shared OpenShift cluster.

As those ports are AFAIK not accessible from outside of the cluster in the current setup, they are not very useful anyway.


ahus1:
Hi @mhajas - as you've been involved in node ports, I'd like you to have a look if you have the time. Thanks!

mhajas (approved):
Thank you @ahus1, LGTM.

PR #375, opened 2023-06-13
modified the exit code to not fail on completion of a no-op

### LINKED ISSUE #371: ROSA scheduled cluster deletion fails on weekends
### Describe the bug

The scheduled job to delete any running clusters is failing on weekends as there are no clusters running.

See https://github.com/keycloak/keycloak-benchmark/actions/runs/5237041733 as an example.

### Version

main

### Expected behavior

Either schedule only runs on weekdays like the create cluster, or it doesn't fail if there is no running clusters to delete.

I'm leaning towards the latter since a cluster may be created manually on weekends.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

kami619:
fixes #371


ahus1 (approved):
LGTM, thanks!

PR #376, opened 2023-06-13
automated reporting of aws costs per Year-To-Date, Month-To-Date, Week-To-Date

kami619:
- currently set to run on Every Sunday at 12 PM UTC
- currently only reports on the high level costs and doesn't report on service level cost split,. we can add it based on the need
- reuses the aws/rosa cli setup action which already exists
- currently YTD, MTD, WTD costs are reported as separate steps, we can merge them if thats the need

### DIFF
--- a/.github/workflows/rosa-cost-report.yml
+++ b/.github/workflows/rosa-cost-report.yml
@@ -1,0 +1,34 @@
+name: ROSA/AWS - Cost Reporting
+
+on:
+  schedule:
+    - cron: '0 12 * * Sun' # Runs At 12:00 PM UTC on Every Sunday
+  workflow_dispatch:
+
+jobs:
+  report:
+    name: Report on ROSA and AWS Costs for the current YTD, MTD and WTD
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup ROSA CLI
+        uses: ./.github/actions/rosa-cli-setup
+        with:
+          aws-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
+          rosa-token: ${{ secrets.ROSA_TOKEN }}
+
+      - name: Report on the Week-to-Date costs
+        run: aws ce get-cost-and-usage --time-period Start=$(date -d "7 days ago" +"%Y-%m-%d"),End=$(date +%Y-%m-%d) --granularity=MONTHLY --metrics BlendedCost --output yaml
+
+      - name: Report on the Week-to-Date costs with daily cost split
+        run: aws ce get-cost-and-usage --time-period Start=$(date -d "7 days ago" +"%Y-%m-%d"),End=$(date +%Y-%m-%d) --granularity=DAILY --metrics BlendedCost --output yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/rosa-cost-report.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Setup ROSA CLI
+        uses: ./.github/actions/rosa-cli-setup
+        with:
+          aws-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
+          rosa-token: ${{ secrets.ROSA_TOKEN }}
+
+      - name: Report on the Week-to-Date costs
+        run: bash -c "aws ce get-cost-and-usage --time-period Start=$(date -d "7 days ago" +"%Y-%m-%d"),End=$(date +%Y-%m-%d) --granularity=DAILY --metrics BlendedCost --output yaml"
```
ahus1, line 24: Would this work even without the `bash -c` prefix, or changed to have `shell: bash`?

.github/workflows/rosa-cost-report.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup ROSA CLI
+        uses: ./.github/actions/rosa-cli-setup
+        with:
+          aws-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-default-region: ${{ vars.AWS_DEFAULT_REGION }}
```
ahus1, line 21: I assume that if we don't pass on the ROSA_TOKEN, the call will fail. Assuming that the AWS CLI is already installed on the action, could you configure it straight away?

ahus1 (changes requested):
Thank you for the change. Is there a completed action someone in your repo?
And: do we really want to publish the costs on the public internet?

.github/workflows/rosa-cost-report.yml
kami619: should be possible run without bash -c prefix., if thats our preference, just making sure to use bash shell in this case

kami619:
> Thank you for the change. Is there a completed action someone in your repo?
> And: do we really want to publish the costs on the public internet?

I will run against my local fork and share the results. I'm not sure if aws cost is sensitive information., I can check with the team once again.

.github/workflows/rosa-cost-report.yml
kami619: and it works without the `bash -c` just fine, will remove that in the final version.

.github/workflows/rosa-cost-report.yml
kami619: yup, its a hard requirement for the rosa-cli-setup.sh script so will add that back.

kami619:
> > Thank you for the change. Is there a completed action someone in your repo?
> > And: do we really want to publish the costs on the public internet?
> 
> I will run against my local fork and share the results. I'm not sure if aws cost is sensitive information., I can check with the team once again.

https://github.com/kami619/keycloak-benchmark/actions/runs/5267303873/jobs/9522385504

@ahus1 here's the output of a local run

kami619:
we are going on a different route to fetch this data

Issue #373, opened 2023-06-12
Ability to view scalability scenario results over time


stianst:

stianst:
Options:

* https://horreum.hyperfoil.io/
* JSON file with GH pages, or something along those lines

ahus1:
Idea: fixed load (100 users per second), then visualize response times and CPU usage.

ahus1:
I did some analysis what can work. This is my suggestion: 

* Create one big JSON file for the summary of the results (`global_stats.json` from Gatling would be a start) and the description of the inputs (like number of pods, CPU limits, etc.). This can be scripted with shell scripts and `jq`. Each run should get a UUID to identify it later. 
* Create an archive of the exported metrics from Prometheus. When we know the start and end/time of the run, this should be possible with `promtool tsdb dump`. Additional filtering might be necessary to reduce the dataset

The data can then be analyzed: 

* The one big JSON can be stored in PostgreSQL in a JSONB column, and visualized in charts in Grafana. SQL then helps us to select the set of scenario results we want to compare.
* The JSON could also be imported to Horreum if we choose to do so later.
* For multidimensional plots, tools like pandas or gnuplot will be better suited. They would also work better when we plan to publish static sites with diagrams. For now, let's stick with tools we know to build interactive tools. 
* The archived metrics can be re-imported to Prometheus, a tool like https://github.com/fgouteroux/prom-push-cli might help here. It seems necessary to deploy a new Prometheus instance on OpenShift as the built-in instance doesn't allow that, probably for good reasons.
* When viewing restored metrics data, it might be helpful to shift the start time a fixed date (for example 0:00 of the current date), so multiple runs can be compared side by side in Grafana even if they started  at different times, and also add a label that identifies the scenario run.

In this issue, I'll focus on creating the big JSON file of the summary, adding the necessary PostgreSQL instances in minikube and OpenShift, and provide the scripting necessary.

kami619:
this approach looks viable, thanks for putting this together @ahus1

stianst:
Where would we deploy PostgreSQL to? Is the JSON file the source of truth and it's just imported into PostgreSQL to be able to query? Wonder if we even need PostgreSQL or would be able to just handle the JSON in-memory?

ahus1:
The JSON file will be the source of truth, and it will be created during the GitHub run using `jq`.
[...]

PR #379, opened 2023-06-15
Schedule daily run of scalability benchmark

### LINKED ISSUE #366: Schedule scalability scenario for currently defined scenarios
We would like to run a client credential grant scalability scenario with some good input parameters to the kcb.sh script on a daily basis on weekdays and capture the mean throughput and mean response time of the endpoints part of the client credential grant flow and write it out to a csv file on a branch where it won't be subject to TTL for archived files.

As part of this task, we also will start a discussion within the team to find out the best input parameters for attributes such as,

- warm-up
- ramp-up
- measurement
- user-think-time
- sla-error-percentage
- sla-mean-response-time
- clients-per-realm
- realms

mhajas:
Closes #366

### DIFF
--- a/.github/workflows/keycloak-scalability-benchmark-on-schedule.yml
+++ b/.github/workflows/keycloak-scalability-benchmark-on-schedule.yml
@@ -1,0 +1,27 @@
+name: ROSA Scheduled Create
+
+on:
+  schedule:
+    - cron: '0 7 * * 1-5' # Runs At 07:00 UTC on every day-of-week from Monday through Friday.
+
+jobs:
+
+  checkout:
+    name: ROSA Scheduled Create cluster
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run ClientSecret scalability benchmark
+        run: |
+          gh workflow run -R keycloak/keycloak-benchmark keycloak-scalability-benchmark.yml \
+            -f "scenarioName=authentication.ClientSecret"
+        if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run AuthorizationCode scalability scenario
+        run: |
+          gh workflow run -R keycloak/keycloak-benchmark keycloak-scalability-benchmark.yml \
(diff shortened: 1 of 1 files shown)


mhajas:
@ahus1 @kami619 I am not sure what exactly should be part of this issue. Should I add also some csv file with results to kcb.sh? Or this will be done as part of #373

kami619:
is it possible to plot the data from the archived jsons in a small summary csv based on the gatling runs ? we cam then use the the other ticket to store this data elsewhere

ahus1:
I'd argue that creating a CSV or JSON different from Grafana is outside of the scope of this issue, and should be part of #373.

mhajas:
I agree with @ahus1 here, as part of this issue, we don't know what format we need, so I would rather postpone it to #373 We can then split the issue into more if needed.

kami619:
> I agree with @ahus1 here, as part of this issue, we don't know what format we need, so I would rather postpone it to #373 We can then split the issue into more if needed.

ok, my attempt for a requirement creep-in didn't work ;) I am fine with it to be part of #373

.github/workflows/keycloak-scalability-benchmark-on-schedule.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+
+  checkout:
+    name: ROSA Scheduled Create cluster
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run ClientSecret scalability benchmark
+        run: |
+          gh workflow run -R keycloak/keycloak-benchmark keycloak-scalability-benchmark.yml \
+            -f "scenarioName=authentication.ClientSecret"
```
kami619, line 16: do we want to define this for ClientSecret scenario as well ? `-f "numberOfEntitiesInRealm=100000"`

kami619: true, but I only see the `numberOfEntitiesInRealm` defined for AuthorizationCode scenario and not for ClientSecret., or am I missing something ? @mhajas

mhajas: `numberOfEntitiesInRealm` defaults to 10000, so it does not need to be added for ClientSecret. Do we want to add it too? For example, if the default changes?

Issue #372, opened 2023-06-12
Add SLO metrics for Client Credential Grant and Authorization Code to Keycloak


ahus1:
Building blocks of the solution:
* Identify the different endpoints in Keycloak that should collect metrics for the two scenarios.
* Collect duration metrics via Micrometer as histograms, so they can be aggregated across a cluster. 
* Collect request counts via Micrometer as histograms, so they can be aggregated across a cluster. 
* Provide a PromQL query / minimal Grafana dashboard which collects the SLO for each scenario. 

**SLO candidates:**

Client credential grant
* 99% of requests serviced in < 50 ms
* 99.99% of requests are successfully completed

Authorization code flow
* 99% of requests serviced in < 200 ms
* 99.99% of requests are successfully completed

**Design discussion:** 

Automatically added metrics for URL endpoints didn't work well with subresources, see https://github.com/keycloak/keycloak/issues/17281. An issue to support this with Quarkus is still open, see https://github.com/quarkusio/quarkus/issues/31497. It might still be worth checking the current state of things as Keycloak has been upgrades to Quarkus 3.1.x. This might or might relate to https://github.com/keycloak/keycloak/issues/9036

While subresource metrics don't work well with Quarkus' out-of-the-box support, they work nicely with the OTEL agent. 

_Summary:_ The simplest way seems to be to add the metrics would be to use the OTEL agent and gather the metrics on a per-URL base with a duration and a HTTP status code. If this doesn't work as expected, we'll consider other options.

**Next steps/out of scope here:** 

* For a test, run a query directly against Prometheus to see if the SLO have been met during the test's duration.
* Evaluate if elastic loadbalancer and haproxy get meaningful metrics

ahus1:
Evaluate if elastic loadbalancer and haproxy get meaningful metrics

mhajas:
The following was delivered as part of this issue:
- OTEL now works with Openshift
- A new dashboard was introduced with SLO data for Client credentials grant
- Jaeger now works with Openshift

SLO dashboard for Authorization code flow was not added, but we decided to close this now and reopen it or create a new issue for the rest of the work when needed.

PR #384, opened 2023-06-20
Make otel work with Openshift and add Grafana dashboard

mhajas:
Relates to #372

WIP - works only with 1 pod at the moment due to PVC access mode

Board with example data (15 minutes run of client credential grant with 100_000 clients and 10_000 clients-per-realm and 10 concurrent users):
screenshot_20230620_154242

### DIFF
--- a/.github/actions/keycloak-create-deployment/action.yml
+++ b/.github/actions/keycloak-create-deployment/action.yml
@@ -24,6 +24,9 @@ inputs:
     description: 'Keycloak pod CPU limit'
   heapMaxSizeMB:
     description: 'Keycloak server maximum Java heap size (in MB)'
+  otel:
+    description: 'Enable OpenTelemetry'
+    default: 'true'
 
 runs:
   using: "composite"
@@ -45,3 +48,4 @@ runs:
         KC_CPU_REQUESTS: ${{ inputs.podCpuRequests }}
         KC_CPU_LIMITS: ${{ inputs.podCpuLimit }}
         KC_HEAP_MAX_MB: ${{ inputs.heapMaxSizeMB }}
+        KC_OTEL: ${{ inputs.otel }}
--- a/provision/common/Taskfile.yaml
+++ b/provision/common/Taskfile.yaml
@@ -88,6 +88,7 @@ tasks:
       - test "{{.KC_METASPACE_INIT_MB}}" == "$(cat .task/var-KC_METASPACE_INIT_MB)"
       - test "{{.KC_METASPACE_MAX_MB}}" == "$(cat .task/var-KC_METASPACE_MAX_MB)"
       - test "{{.KC_CUSTOM_INFINISPAN_CONFIG}}" == "$(cat .task/var-KC_CUSTOM_INFINISPAN_CONFIG)"
+      - test "{{.KC_DISABLE_STICKY_SESSION}}" == "$(cat .task/var-KC_DISABLE_STICKY_SESSION)"
 
   mvnw:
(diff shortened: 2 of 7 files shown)


mhajas:
Please, do not merge. It seems the PR does not contain my latest changes to the dashboard. WIll update it tomorrow.

mhajas:
I updated the PR with the latest changes, it still depends on #387 so I will leave it as a draft.

Here is a dashboard with more pods:

ahus1:
Thank you for providing the OTEL configuration and the dashboard. 

The OTEL part is working, which is nice to see, and ready to be merged. 

It was a good idea to put in the cache hit rates into the diagram - with the examples I found on the shared cluster, there seems to be something strange going on which calls for a deeper analysis next week.

I gave the dashboard a try on the shared instance. Here's what I found:

I see that the metrics are calculated from the counters that reset only on Keycloak's restart. I'd suggest that that the metrics should be moving averages, both for the cache hit and the response times. This will show us how the performance changes over time, and it is also the way an SRE would analyze it for a long-running Keycloak instance with different traffic patterns during a day.

See below for some PromQL queries: 

Service-level measurement of requests faster than 250 ms. First rate, then sum. The numerator needs to drop the "le" label, so it has the same number of labels as the denominator.

> sum(rate(http_server_duration_bucket{http_route="/realms/{realm}/protocol/{protocol}/token", le="250.0", namespace="$namespace"}[5m])) without (le) / rate(http_server_duration_count{http_route="/realms/{realm}/protocol/{protocol}/token", namespace="$namespace"}[5m])

(I found a strange result where both counters don't seem to be incremented atomically, so we went above 100% in one occasion)

Similar for the cache for realms. The hit rate is only reset at restart, or when it is manually reset. Therefore use a moving average: 

> rate(vendor_cache_manager_keycloak_cache_realms_statistics_hits{namespace="${namespace}"}[5m]) / (rate(vendor_cache_manager_keycloak_cache_realms_statistics_hits{namespace="${namespace}"}[5m]) + rate(vendor_cache_manager_keycloak_cache_realms_statistics_misses{namespace="${namespace}"}[5m]))

Same for users cache: 

> rate(vendor_cache_manager_keycloak_cache_users_statistics_hits{namespace="${namespace}"}[5m]) / (rate(vendor_cache_manager_keycloak_cache_users_statistics_hits{namespace="${namespace}"}[5m]) + rate(vendor_cache_manager_keycloak_cache_users_statistics_misses{namespace="${namespace}"}[5m]))

The response time histogram shows the histogram as reported by each Keycloak instance at the end of the selected time interval. Any restart of a Keycloak during the test would reset the counter for that Keycloak instance. This might be a bit hard to understand for users. Maybe we could come up with a PromQL query that would aggregate all the requests in the selected time window even if a Keycloak instance was restarted. 

To go forward: We could merge it as is (preferred), or without the dashboard (if you think it is not ready even for a draft version). And next week we'll analyze some runs and update the dashboard as necessary.

mhajas:
Thank you @ahus1! I updated the dashboard, and here is how it looks now:
screenshot_20230627_120240

I tried to update the histogram as you suggested, but it is more complicated. After restarting it shows two histograms, meaning there are two different metrics (different instance label). I tried to use aggregation `sum without(instance)` but it didn't work as after the aggregation it shows the latest number. What we would need is to somehow find all drops to 0 and then sum the last number before the drop to current counter. If there is no simple solution I would prefer to not do this as part of this PR.

ahus1 (approved):

PR #387, opened 2023-06-22
Support EFS on ROSA for ReadWriteMany PVCs

### LINKED ISSUE #386: Support EFS on ROSA for ReadWriteMany PVCs
### Description

For downloading the OTEL agent only once, and then share it between PODs, it would be good to use PVCs with ReadWriteMany

### Discussion

_No response_

### Motivation

Be able to use share a PVC between Pods.

### Details

_No response_

ahus1:
There was a successful create run here to create one cluster here: https://github.com/ahus1/keycloak-benchmark/actions/runs/5347697967/jobs/9696556604

There was a successful run to delete the cluster here: https://github.com/ahus1/keycloak-benchmark/actions/runs/5349788947/jobs/9701503197

To test a Pod with EFS based PVCs, there is a script `rosa_efs_test.sh`.

Closes #386

### DIFF
--- a/provision/aws/rosa_delete_cluster.sh
+++ b/provision/aws/rosa_delete_cluster.sh
@@ -8,6 +8,9 @@ fi
 CLUSTER_NAME=${CLUSTER_NAME:-$(whoami)}
 if [ -z "$CLUSTER_NAME" ]; then echo "Variable CLUSTER_NAME needs to be set."; exit 1; fi
 
+# Cleanup might fail eif EFS hasn't been configured for the cluster. Ignore any failures and continue
+./rosa_efs_delete.sh || true
+
 CLUSTER_ID=$(rosa describe cluster --cluster "$CLUSTER_NAME" | grep -oPm1 "^ID:\s*\K\w+")
 echo "CLUSTER_ID: $CLUSTER_ID"
 
--- a/.github/workflows/rosa-cluster-create.yml
+++ b/.github/workflows/rosa-cluster-create.yml
@@ -5,12 +5,12 @@ on:
     inputs:
       clusterName:
         description: 'Name of the cluster'
-        type: text
+        type: string
       computeMachineType:
         description: 'Instance type for the compute nodes'
         required: true
         default: m5.xlarge
-        type: text
+        type: string
       multiAz:
         description: 'Deploy to multiple availability zones in the region'
         required: true
@@ -19,8 +19,8 @@ on:
       replicas:
         description: 'Number of worker nodes to provision'
         required: true
-        default: 2
-        type: text
+        default: '2'
+        type: string
 
 env:
   OPENSHIFT_VERSION: 4.12.19
--- a/.github/workflows/rosa-cluster-delete.yml
+++ b/.github/workflows/rosa-cluster-delete.yml
@@ -5,12 +5,12 @@ on:
     inputs:
       clusterName:
         description: 'Name of the cluster'
-        type: text
+        type: string
       deleteAll:
         description: 'Delete all clusters'
         required: true
         default: 'no'
-        type: text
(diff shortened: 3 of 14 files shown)


ahus1:
@kami619 / @mhajas: Here's the PR to review. The update AsciiDoc contains links to the guides which I used.

EFS is now deployed automatically to all new clusters created. 

cc: @tkyjovsk - pinging you as you're one of the team members with a lot of AWS experience.

provision/aws/rosa_delete_cluster.sh
```diff
@@ -8,6 +8,9 @@ fi
 CLUSTER_NAME=${CLUSTER_NAME:-$(whoami)}
 if [ -z "$CLUSTER_NAME" ]; then echo "Variable CLUSTER_NAME needs to be set."; exit 1; fi
 
+# Cleanup might fail eif EFS hasn't been configured for the cluster. Ignore any failures and continue
```
kami619, line 11: where is the input to control the EFS configuration to be set or not set ?

kami619:
@ahus1 general question on how much additional cluster creation/deletion time due to this ? do we have to adjust our schedules due to that ?

provision/aws/rosa_delete_cluster.sh
ahus1: EFS is the new default, and once all old clusters have been destroyed, all new will have EFS. 

I didn't want to have EFS as an option, as it makes the setup more complex.

kami619: thanks, so line #12 in this script is pretty no-op after all the old rosa clusters are deleted, right ?

ahus1:
@kami619 - Good question. The creation time this adds less than 5 minutes, the deletion time is even less, so I consider this negligible.

kami619:
> @kami619 - Good question. The creation time this adds less than 5 minutes, the deletion time is even less, so I consider this negligible.

thanks, that answers my question.

provision/aws/rosa_delete_cluster.sh
ahus1: Hm, "no-op" means to me: does nothing. I'd say it is quite the opposite: it is a no-op (does nothing) for old clusters. It only does something for the new ones with EFS. 

The current script also handles a situation when EFS creation failed for any reason, and we want to delete that part of the cluster which has been created. So I'd say I'm +1 to keep it as is for now.

kami619:
also @ahus1 would adding any PVC's for postgres pod, would that be out of scope for this PR ? apart from the efs test script, we dont have a consumer from provisioning for this as part of this PR right ?

provision/aws/rosa_delete_cluster.sh
kami619: oh ok, sorry, i got it the other way around, then it makes sense, this is the catch all for any exceptions from the efs creation. thanks and it makes sense to leave it as is.

ahus1:
@kami619 - the OTEL agent would be the first consumer for this. 

If we want to add PVCs for PostgreSQL, this would be out of scope for this PR. For PostgreSQL we would rather use ReadWriteOnce, as no two PostgreSQL containers would share a volume. Once we touch that, we'll need to adjust clearing the PostgreSQL database on Keycloak reset. 

Please file a new issue for this.

kami619 (approved):
I think I am good with the functionality of the script, I am not well versed with the aws cli usage standards, but I think we can tune that as we find issues., but would like @tkyjovsk to give this a swirl if he has some time, if not we can merge it, given we have two successful runs.

kami619:
alright merging this for now. @tkyjovsk please add any feedback retroactively or create issues on github if you find something is not upto the mark.

PR #390, opened 2023-06-23
Archive rosa logs for creating and deleting clusters

mhajas:


ahus1 (changes requested):
I assume we want to do this even if the previous action fails. The way it is now, it will only archive them on success.

mhajas:
Thank you @ahus1! It makes sense, I updated the PR.

ahus1 (approved):
Thank you - LGTM!

PR #395, opened 2023-06-26
Do not use colon in archived log files

mhajas:
Any idea for a nicer fix? I think it is important to have the date and time there.

### DIFF
--- a/provision/aws/rosa_create_cluster.sh
+++ b/provision/aws/rosa_create_cluster.sh
@@ -56,14 +56,18 @@ fi
 
 mkdir -p "logs/${CLUSTER_NAME}"
 
+function custom_date() {
+    echo "$(date '+%Y%m%d-%H%M%S')"
+}
+
 echo "Creating operator roles."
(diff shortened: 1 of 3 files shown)


provision/aws/rosa_create_cluster.sh
```diff
@@ -56,14 +56,18 @@ fi
 
 mkdir -p "logs/${CLUSTER_NAME}"
 
+function custom_date() {
+    echo "$(date -uIs | sed -e "s/:/-/g")"
```
ahus1, line 60: You might be looking for something like this:

```suggestion
    echo "$(date '+%Y%m%d-%H%M%S')"
```

Same below for the other date.

ahus1 (changes requested):
See comment below for a formatting of a date which is a bit different.

provision/aws/rosa_create_cluster.sh
mhajas: Thanks, done

ahus1 (approved):
LGTM!

PR #402, opened 2023-06-29
updates for optimizing failure scenario

### LINKED ISSUE #393: Optimise failure scenario
- When killing a pod, wait for it to be up again to kill another
  - measure the time it takes for the pod to come up, and possibly also capture the log just after it became ready, so we have both the timing and the messages
- Save the logs before killing the pod (in .zip file using command line)
- Pulling all KC logs before end of the test run
- Capture the complete pod resource before killing and end of run
- Find bottlenecks

kami619:
fixes #393 

Client Secret Failure run -https://github.com/kami619/keycloak-benchmark/actions/runs/5402389905/jobs/9813602924

Auth Code Failure run - https://github.com/kami619/keycloak-benchmark/actions/runs/5404050859/jobs/9817708034

----
co-author : @mhajas

### DIFF
--- a/benchmark/src/main/content/bin/kc-chaos.sh
+++ b/benchmark/src/main/content/bin/kc-chaos.sh
@@ -3,15 +3,18 @@
 set -e
 
 : ${INITIAL_DELAY_SECS:=30}
-: ${CHAOS_DELAY_SECS:=60}
+: ${CHAOS_DELAY_SECS:=10}
 : ${PROJECT:="runner-keycloak"}
 
-echo -e "\033[0;31mINFO:$(date '+%F-%T-%Z') Entering Chaos mode, with an initial delay of $INITIAL_DELAY_SECS seconds"
+LOGS_DIR=$1
+
+echo -e "\033[0;31mINFO:$(date '+%F-%T-%Z') Entering Chaos mode, with an initial delay of $INITIAL_DELAY_SECS seconds\033[0m"
 sleep $INITIAL_DELAY_SECS
-echo -e "INFO:$(date '+%F-%T-%Z') Running Chaos scenario - Delete random Keycloak pod"
+echo -e "\033[0;31mINFO:$(date '+%F-%T-%Z') Running Chaos scenario - Delete random Keycloak pod\033[0m"
+
+ATTEMPT=0
 while true; do
-
-
+  ATTEMPT=$[ATTEMPT + 1]
   RANDOM_KC_POD=$(kubectl \
     -n "${PROJECT}" \
     -o 'jsonpath={.items[*].metadata.name}' \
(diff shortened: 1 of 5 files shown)


benchmark/src/main/content/bin/kcb.sh
```diff
@@ -34,6 +34,8 @@ JAVA_OPTS="${JAVA_OPTS} -Xmx1G -XX:+HeapDumpOnOutOfMemoryError"
 DEBUG_MODE="${DEBUG:-false}"
 DEBUG_PORT="${DEBUG_PORT:-8787}"
 
+CHAOS_MODE="${CHAOS_MODE:-false}"
```
mhajas, line 37: For distributed runners, this mode is not very usable as we don't want to run the chaos script from all runners, but only once.

We can move the trigger back to the workflow (the log directory can be obtained also there) and then we can print the info from the pod at the end of kcb script from another workflow step. This way we would also get rid of coupling of kcb script with Openshift. Any thoughts @kami619 @ahus1 @tkyjovsk?

benchmark/src/main/content/bin/kc-chaos.sh
```diff
@@ -3,25 +3,41 @@
...
     -o 'jsonpath={.items[*].metadata.name}' \
     get pods -l app=keycloak | \
       tr " " "\n" | \
       shuf | \
       head -n 1)
-  echo -e "\033[0;31mINFO:$(date '+%F-%T-%Z') Killing Pod '${RANDOM_KC_POD}' and waiting for ${CHAOS_DELAY_SECS} seconds"
+
+  kubectl get pods -n "${PROJECT}" -l app=keycloak -o wide
+  kubectl logs -f -n "${PROJECT}" "${RANDOM_KC_POD}" > "$LOGS_DIR/${ATTEMPT}-${RANDOM_KC_POD}.log" 2>&1 &
+  kubectl describe -n "${PROJECT}" pod "${RANDOM_KC_POD}" > "$LOGS_DIR/${ATTEMPT}-${RANDOM_KC_POD}-complete-resource.log" 2>&1
```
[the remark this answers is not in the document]
kami619: yup, we had this in mind to parse the information and create a json file akin to what you did for SUT information, but as the first pass we are storing in the log files. I would suggest at-least this part to be done in a subsequent PR to speed things up.

benchmark/src/main/content/bin/kcb.sh
kami619: thanks for the review comments everyone. given the majority feedback leans towards modularization of this, I shall work with Michal tomorrow to come up with a better design and implement that.

kami619: in order to keep the scope intact and not let the design decisions take up more time for this PR to get merged. we are going to merge this as is created a [ticket](https://github.com/keycloak/keycloak-benchmark/issues/405) to tackle the modularization of script and solving the puzzle with distributed load tests in mind as the next increment.

PR #413, opened 2023-07-04
Setup OpenShift logging to capture all logs

### LINKED ISSUE #409: Setup OpenShift logging to capture all logs (even of Container restarts)

ryanemerson:
Enabling by default is increasing memory usage by ~7GB.

Closes #409


doc/kubernetes/modules/ROOT/pages/installation-openshift.adoc
```diff
@@ -66,6 +66,14 @@ http://grafana.apps.**<domain name>**
 NOTE: This is a shared resource for all users using the OpenShift cluster.
 While the next section describes how to install multiple Keycloaks in different namespaces, this doesn't apply to the Grafana instance.
 
+== Openshift Logging
+Openshift logging is enabled by default. All application and infrastructure pod logs are stored in a non-replicated
```
ahus1, line 70: The way you set this up is that OpenShift logging is only enabled after "task monitoring" has been executed, therefore it is not enabled by default. 

Do we want it to be enabled by default? Then it should probably be called from "rosa_create_cluster.sh".

provision/openshift/logging.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: operators.coreos.com/v1alpha1
```
ahus1, line 1: The ElasticSearch operator failed to start on a fresh installation on gh-keycloak with the message: 

> failed to populate resolver cache from source operatorgroup-unavailable/openshift-operators-redhat: found 0 operatorgroups in namespace openshift-operators-redhat: expected 1

In a previous setup, if you installed it via the UI first, that group might have already existed. See PR https://github.com/keycloak/keycloak-benchmark/pull/415 for such an operator group (usually they should automatically named, I picked a static name for this instance)

ahus1 (changes requested):
Thanks for providing this change. I ran into some obstacles, and also have some questions - please see below.

doc/kubernetes/modules/ROOT/pages/installation-openshift.adoc
ryanemerson: Argh my bad, I made the assumption that the monitoring task was always called. I've updated it so that logging is installed as part of `rosa_create_cluster.sh` as I assume we will want this for all benchmark runs in the event of unexpected states being encountered.

ryanemerson:
Thanks for the review @ahus1, updated.

ahus1 (approved):
Thank you for this change to have a minimal logging available in OpenShift. 

I updated the docs, and also enabled the console plugin which allows direct access to the logs of a pod.

PR #418, opened 2023-07-10
Fix user permissions in the `benchmark` Ansible role.

tkyjovsk:
Replaced hard-coded `ec2-user` with dynamic Ansible value `{{ ansible_ssh_user }}`.


ahus1 (approved):
Tested it, and it works for me.

PR #425, opened 2023-07-12
Do not use spot instances

mhajas:
When we use spot instances sometimes a worker node is removed and all pods running on that node are deleted. This can result in removing more than 1 pod, which means we are losing some sessions. This is something we cannot afford in our scalability testing, therefore removing the switch. 

I am running Keycloak with 600k user sessions at the moment with regular instances (no spot), and everything works well while with spot instances there were regular pod restarts.

### DIFF
--- a/provision/aws/rosa_create_cluster.sh
+++ b/provision/aws/rosa_create_cluster.sh
@@ -87,7 +87,7 @@ echo
 
 ./rosa_recreate_admin.sh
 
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
I agree on this, as such a node failure will render an ongoing load test invalid and non-reproducible. In addition to that, we've seen days and hours where we couldn't spin up any spot instances because none where available at that time in that region, which slows us down in our project efforts.

PR #431, opened 2023-07-18
Setup an Aurora PostgreSQL database

### LINKED ISSUE #420: Setup an Aurora PostgreSQL database
To test two Keycloak instances in different OpenShift environments, they need to connect to a shared database. A PostgreSQL type Aurora database would be an approach here.

As this is to enable the #417, a minimal setup is sufficient (no replicas needed). 

Assumption: Instead of a PostgreSQL database deployed on OpenShift, this points to a database inside the same region, assuming that both clusters run in the same region.

ryanemerson:
This PR allows for an AWS Aurora DB cluster to be created in a specified region and Keycloak can then be deployed to connect to that. I have made it so that Aurora DB can be deployed in a different region to the Keycloak instance, as this will be required for xsite deployments eventually.

My implementation is based upon [this blog](https://cloud.redhat.com/blog/using-vpc-peering-to-connect-an-openshift-service-on-an-aws-rosa-cluster-to-an-amazon-rds-mysql-database-in-a-different-vpc) so that the DB is not exposed publicly and is limited to ROSA VPC -> Aurora VPC communication.

GH Action Workflow:

1. `Aurora Create` - Creates an Aurora DB cluster with a single instance in the specified region
2. `Keycloak - Create deployment` - Updated so that a peering connection is established between the Rosa Cluster VPC -> Aurora VPC, as well as updating the Keycloak deployment to use the Aurora endpoint for DB connections via an External K8s Service.
3. `Keycloak - Delete deployment` - Updated to remove the peering connection
4. `ROSA Cluster - Delete` - Also updated to remove peering connections

Unfortunately with 2. I was hitting the max number of input fields allowed for a GH action (10), so I had to hack around this by providing multiple Aurora fields as JSON. Suggestions on better ways to approach this are very welcome :slightly_smiling_face: 

Couple of things missing still:
- [x] Ability to reset Aurora DB
- [x] Add Aurora delete to Reaper action
- [x] Documentation

Closes #420


doc/kubernetes/modules/ROOT/pages/storage/aurora-postgres.adoc
```diff
@@ -0,0 +1,95 @@
...
+AURORA_REGION= # The AWS region hosting the Aurora cluster
+----
+
+This creates the necessary VPCs, subnets and routes required by an Aurora cluster as well as a single Aurora instance
+for said cluster. The script waits until both the cluster and instance are available. If the cluster already exists,
+a message indiciating this is displayed and the script will fail with exit code 1.
+
+== Connecting ROSA cluster to Aurora Cluster
+
+In order to follow security best practices, the Aurora cluster is not exposed publicly over the internet. Instead, it'sa
```
tkyjovsk, line 22: "it'sa" — typo

doc/kubernetes/modules/ROOT/pages/storage/aurora-postgres.adoc
```diff
@@ -0,0 +1,95 @@
...
+AURORA_REGION= # The AWS region hosting the Aurora cluster
+CLUSTER_NAME= # The name of the ROSA cluster to establish the peering connectin with
+AWS_REGION= # The AWS region hosting the ROSA cluster
+----
+
+== Deleting an Aurora Cluster
+Before deleting an Aurora cluster it's first necessary for all Peering Connections established with ROSA cluster(s) to
+be removed.
+
+To remove an Aurora cluster, execute `./provision/aws/rds/aurora_create.sh` with the following env:
```
tkyjovsk, line 89: `aurora_create.sh` → `aurora_delete.sh`

provision/aws/rds/aurora_create_peering_connection.sh
```diff
@@ -0,0 +1,87 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source ${SCRIPT_DIR}/aurora_common.sh
+
+sh ${SCRIPT_DIR}/../rosa_oc_login.sh
+
+export AWS_REGION=$(rosa describe cluster -c ${CLUSTER_NAME} -o json | jq -r .region.id)
+
+AURORA_VPC=$(aws ec2 describe-vpcs \
+  --filters "Name=cidr,Values=${AURORA_VPC_CIDR}" \
```
tkyjovsk, line 16: This will look up all VPCs in the region matching the `${AURORA_VPC_CIDR}`. This will be problematic if 2 or more users provision Aurora in the same region with the same CIDR, eg. the default one.

Could we tag the VPCs with `${AURORA_CLUSTER}` name instead, and then look up via the tag? This should be unique within the region.

Same would apply for `aurora_delete_peering_connection.sh` and `aurora_delete.sh` scripts.

provision/aws/rds/aurora_common.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+if [ -f ./.env ]; then
+  source ./.env
+fi
+
+export AURORA_CLUSTER=${AURORA_CLUSTER:-"keycloak"}
+export AURORA_INSTANCE=${AURORA_INSTANCE:-"${AURORA_CLUSTER}-instance-1"}
```
tkyjovsk, line 9: Looks like there are some constraints for values of `db-instance-identifier`. For example it looks like it is not possible to use `_`. If the default is used here it puts the same constraints on the value of the `AURORA_CLUSTER` variable. Maybe we should put a note in the docs.

provision/aws/rds/aurora_create.sh
```diff
@@ -0,0 +1,86 @@
...
+)
+if [ -n "${EXISTING_INSTANCE}" ]; then
+  echo "Aurora instance '${AURORA_INSTANCE}:${AWS_REGION}' already exists"
+  exit 1
+fi
+
+# Create the Aurora VPC
+AURORA_VPC=$(aws ec2 create-vpc \
+  --cidr-block ${AURORA_VPC_CIDR} \
+  | jq -r '.Vpc.VpcId'
```
tkyjovsk, line 25: The `aws` commands which are piping into `jq` should explicitly use `--output json`.

provision/aws/rds/aurora_create_peering_connection.sh
ryanemerson: > Could we tag the VPCs with `${AURORA_CLUSTER}` name instead, and then look up via the tag? This should be unique within the region.

Good point,  I'll add a tag.

provision/aws/rds/aurora_common.sh
ryanemerson: I've added an explanation about the Aurora cluster naming restrictions. I couldn't find details of restrictions on db-instance naming in [this doc](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.CreateInstance.html#Aurora.CreateInstance.Settings), the only restriction on underscores I could find was with regards to the DB name.

ryanemerson:
Working as expected when an Aurora cluster is created in the same region as the Keycloak ROSA cluster, however there is an issue if these differ. Investigating..

provision/aws/rds/aurora_create.sh
```diff
@@ -0,0 +1,92 @@
...
+    --engine ${AURORA_ENGINE} \
+    --master-username ${AURORA_USERNAME} \
+    --master-user-password ${AURORA_PASSWORD} \
+    --vpc-security-group-ids ${AURORA_SECURITY_GROUP_ID} \
+    --db-subnet-group-name ${AURORA_SUBNET_GROUP_NAME}
+
+aws rds create-db-instance \
+  --db-cluster-identifier ${AURORA_CLUSTER} \
+  --db-instance-identifier ${AURORA_INSTANCE} \
+  --db-instance-class db.t4g.large \
```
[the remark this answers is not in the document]
tkyjovsk: It might also be useful to have the `--engine-version` under direct control. @ahus1 @kami619 WDYT?

provision/aws/rds/aurora_create.sh
```diff
@@ -0,0 +1,92 @@
...
+  --tag-specifications "ResourceType=vpc, Tags=[{Key=AuroraCluster,Value=${AURORA_CLUSTER}}]" \
+  --output json \
+  | jq -r '.Vpc.VpcId'
+)
+
+# Create the Aurora Subnets
+SUBNET_A=$(aws ec2 create-subnet \
+  --availability-zone "${AWS_REGION}a" \
+  --vpc-id ${AURORA_VPC} \
+  --cidr-block 192.168.0.0/19 \
```
tkyjovsk, line 34: Looks like the CIDR spec is hardcoded here. If I override the default value of `AURORA_VPC_CIDR` this will fail with
```
An error occurred (InvalidSubnet.Range) when calling the CreateSubnet operation: The CIDR '192.168.0.0/19' is invalid.
```

Same for SUBNET_B below.

provision/aws/rds/aurora_create.sh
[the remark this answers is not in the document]
ryanemerson: I can add the `--engine-version` to the Aurora cluster no problem, as well as making the `--db-instance-class` configurable. However it's not possible to configure `--allocated-storage`, `--storage-type`, `--iops` as these are not applicable to Aurora DB instances.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Overview.StorageReliability.html

ryanemerson:
@mhajas Inter-region VPC is working as expected now.

mhajas (changes requested):
Thanks @ryanemerson, works for me except for one small issue in Taskfile. See my comment in the code.

ahus1:
@ryanemerson / @mhajas - can this PR be merged with the change that Michal suggested?

mhajas:
@ahus1 I am ok with merging after my suggestion is addressed. 

Maybe a nice to have would be adding more logging into the scripts to have more insight into what is happening. However, I would say we can add this later if we encounter some problems.

ryanemerson:
I've accepted the commit. I'm happy for it to be merged as is and we can improve logging etc as required.

mhajas (approved):
Nice work @ryanemerson!!

PR #434, opened 2023-07-19
refresh token load test scenario

ddurham2:
refresh token load test scenario

### DIFF
--- a/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
@@ -245,9 +245,43 @@ class KeycloakScenarioBuilder {
           status.is(200),
           // other elements like the access token not captured as we don't need them in the current scenarios and want to save memory
           jsonPath("$..id_token").find.saveAs("idToken"),
+          jsonPath("$..access_token").find.saveAs("accessToken"),
+          jsonPath("$..refresh_token").find.saveAs("refreshToken"),
+          jsonPath("$..expires_in").find.saveAs("expiresIn"),
         )
       )
+      // we must re-write these without the 'Secure' attribute, because otherwise it won't be submitted in future requests  (TODO -- or do we? this was a guess at the problem)
+      .exec(getCookieValue(CookieKey("AUTH_SESSION_ID").withDomain("192.168.49.2").withPath("/auth/realms/test-realm/").withSecure(true)))
+        .exec(addCookie(Cookie("AUTH_SESSION_ID", session => session("AUTH_SESSION_ID").as[String]).withDomain("192.168.49.2").withPath("/auth/realms/test-realm/")))
+      .exec(getCookieValue(CookieKey("KEYCLOAK_SESSION").withDomain("192.168.49.2").withPath("/auth/realms/test-realm/").withSecure(true)))
+        .exec(addCookie(Cookie("KEYCLOAK_SESSION", session => session("KEYCLOAK_SESSION").as[String]).withDomain("192.168.49.2").withPath("/auth/realms/test-realm/")))
+      .exec(getCookieValue(CookieKey("KEYCLOAK_IDENTITY").withDomain("192.168.49.2").withPath("/auth/realms/test-realm/").withSecure(true)))
+        .exec(addCookie(Cookie("KEYCLOAK_IDENTITY", session => session("KEYCLOAK_IDENTITY").as[String]).withDomain("192.168.49.2").withPath("/auth/realms/test-realm/")))
       .exec(session => (session.removeAll("code")))
+      .exitHereIfFailed
+    this
+  }
+
+  def refreshToken(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .exec(http("RefreshToken")
+        .post(TOKEN_ENDPOINT)
+        .headers(UI_HEADERS)
+        .formParam("grant_type", "refresh_token")
+        .formParam("refresh_token", "${refreshToken}")
+        .formParam("client_id", "${clientId}")
+        //.formParam("client_secret", "${clientSecret}")
+        //.formParam("redirect_uri", "${redirectUri}")
+        .check(
+          status.is(200),
+          // other elements like the access token not captured as we don't need them in the current scenarios and want to save memory
+          jsonPath("$..id_token").find.saveAs("idToken"),
+          jsonPath("$..access_token").find.saveAs("accessToken"),
+          jsonPath("$..refresh_token").find.saveAs("refreshToken"),
+          jsonPath("$..expires_in").find.saveAs("expiresIn"),
+        )
+      )
+      //.exec(session => (session.removeAll("code")))
       .exitHereIfFailed
     this
   }
--- a/benchmark/src/main/scala/keycloak/scenario/authentication/RefreshToken.scala
+++ b/benchmark/src/main/scala/keycloak/scenario/authentication/RefreshToken.scala
@@ -1,0 +1,14 @@
+package keycloak.scenario.authentication
+
+import keycloak.scenario.{CommonSimulation, KeycloakScenarioBuilder}
+
+class RefreshToken extends CommonSimulation {
+
+  setUp("Authentication - Refresh Token", new KeycloakScenarioBuilder()
+    .openLoginPage(true)
+    .loginUsernamePassword()
+    .exchangeCode()
+    .refreshToken()
+    .logout(true))
+
+}


benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala
```diff
@@ -245,13 +245,47 @@ class KeycloakScenarioBuilder {
           status.is(200),
           // other elements like the access token not captured as we don't need them in the current scenarios and want to save memory
           jsonPath("$..id_token").find.saveAs("idToken"),
+          jsonPath("$..access_token").find.saveAs("accessToken"),
+          jsonPath("$..refresh_token").find.saveAs("refreshToken"),
+          jsonPath("$..expires_in").find.saveAs("expiresIn"),
         )
       )
+      // we must re-write these without the 'Secure' attribute, because otherwise it won't be submitted in future requests  (TODO -- or do we? this was a guess at the problem)
+      .exec(getCookieValue(CookieKey("AUTH_SESSION_ID").withDomain("192.168.49.2").withPath("/auth/realms/test-realm/").withSecure(true)))
```
kami619, line 254: looks like we are assuming `test-realm` to be the only realm to fetch the cookie from ? and what is the reason behind setting specific cookies in the session ?

ddurham2: it's in the associated Issue about the failing refresh operation, but that's fixed now and this was just temporary code that's been removed now

kami619: great, thanks for taking care of that.

kami619:
@ddurham2 I am currently testing this locally, do you think anything else is left on this PR before we can take this forward ?

ddurham2:
Yes, I'm still tweaking it locally.  I'm both doing tests and adjusting KC and adjusting the tests as I go.  I did get it up to 100k on a local minikube deployment.  I still need to figure out how to specify some of the testing parameters which are new for this test, I think.  
I'll try and commit some things shortly.

benchmark/src/main/content/bin/kcb.sh
```diff
@@ -29,7 +29,7 @@ DIRNAME=$(dirname "$RESOLVED_NAME")
 
 # Default values
 JAVA_OPTS="-server"
-JAVA_OPTS="${JAVA_OPTS} -Xmx1G -XX:+HeapDumpOnOutOfMemoryError"
+JAVA_OPTS="${JAVA_OPTS} -Xmx4G -XX:+HeapDumpOnOutOfMemoryError"
```
ddurham2, line 32: I saw the gatling java process up in the 3.5GB range when ramping up to 100k users

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
```diff
@@ -27,6 +27,10 @@ abstract class CommonSimulation extends Simulation {
     var default = http
       .acceptHeader("application/json")
       .disableFollowRedirect
+      .shareConnections
+      .connectionHeader("keep-alive")
+      .useAllLocalAddresses
+      .disableCaching
```
ddurham2, line 33: more real world?

https://gatling.io/docs/gatling/reference/current/http/protocol/#disablecaching

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
```diff
@@ -27,6 +27,10 @@ abstract class CommonSimulation extends Simulation {
     var default = http
       .acceptHeader("application/json")
       .disableFollowRedirect
+      .shareConnections
+      //.connectionHeader("keep-alive")
+      .useAllLocalAddresses
```
ddurham2, line 32: In getting this test up to 100k I ran into networking issues beyond around 30k.

This was due to running out of src ports on a single local IP in use.  

At different times I saw different things on netstat...

At one point, I saw an exhaustion of ports due to 30k TIME_WAIT sockets.

Then I enabled [sharedConnections](https://gatling.io/docs/gatling/reference/current/http/protocol/#shareconnections), and things seem to be improved by it only keeping around 4-5k connections open.

But then that quit being its behavior for some reason and I would see 10s of thousands of ESTABLISHED connections.  And that would happen with or without keep-alive.

I believe it's probably affected by how well the server can keep up.  If it can service the requests faster than gatling needs to create a new request, then (with shareConnections enabled) it can reuse a connection from the pool.  But if the server can't keep up, it's a cascading problem because gatling must make a new connection to the server, incurring some extra load.

So, further, I enabled useAllLocalAddresses so that I could define new srcIPs for it to have available so there can be more than the 30k (based on ephemeral port range) between a single local IP and the single server IP/port.
e.g. ```
            ip addr add 192.168.40.1/32 dev lo label lo:40
            ip addr add 192.168.41.1/32 dev lo label lo:41
            ip addr add 192.168.42.1/32 dev lo label lo:42
            ip addr add 192.168.43.1/32 dev lo label lo:43
            ip addr add 192.168.44.1/32 dev lo label lo:44
```

Also, on my system I set `sysctl net.ipv4.tcp_tw_reuse=1` so that it can make use of TIME_WAIT sockets if necessary.

File: benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala [hunk 5]
```diff
@@ -720,5 +722,33 @@ class KeycloakScenarioBuilder {
...
+          jsonPath("$..refresh_token").find.saveAs("refreshToken"),
+          jsonPath("$..expires_in").find.saveAs("expiresIn"),
+        )
+      )
+      .exitHereIfFailed
+  }
+
+  def repeatRefresh(): KeycloakScenarioBuilder = {
+    chainBuilder = chainBuilder
+      .repeat(10, "refresh_i") {  //ToDo how to parameterize the count
```
ddurham2, line 748: So, any ideas how we can specify this from the outside?  Should we introduce a new config value?

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
kami619: well, wouldn't the real world be actually something like browser caching 
```
Expires header
Cache-Control header
Last-Modified header
ETag
```

ahus1: Hi @ddurham2 - my take on these parameters: 

* shareConnections: This would share connections across multiple clients, that is: different users would use the same HTTP connection. While this allows more requests per load driver, and prevents problems with the TCP_WAIT state of connections, it delivers false results about the target as you won't generate the overhead of a TLS handshake. Therefore I advise against it. 
BTW: As long as you use local addresses (like connecting to a local minikube), the TCP_WAIT problem doesn't occur.

* connectionHeader("keep-alive"): This is actually optional in HTTP/1.1 and not needed to have keep alive working. I recently debugged it, and I don't see a need to transmit it while still getting the keep-alive behavior.

* useAllLocalAddresses: I would be OK to use it, if it was chosen to create more connections from a single host. If this is the case, I'd like to see a comment in the code explaining the reasoning behind it.

* disableCaching: I wouldn't use it, as real browsers would cache. At the same time I'd like to learn more why you've chosen it.

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
ahus1: @ddurham2 - It looks to me as if you're hitting a limit of the local minikube setup. In the parent issue I see that you created 6 pods, and then ran a high load against that setup.

While the local minikube setup is IMHO great to develop scenarios, reliable performance numbers of large systems are only available in a production-like setup, which would require a cloud setup. 

Changing parameters like `shareConnections` might make the scenario pass on minikube in a local setup, but won't show the right load created due to TLS connection reuse. So I'd rather recommend to scale down the scenario when running it on minikube.

For a setup to run a production load against that number of pods, I'd recommend using a cloud setup. 
Unfortunately those test setups then require someone to pay the cloud bill. 

Recently this project added scripts to deploy small EC2 instances to drive the load, see [Running benchmarks via Ansible and EC2](https://www.keycloak.org/keycloak-benchmark/benchmark-guide/latest/run/running-benchmark-ansible). There are also instructions on how to deploy an [OpenShift instance with ROSA to run Keycloak](https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/prerequisite/prerequisite-rosa).

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
kami619: thanks @ahus1 for sharing your view on these parameters, I am not very well versed with all these parameters except `disableCaching` and I agree with your reasoning of using some of it here.

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
ahus1: @ddurham2 - one more thought about `shareConnections` - if you think this is essential for your testing, you could place it behind an option, document it, and keep the default as before.

kami619: I agree with keeping the load parameters low enough to suit the local setup and one of the other future alternatives is to extend the `Ansible playbook` to work with regular on-prem nodes or the cloud provider of your choice., but that is something we wont be prioritizing soon enough.

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
ddurham2: fair enough.

I've adjusted things and put sharedConnections behind a --shared-connections=true argument

benchmark/src/main/scala/keycloak/scenario/CommonSimulation.scala
ddurham2: done

kami619:
@ahus1 - I did the test with both minikube and openshift and also with the ec2 load runners and local benchmark, things looked good to me. do you want to give this a last look, before I merge it ?

kami619 (approved):

ahus1:
Thank you @ddurham2 for contributing this change and @kami619 for reviewing it thoroughly! 

I added one tiny bit in f0d86cb364debe392fe813e20e535219a881830f I remember from a previous life, that one using shared connections, it used to make sense to increase he maxConnectionsPerHost. 

I'm now merging this.

ddurham2:
Thanks guys.  Much appreciated!

PR #436, opened 2023-07-19
Hide security-sensitive data from Ansible logs.

### LINKED ISSUE #435: ec2 load runners creation ansible playbook outputs private rsa key to the console
### Describe the bug

security lapse in the ansible playbook output which documents rsa private key

### Version

### Expected behavior

we should find a way to mask the output for the rsa private key thats use for a particular stage in the ansible playbook

### Actual behavior

When you run `./aws_ec2.sh create <REGION>` it prints something like the below output without masking the rsa private key onto the console in place of `XXXXXXXXXX`

```
TASK [aws_ec2 : Create Key] **********************************************************************************************************************************
changed: [localhost] => {"changed": true, "key": {"fingerprint": "36:33:94:d6:d4:6e:c0:0e:51:21:c7:33:f2:37:33:c4:1b:63:4b:58", "id": "key-04e66811501cbb005", "name": "benchmark_kakella", "private_key": "-----BEGIN RSA PRIVATE KEY-----XXXXXXXXXX \n-----END RSA PRIVATE KEY-----", "tags": {}}, "msg": "key pair created"}
```

### How to Reproduce?

_No response_

### Anything else?

_No response_

tkyjovsk:
Security sensitive data like key pairs, IPs, etc. are now hidden from Ansible logs by default.

For debugging purposes these logs can be explicitly enabled by setting the Ansible variable `no_log_sensitive` to `false`.

Closes #435

### DIFF
--- a/ansible/roles/aws_ec2/defaults/main.yml
+++ b/ansible/roles/aws_ec2/defaults/main.yml
@@ -7,3 +7,5 @@ ami_name: RHEL-8.8.0_HVM-20230503-x86_64-54-Hourly2-GP2
 instance_type: t3.micro
 instance_volume_size: 20
 instance_device: /dev/sda1
+
+no_log_sensitive: yes
--- a/ansible/roles/aws_ec2/tasks/create-resources.yml
+++ b/ansible/roles/aws_ec2/tasks/create-resources.yml
@@ -1,6 +1,7 @@
 - name: Get Ansible Control Host's public IP
   shell: curl -ks ifconfig.me
   register: control_host_ip
+  no_log: "{{ no_log_sensitive }}"
 
 - name: Create Security Group
   amazon.aws.ec2_group:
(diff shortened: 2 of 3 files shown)


kami619 (approved):
Thanks for fixing this 

It works as expected on my local

```
TASK [aws_ec2 : Get Ansible Control Host's public IP] ***************************************************************************************************************************************************************************
changed: [localhost] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}

TASK [aws_ec2 : Create Security Group] ******************************************************************************************************************************************************************************************
changed: [localhost] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}

TASK [aws_ec2 : Create Key] *****************************************************************************************************************************************************************************************************
changed: [localhost] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}

TASK [aws_ec2 : Save Private Key on Ansible Control Machine] ********************************************************************************************************************************************************************
changed: [localhost] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}

```

PR #440, opened 2023-07-21
Adding FD_SOCK2 -> FD_SOCK workaround

### LINKED ISSUE #439: Run tests Infinispan tests with FDSOCK instead of FD_SOCK2
### Describe the bug

As described in https://github.com/keycloak/keycloak/issues/21092, there are constellations where FD_SOCK2 will fail. 

This this applies the workaround described in the issue to isolation us from those kind of failures in our tests until a workaround is available. 

### Version

nightly

### Expected behavior

Run tests without worrying about the existing bug.

### Actual behavior

Pod chaos experiments fail, and while we look for new causes, the old causes shouldn't hold us back.

ahus1:
Closes #439

### DIFF
--- a/provision/minikube/keycloak/config/kcb-infinispan-cache-config.xml
+++ b/provision/minikube/keycloak/config/kcb-infinispan-cache-config.xml
@@ -21,6 +21,21 @@
         xsi:schemaLocation="urn:infinispan:config:14.0 https://www.infinispan.org/schemas/infinispan-config-14.0.xsd"
         xmlns="urn:infinispan:config:14.0">
 
+    <!-- Workaround for https://github.com/keycloak/keycloak/issues/21092 -->
+    <jgroups>
+        <stack name="kubernetes-with-fdsock" extends="kubernetes">
+            <!-- When using an embedded stack, replacement is done by Infinispan which requires environment variables to be
+            prefixed with ".env", while JGroups falls back to environment variables without the prefix.
+            The Keycloak Operator passes this information in an enviornment variable `jgroups.dns.query`.
+            See https://github.com/keycloak/keycloak/issues/21830 for a discussion.
+            -->
+            <dns.DNS_PING dns_query="${env.jgroups.dns.query}" />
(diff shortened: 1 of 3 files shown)


ryanemerson (approved):

ryanemerson:
LGTM. I don't have sufficient permissions to merge though.

kami619:
@ryanemerson where you able to give it a local test ? I can merge the PR if it worked for you.

ryanemerson:
I did and the cluster formed as expected. I just double checked the logs though and `ISPN000901` is still present, so hold off merging and I will investigate if this workaround for 21830 works on a standalone Infinispan instance.

ahus1:
@ryanemerson - I debugged it and found that in DNS_PING the correct query hostname is present. I assume this happens when it evalues the parent stack. 

If you want to enable debugging, add the following to the Keycloak CR:

```
  additionalOptions:
    - name: log-level
      value: "INFO,org.jgroups:debug"
```

ryanemerson:
> I debugged it and found that in DNS_PING the correct query hostname is present. I assume this happens when it evalues the parent stack.

We're still getting the `ISPN000901` warning because the Infinispan parser is processing both the "kubernetes" and the "kubernetes-with-fdsock" stacks. I think we should remove DNS_PING from the "kubernetes-with-fdsock" stack as it's not removing the warning. We can then just wait on the actual fix in the Keycloak operator, as this doesn't affect functionality.

The `FD_SOCK` changes LGTM.

mhajas:
This PR is trying to address an issue with state transfer, is the issue caused by a different problem than: https://github.com/keycloak/keycloak/issues/21256

ryanemerson:
> This PR is trying to address an issue with state transfer, is the issue caused by a different problem than: [keycloak/keycloak#21256](https://github.com/keycloak/keycloak/issues/21256)

Yes, this fixes an issue with `FD_SOCK2` at the JGroups level https://github.com/keycloak/keycloak/issues/21092

ahus1:
Merged as of @ryanemerson's review.

PR #442, opened 2023-07-21
Fix work cache in cross-dc setup

mhajas:


provision/infinispan/create_ispn_clusters.sh
```diff
@@ -214,6 +214,38 @@ spec:
 EOF
 }
 
+function deploy_replicated_cache_cr() {
```
ryanemerson, line 217: Isn't this effectively the same as `deploy_distributed_cache_cr` except that we specify `replicatedCache`? I think it would be simpler to just make the cache type a parameter of the original `deploy_cache_cr` method so that we  don't have to maintain duplicated code.

kami619:
I can see the work cache stats coming up in a non cross-site deployment. LGTM. @mhajas what is the best way to test this against a cross-site setup given that PR is still in progress ?

provision/infinispan/create_ispn_clusters.sh
mhajas: @ryanemerson there is also the number of owners. I would say we can leave this as is now, as we plan to migrate to Helm anyway. WDYT?

ahus1:
@ryanemerson / @mhajas / @kami619 - I saw the discussions above, and I wonder if this can be merged before today's meeting, and we can then plan the next steps?

provision/infinispan/create_ispn_clusters.sh
ryanemerson: Yeah that's fine, it was just a nitpick :+1:

ryanemerson (approved):

ryanemerson:
@ahus1 Sure, works for me.

PR #443, opened 2023-07-25
Add missing checkout step to Aurora reaper

ryanemerson:
The reaper is failing with:

```
/home/runner/work/_temp/c0ca7b2b-7d9b-46f9-840d-25903c252923.sh: line 1: ./provision/aws/rds/aurora_cluster_reaper.sh: No such file or directory
```
This is because the code has not been checked out.


ahus1 (approved):
Thank you for looking into this!

PR #452, opened 2023-07-26
Configure Keycloak worker thread pool to be aligned with JGroup thread pool size

ahus1:
This prepares the setup for the next upcoming load test where we align the Keycloak workers.

@kami619 - please review the docs and test in a single cluster with 100_000 sessions, 200 ups sustained load 100 pod kills (like the previous run where we saw the spikes).

Thanks!


doc/kubernetes/modules/ROOT/pages/load-behavior.adoc
```diff
@@ -32,9 +33,17 @@ When the load and the number of threads increases, the bottleneck will usually b
...
 +
-With the number of threads in the executor pool being an order of magnitude larger than the number of database connections and with requests failing when no database connection is available within the https://quarkus.io/guides/all-config#quarkus-agroal_quarkus.datasource.jdbc.acquisition-timeout[`quarkus.datasource.jdbc.acquisition-timeout`] (5 seconds default), this is an effective https://en.wikipedia.org/wiki/Demand_response#Load_shedding[load-shedding behavior] where it returns an error response instead of queueing requests for an indefinite amount of time.
+With the number of threads in the executor pool being an order of magnitude larger than the number of database connections and with requests failing when no database connection is available within the https://quarkus.io/guides/all-config#quarkus-agroal_quarkus.datasource.jdbc.acquisition-timeout[`quarkus.datasource.jdbc.acquisition-timeout`] (5 seconds default), this is somwhat of a https://en.wikipedia.org/wiki/Demand_response#Load_shedding[load-shedding behavior] where it returns an error response instead of queueing requests for an indefinite amount of time.
```
pruivo, line 36: typo `somwhat` -> somewhat

doc/kubernetes/modules/ROOT/partials/executor-jgroups-thread-calculation.adoc
```diff
@@ -0,0 +1,2 @@
+Due to the bug https://issues.redhat.com/browse/ISPN-14780[ISPN-14780] there is currently no known way to configure the size different from the default JGroups 100 max threads, so assuming a Keycloak cluster with 4 Pods, each Pod shouldn't have more than 25 worker threads.
```
pruivo, line 1: That is not entirely true since you can still configure the thread pool in the extended stack. As an example
```xml
    <jgroups>
        <stack name="kubernetes-with-fdsock" extends="kubernetes">
            <TCP thread_pool.max_threads="${env.JGROUPS_MAX_THREADS:200}" />
            <!-- When using an embedded stack, replacement is done by Infinispan which requires environment variables to be
            prefixed with ".env", while JGroups falls back to environment variables without the prefix.
            The Keycloak Operator passes this information in an enviornment variable `jgroups.dns.query`.
            See https://github.com/keycloak/keycloak/issues/21830 for a discussion.
            -->
            <dns.DNS_PING dns_query="${env.jgroups.dns.query}" />
            <!-- Workaround for problems with FD_SOCK2, which might fail when the other nodes are not ready yet.
            See https://github.com/keycloak/keycloak/issues/21092 -->
            <FD_SOCK stack.combine="REPLACE" stack.position="FD_SOCK2"/>
        </stack>
    </jgroups>
```

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -148,6 +148,18 @@ spec:
             imagePullPolicy: Never
 {{ end }}
             env:
+              # Limiting Keycloak thread pool as it might lead to an exhaustion of the JGroups thread pool.
+              # The combined worker thread pool size of all Keycloak instances shouldn't exceed the size of the JGroups thread pool on each node.
```
mhajas, line 152: I am not following the reasoning behind `all Keycloak instances shouldn't exceed the size of the JGroups thread pool`. Isn't JGroups thread pool configured per pod? Based on that reasoning, shouldn't we set `QUARKUS_THREAD_POOL_MAX_THREADS` to `100`?

ahus1: As I understand it, the Keycloak worker threads use JGroups to connect to other Keycloak instances to access the cache there. Which node they connect to depends on where the session is stored. It might be that at a given moment in time all nodes connect with all their threads to the same Keycloak node, and even then this shouldn't exceed the number of JGroup threads on that node (this is my understanding). 

In practice, we've seen that 200 threads on 6 Keyloak Pods lead to a JGroups thread pool of 100 being exhausted in several occasions. This PR will use 16 Keycloak threads, that the sum of all Keycloak threads will never exceed 100 JGroups threads. Let's see how the system behaves, maybe the optimal values will be somewhere in the middle, as some threads will use local data only (about 33% of them). Let's use this PR and see once the numbers come in.

mhajas: Thanks for the explanation @ahus1, makes sense to me.

ahus1: From yesterday's chaos testing: With 16 worker threads on 6 pods, the JGroup worker pool was at around 60 active threads which is well below the maximum of 100 which could cause problems. It also shows that while the number of workers could be a bit higher, but not much. Keeping it at this number for now in the PR.

kami619:
@ahus1 do you a run based off of the latest changes, if not I think the changes look good from the results we had from yday.

doc/kubernetes/modules/ROOT/partials/executor-jgroups-thread-calculation.adoc
ahus1: @pruivo - I tried this. I debugged that the value was correctly picked up in `StringPropertyReplacer`, but it never made it to the `TCP` instance, where the thread pool was still set up to be 100 threads only. 

So I guess I'll wait until ISPN-14780 fixed in an upcoming release.

image

ahus1:
> Do you think we need a run based off of the latest changes, if not I think the changes look good from the results we had from yday.

@kami619 - I think we're good to merge this one. We should just let everyone know that we now restrict the number of Keycloak worker threads. We might lift that later with the next Infinispan upgrade to 200.

PR #455, opened 2023-07-28
Refine Ansible playbook with max users per node and async execution

ahus1:
This should help when the SSH connection is unstable. 
When running it this way, there are a lot of additional log messages for the polling - I hope you don't mind.

It also catches the problem automatically which was previous only documented, but not enforced.

### DIFF
--- a/ansible/roles/benchmark/tasks/run.yml
+++ b/ansible/roles/benchmark/tasks/run.yml
@@ -13,7 +13,7 @@
   - debug: var=local_results_dir
 
   - set_fact: node_count="{{ groups['benchmark'] | length }}"
-  
+
   - set_fact: kcb_params_workload_per_node="{{ kcb_params }}"
   - when: '"--concurrent-users=" in kcb_params'
     block:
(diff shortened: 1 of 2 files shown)


doc/benchmark/modules/ROOT/pages/run/running-benchmark-ansible.adoc
```diff
@@ -39,12 +39,19 @@ The most important parameters to customize in `env.yml` are:
...
+As the network stack of the load driver can be congested with a lot of connections in the `TIME_WAIT` state, consider using one EC2 instance per 250 users per seconds.
 
 `instance_type`:: Size of instances to be created, see https://aws.amazon.com/ec2/instance-types/[AWS instance types].
 
 === Creating EC2 instances
 
+NOTE: Instances will be bound to the IP address of the system which creates them!
+
+When creating the instances, the public IP address of the host creating the machines is added to the EC2 security group, and only this IP address is allowed to log in to the load drivers via SSH.
+When the public IP address changes you'll need to re-run the `create` command.
```
mhajas, line 51: Is it necessary to delete and then create? Or is it enough to run create?

ahus1: From what I understand and tested, the create is idempotent and will update existing instances of the EC2 security group as needed. So no delete is necessary to refresh the configuration.

ansible/roles/benchmark/tasks/run.yml
```diff
@@ -28,43 +28,50 @@
     block:
     - set_fact: users_per_sec_per_node="{{ (kcb_params | regex_search('--users-per-sec=(\S+)', '\\1') | join('') | int) / (node_count|int) }}"
     - debug: var=users_per_sec_per_node
+    - debug: msg="The users per second per node should not exceed 250, but is {{ users_per_sec_per_node }}. Add more nodes to create the load you asked for."
```
kami619, line 31: should `- debug: msg="The users per second per node should not exceed 250, but is {{ users_per_sec_per_node }}. Add more nodes to create the load you asked for."` this be under the `failed_when` ? because right now it prints this when you run the playbook, which might be little confusing, even though this is a nitpick to some extent.

```TASK [benchmark : debug] ***************************************************************************************************************************
ok: [34.247.166.115] => {
    "users_per_sec_per_node": "100.0"
}

TASK [benchmark : debug] ***************************************************************************************************************************
ok: [34.247.166.115] => {
    "msg": "The users per second per node should not exceed 250, but is 100.0. Add more nodes to create the load you asked for."
}


```

kami619 (approved):
tested both positive and negative flows for users-per-sec condition, it works as expected. Couldn't test the ssh connection in a reliable way, but hope the provided async poll fix works seamlessly. LGTM overall, thanks for improving the playbooks further @ahus1

ansible/roles/benchmark/tasks/run.yml
ahus1: Thank you, I refined the message to be conditional depending on whether it exceeds the limit.

PR #456, opened 2023-07-28
Client failover with Route 53

### LINKED ISSUE #446: Failover of loadbalancer for 2 DCs
Investigate how a cross-DC failover for the loadbalancer can work. 

Possibly summarize the finding in a single slide and present it to the team. Investigate on how this can be scripted for an automated test.

* Evaluate Route 53 or other options

ryanemerson:
I have documented a client failover setup I have working with Route 53. The health endpoints just check to see if a connection is available at the minute, but we will want to make this more robust in the future.

As part of this work I needed to be able to explicitly configure the `spec.hostname.hostname` field on the Keycloak CR, so I have also added a commit detailing that.

Closes #446


ahus1:
Thank you for providing the scripts and the docs. I added one commit to make the Antora build complete by registering the module and moving the navigation file to a place where it is picked up.

If we need this module or if the content can be placed in an existing one, I leave it for @ryanemerson and @mhajas to decide.

ryanemerson:
Thanks for the additional commit @ahus1, I've combined this with my docs commit to cleanup the log. 



> If we need this module or if the content can be placed in an existing one, I leave it for @ryanemerson and @mhajas to decide.

My rationale for adding a new module was that the amount of resilience related content is only going to increase as we start to explore other areas, such as Aurora DB failover etc. So having a separate module would potentially make it easier for us to re-use this content for customer facing docs in the future.

ahus1 (approved):
Thank you @ryanemerson for providing this docs and the update to the scripts to deploy a cross-DC setup. 

As discussed in the review meeting, it would be a great addition to have a more scripted setup so we can have multiple cross-DC setups with that single domain. We discussed to use either prefixes or subdomains for that. 

As agreed I'm now merging this PR so people can start using the single cluster. A second PR would then take care of the automation and updated documentation.

PR #461, opened 2023-08-03
Automate Client Failover

### LINKED ISSUE #446: Failover of loadbalancer for 2 DCs

ryanemerson:
Closes #446


ahus1 (approved):
Hi @ryanemerson - thank you for this nice automation!

I pushed some documentation changes in 7384622ee5c53da35011b2d5c7260e3a55987282 and also added "one-image-to-explain-everything". 

It doesn't allow a loadbalancer to survive the daily reaper; we can discuss in today's meeting if this is necessary and when this should be implemented. 

I agree a random part of the subdomain is helpful to allow multiple test setup in the same OpenShift clusters. At the same time I found the random names a bit difficult as they didn't show me which clusters where involved. So I decided to prefix the two cluster names to the subdomain. Please let me know what you think about this, and it this needs to change.

ryanemerson:
Thanks for all the additions. The image and the subdomain changes are a big improvement :slightly_smiling_face:

Issue #444, opened 2023-07-25
Investigate automation of ROSA scaling benchmark


ahus1:
Understand and  discuss how this can be automated

https://www.keycloak.org/keycloak-benchmark/benchmark-guide/latest/report/rosa-benchmark-key-results

andyuk1986:
So after running/understanding the benchmark and it's steps, having chat with Alexander, gathering here the info discussed:

1. The first step would be to put all the steps into a GitHub action which could run against the shared cluster gh-keycloak. The tricky part is (which is not yet automated) would be to read the metrics from OpenShift once the test is complete. The document could then be update manually first, and later automatically.
2. Start with a single data point (like 200 users per second for authorization code). We would later add more data points (like 50 and 300) to show that the resource utilization is linear. 
3. The script "dataset-import.sh" uses a special URL to create data on the server. There is a second URL which allows a minimal interaction with Keycloak internal caches, for example clearing them, checking if an entry exists, or showing the number of entries. This might come in handy in the later steps of this test. And we'll use it also for our Active/Passive tests to verify if an entry exists on the other datacenter.

andyuk1986:
PR for first part of implementation is sent: https://github.com/keycloak/keycloak-benchmark/pull/462

andyuk1986:
PR for second and final part of implementation is sent: #492 

So what has been done:
1. Workflow is created for running 3 keycloak benchmarks on created EC2 instances for measuring different parameters (vCPU, memory) for user logins per second, credential grants per second and user sessions. 
2. The actions are created which are retrieving corresponding Prometheus metrics before and after the benchmarks are running and then calculate the numbers of supported criterias (user logins per second, credential grants per second and user sessions) for current environment.
3. The JSON file is generated and archived in the end of the workflow, which stores all calculated data. An example of this json document is (example json calculated for benchmarks running for 5 minutes):
```
{
  "Testing memory for creating sessions": {
    "500 MB Memory per pod in 3 node cluster handles:": "114831 Active User Sessions"
  },
  "Testing CPU usage for user logins": {
    "1vCPU per pod in 3 node cluster handles:": "45 User Logins Per Second"
  },
  "Testing CPU usage for client credential grants": {
    "1vCPU per pod in 3 node cluster handles:": "478 Client Credential Grants Per Second"
  }
}
```
4. The workflow also archives the Gatling aggregat report and stats.json files.

PR #462, opened 2023-08-07
Automation of Rosa Scaling Benchmark (#444): Part 1

andyuk1986:
Refers to #444

### DIFF
--- a/.github/actions/ec2-create-instances/action.yml
+++ b/.github/actions/ec2-create-instances/action.yml
@@ -1,0 +1,29 @@
+name: Create EC2 instances
+description: Creates EC2 instances
+
+inputs:
+  region:
+    description: 'The AWS region used to host the EC2 instances.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - id: aws-ec2-requirements-install
+      name: Install the required Ansible AWS collections.
+      shell: bash
+      run: |
+        ./aws_ec2.sh requirements
+        pipx inject ansible-core boto3 botocore
+      working-directory: ansible
+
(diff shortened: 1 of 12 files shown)


andyuk1986:
I am archiving the results directory which contains already aggregated report. Also as a summary I am archiving the stats.json files (saw that in other workflow). Please let me know, if it is necessary and if I need to commit results or summary or both to git repo (also have seen in other workflow).

ahus1:
@andyuk1986 - I did a first review, and had a successful build when I chose just some very small values. See https://github.com/ahus1/keycloak-benchmark/actions/runs/5795138289/job/15706106755

I was wondering if I could remove the `deployMonitoring`, as the metrics would be recorded by Prometheus anyway. Let me know your thoughts.  

I'm now doing another full run to see the result.

Some notes about the changes I did: 
* Actions don't have types, and their defaults are always strings
* Create and delete work without start and stop
* I made the measurement period a parameter, to have faster smoke tests after script changes
* users per realm now matches the number of created entities

andyuk1986:
@ahus1 - thanks a lot for the review. Very nice solution for ansible environment file, thanks for adding it. And yes, the deployMonitoring flag can be removed as prometheus records metrics anyway. I will remove that and I can squash commits to one.

andyuk1986:
I have removed the `deployMonitoring` implementation part. May I squash the commits?

ahus1:
@andyuk1986 - yes, merging would have been ok, but I also have the button on the GitHub UI to squash them during the merge - and this is what I did.

Thank you for your contribution!

PR #468, opened 2023-08-07
Use static password in the Hot Rod endpoint

pruivo:
2 changes included in the PR:
* Removed `KC_ISPN_PORT` since it is not possible to change it. 
* Static password (shared password) set for the Hot Rod Endpoint

### DIFF
--- a/provision/infinispan/ispn-helm/templates/infinispan.yaml
+++ b/provision/infinispan/ispn-helm/templates/infinispan.yaml
@@ -1,4 +1,13 @@
 {{- $namespace := .Values.namespace | required ".Values.namespace is required." -}}
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: connect-secret
+  namespace: {{ $namespace }}
+data:
+  identities.yaml: {{ (.Files.Get "identities.yaml") | replace "_password_" .Values.hotrodPassword | b64enc }}
+---
 apiVersion: infinispan.org/v1
 kind: Infinispan
 metadata:
@@ -17,6 +26,8 @@ spec:
     categories:
       org.infinispan: {{ .Values.logging.infinispan }}
       org.jgroups: {{ .Values.logging.jgroups }}
+  security:
+    endpointSecretName: connect-secret
   service:
     type: DataGrid
     {{ if .Values.crossdc.enabled }}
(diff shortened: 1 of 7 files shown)


provision/openshift/Taskfile.yaml
```diff
@@ -198,19 +195,16 @@ tasks:
         if [ "{{.KC_ISPN_NAMESPACE}}" != "" ]; then
         echo "true" > .task/remote-store-enabled
         echo "infinispan.{{.KC_ISPN_NAMESPACE}}.svc" > .task/remote-store-host
-        oc -n {{.KC_ISPN_NAMESPACE}} get secrets infinispan-generated-secret -o 'jsonpath={.data.identities\.yaml}' | base64 -d | yq .credentials[0].username > .task/remote-store-username
-        oc -n {{.KC_ISPN_NAMESPACE}} get secrets infinispan-generated-secret -o 'jsonpath={.data.identities\.yaml}' | base64 -d | yq .credentials[0].password > .task/remote-store-password
+        oc -n {{.KC_ISPN_NAMESPACE}} get secrets connect-secret -o 'jsonpath={.data.identities\.yaml}' | base64 -d | yq .credentials[0].password > .task/remote-store-password
```
pruivo, line 198: Should I keep fetching the password from the secret or convert this toa variable?

kami619: is this a secret generated by the operator ? @pruivo how much of an effort would it be to fetch it from the IAM manager ? `--secret-id "$KEYCLOAK_MASTER_PASSWORD_SECRET_NAME"`

provision/infinispan/ispn-helm/templates/infinispan.yaml
```diff
@@ -1,4 +1,13 @@
 {{- $namespace := .Values.namespace | required ".Values.namespace is required." -}}
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: connect-secret
```
pruivo, line 6: @kami619 the secret is generated here with the password from AWS.

kami619: great, thanks

kami619: in this case we don't need it to be a variable.

kami619 (approved):
lgtm

PR #472, opened 2023-08-09
Ciam 6580 issue 471 hiearchical groups

### LINKED ISSUE #471: Extend DatasetResourceProvider to create hierarchical groups with many subgroups
### Description

We have observed some performance issues with Admin Console groups where there are many subgroups: Paging through the subgroups has some issues while paging through top level groups were fine. This may be related to the fact that DatasetResourceProvider only created top level groups and this case is well tested. 

### Discussion

_No response_

### Motivation

We would like to be able create hierarchical  group structures to help us test with various scenarios with Admin Console/Groups navigation and search.

### Details

The paging of the subgroups issue has mostly been addressed by[ #21520](https://github.com/keycloak/keycloak/pull/21520/) We still would like to have a way to create data that will help testing performance of partial searches on group names and their UI components. A similar issue exists with Admin Console/Identity Providers but this issue is to address only the groups.

We are proposing to extend the DatasetResourceProvider with an extra boolean parameter in " createGroups" method.  This boolean parameter could be called "hiearchicalGroups" and would have a default value of false to ensure backward compatibility. If "hiearchicalGroups" is "true" than the first group created would the be parent group of the rest of the groups created. This would mimic the one case that lead to performance issues in our experience.  Alternatively,  we can also define a depth parameter of which value would denote the first number of groups to be direct subgroup of the prior group, and the remaining groups being the leaf subgroups  of the "depth"th group. The "depth" would have a default value of "1" meaning that createGroups method would create only on parent group; if the "hiearchicalGroups" is false then the "depth" parameter would not be operative.

eatikrh:
This PR creates hierarchical groups using the configuration parameters: groups-count-each-level, groups-hierarchy-depth, groups-with-hierarchy.  Boolean parameter groups-with-hierarchy is "false" by default. With the default values it works as before. 

With groups-with-hierarchy=true, it ignores groups-per-realm parameter and creates the group tree structure as defined by the other parameters. This will be groups-count-each-level ^ groups-hierarchy-depth. 

The implementation honors groups-per-transaction.

We adopted a subgroup naming convention that uses "." in the group names which allows finding the parent group even if it was created in a previous transaction.

Closes #471

### DIFF
--- a/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
+++ b/dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
@@ -33,7 +33,19 @@ import org.keycloak.events.Event;
 import org.keycloak.events.EventStoreProvider;
 import org.keycloak.events.EventType;
 import org.keycloak.http.HttpRequest;
-import org.keycloak.models.*;
+import org.keycloak.models.AuthenticatedClientSessionModel;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.GroupModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.models.KeycloakUriInfo;
+import org.keycloak.models.PasswordPolicy;
+import org.keycloak.models.RealmModel;
+import org.keycloak.models.RealmProvider;
+import org.keycloak.models.RoleModel;
+import org.keycloak.models.UserCredentialModel;
+import org.keycloak.models.UserModel;
+import org.keycloak.models.UserSessionModel;
 import org.keycloak.models.cache.CacheRealmProvider;
 import org.keycloak.models.utils.KeycloakModelUtils;
 import org.keycloak.protocol.oidc.OIDCAdvancedConfigWrapper;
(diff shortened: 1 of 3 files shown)


dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
```diff
@@ -972,12 +979,49 @@ private void createClients(RealmContext context, Task task, KeycloakSession sess
         task.debug(logger, "Created %d clients in realm %s", context.getClientCount(), context.getRealm().getName());
     }
 
-    private void createGroups(RealmContext context, int startIndex, int endIndex, KeycloakSession session) {
+    private String getGroupName(boolean hiearchicalGroups, int hierarchyDepth, int countGroupsAtEachLevel, String prefix, int currentCount) {
```
ahus1, line 982: The parameter `hierarchyDepth` is not used, please review/test the code.

ahus1 (changes requested):
Hi @eatikrh - thank you for providing this PR. I pushed one review commit with aligns some code formatting and avoids star imports. 

I also found that 

```
name1.split(GROUP_NAME_SEPARATOR);
```

might not lead to the results you want to see, as the parameter is a regex. Passing in a dot (`.`) would match any character. I've suggested

```
name1.split(Pattern.quote(GROUP_NAME_SEPARATOR));
```

instead. 

There is also one unused parameter. Please have a look. 

Once you're ok with the code, please have a team member of your team test this code if it delivers the results you want to see (I've seen you've discussed it with @mposolda in the parent issue), and then let them add a comment/review here. Then it can be merged by one of the maintainers of this project.

Thanks!

dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
eatikrh: Hi @ahus1
Thank you for the review and the fixes. Yes "." is tricky in regex. I removed the unused parameter and a typo.

eatikrh:
@mposolda can please also take a look at this PR? Also do we have a Code Style template for IntelliJ? 
Thanks

ahus1:
@eatikrh - there is no code style for this ... once this PR is merged, I'll see that we add a minimal Checkstyle configuration.

dataset/src/main/java/org/keycloak/benchmark/dataset/config/DatasetConfig.java
```diff
@@ -119,6 +119,16 @@ public class DatasetConfig {
     @QueryParamIntFill(paramName = "groups-per-transaction", defaultValue = 100, operations = { CREATE_REALMS })
     private Integer groupsPerTransaction;
 
+    @QueryParamFill(paramName = "groups-with-hierarchy", defaultValue = "false", operations = { CREATE_REALMS })
```
mposolda, line 122: Is it possible to add some small comments to the 3 new parameters of this PR? See comments for other parameters in this class.

IMO something like you described in the description of this PR is exactly what could be nice here, so the usage is clear without a need to look into the code.

mposolda (commented):
@eatikrh Thanks for the PR! It looks great. I've added one comment inline regarding some docs for new parameters.

Also it can be ideal to add some documentation with examples into this file https://github.com/keycloak/keycloak-benchmark/blob/22799bc65e39f8e353f2fb57a6938d051682b015/doc/dataset/modules/ROOT/pages/using-provider.adoc . Is it possible to add some new section like `Create many groups` with some examples for hierarchical and non-hierarchical groups?

dataset/src/main/java/org/keycloak/benchmark/dataset/config/DatasetConfig.java
eatikrh: Thank you for the review. Added code comments for the new parameters and added the "Create many groups" section in the usage documentation. Let me know it is too wordy.

ahus1 (approved):
Thank you for the changes. I updated the docs a little as I am a fan of [AsciiDoc description lists](https://docs.asciidoctor.org/asciidoc/latest/lists/description/).

PR #474, opened 2023-08-10
Utilise bash shell when calling aurora_delete.sh from reaper script

ryanemerson:


ahus1 (approved):
Thank you for spotting this! 

Out of curiosity: Would it also work to remote the "sh", and call it without? What would be the difference?

ryanemerson:
> Out of curiosity: Would it also work to remote the "sh", and call it without? What would be the difference?

Hmm yeah, that works as well! Not sure why I didn't do that originally :thinking:

Issue #467, opened 2023-08-07
Aurora database does not work with two OCP clusters in different region than Aurora


mhajas:
In the following setup:
1. OCP1 in eu-west1
2. OCP2 in eu-west3
3. Aurora in eu-west2

The setup doesn't work. The OCP with the peering connection created as the first is not working and the cluster is not able to acquire JDBC connections.

Steps to reproduce:
1. Create OCP1 OCP2 and Aurora database (Infinispan should not be important for reproducing)
2. Create a peering connection for OCP1 and Aurora
3. Create a peering connection for OCP2 and Aurora
4. Create KC deployment connecting to Aurora in both OCP1 and OCP2
5. OCP1 is not able to connect to Aurora and fails with the exception from the end of this report.

Additional info:
1. In the AWS console it seems everything is ok and peering connections are created for both VPCs (OCP1 and OCP2)
2. The same is observed if both OCPs are in the same region but Aurora is in a different one
3. In the situation above if we remove the peering connection for OCP1 and then create it again OCP1 starts working, but OCP2 stops working (database health check fails).

```
{
    "timestamp": "2023-08-07T14:18:42.568Z",
    "sequence": 8823,
    "loggerClassName": "org.jboss.logging.Logger",
    "loggerName": "org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler",
    "level": "ERROR",
    "message": "Error details:",
    "threadName": "main",
    "threadId": 1,
    "mdc": {},
    "ndc": "",
    "hostName": "keycloak-0",
    "processName": "QuarkusEntryPoint",
    "processId": 1,
    "exception": {
        "refId": 1,
        "exceptionType": "java.lang.RuntimeException",
        "message": "Failed to obtain JDBC connection",
        "frames": [
            {
                "class": "org.keycloak.quarkus.runtime.storage.database.jpa.AbstractJpaConnectionProviderFactory",
                "method": "getConnection",
                "line": 52
            },
            {
                "class": "org.keycloak.quarkus.runtime.storage.legacy.database.LegacyJpaConnectionProviderFactory",
                "method": "postInit",
                "line": 114
            },
            {
                "class": "org.keycloak.quarkus.runtime.integration.QuarkusKeycloakSessionFactory",
                "method": "init",
                "line": 105
            },
            {
                "class": "org.keycloak.quarkus.runtime.integration.jaxrs.QuarkusKeycloakApplication",
                "method": "createSessionFactory",
                "line": 45
[...]

PR #475, opened 2023-08-10
Aurora multi region

ryanemerson:
It's necessary for each ROSA cluster to use a non-overlapping
machine-cidr range when deploying across multiple AWS regions. If the
cidr range overlaps, then Aurora DB responses will not be correctly
routed to cluster than initiated a request.

Resolves https://github.com/keycloak/keycloak-benchmark/issues/467

@ahus1 @mhajas Utilising VPC peering connections to limit DB accessibility is the way to go when deploying production systems, but I'm starting to think it might be getting too restrictive for benchmarking with many clusters. If this continues be a pain point, I think we should consider documenting the steps required for said setup and then simply revert to using Aurora endpoints exposed over the internet.

### DIFF
--- a/.github/workflows/rosa-cluster-create.yml
+++ b/.github/workflows/rosa-cluster-create.yml
@@ -21,6 +21,11 @@ on:
         required: true
         default: '2'
         type: string
+      machineCidr:
+        description: 'Machine CIDR of the ROSA cluster'
+        required: true
+        default: '10.0.0.0/16'
+        type: string
 
 concurrency: cluster_${{ github.event.inputs.clusterName || format('gh-{0}', github.repository_owner) }}
 
@@ -54,6 +59,7 @@ jobs:
           COMPUTE_MACHINE_TYPE: ${{ inputs.computeMachineType }}
           MULTI_AZ: ${{ inputs.multiAz }}
           REPLICAS: ${{ inputs.replicas }}
+          MACHINE_CIDR: ${{ inputs.machineCidr }}
 
       - name: Archive ROSA logs
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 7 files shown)


mhajas (commented):
Thank you for investigating @ryanemerson! What is an example of non-overlapping cidr? Can we do, for example, 10.0.0.x where we increase x with each cluster we deploy? We can remember the last used x (for example in a aws secret?) and increase each time we deploy a new cluster. WDYT?

ryanemerson:
That could work :slightly_smiling_face:. We would probably need to use a list of taken cidrs to allow for clusters that have been kept alive etc.

I'm by no means a cidr/networking expert, but we could do `10.x.0.0/16`. That should allow each cluster to have access to all `10.x.y.z` addresses.

ahus1:
> I'm by no means a cidr/networking expert, but we could do `10.x.0.0/16`.

+1 for that, that sounds good for me.

I still wonder why this breaks "when deploying across multiple AWS regions": Even now all cluster have the same machine IP range, how does this work when everything is deployed to the same region?

BTW, the link above to the issues doesn't work for me. Is this a private filter you're referencing?

ryanemerson:
> I still wonder why this breaks "when deploying across multiple AWS regions": Even now all cluster have the same machine IP range, how does this work when everything is deployed to the same region?

I'm not sure tbh. I think it's because atm the ROSA clusters in the same region have the same machine cidr, so the traffic is somehow still able to route it's way back despite the use of different VPCs.


> BTW, the link above to the issues doesn't work for me. Is this a private filter you're referencing?

This was just a leftover from some automation I have that doesn't fit this project. Removed.

ryanemerson:
> Maybe when running within a region it is behaving like a bridge, and we are lucky when the IP addresses of a cluster (which are distributed randomly) aren't identically.
> 
> And across regions it is more routing, and those networks can't be routed?

Yes I think it's something like this.



> I'll merge this PR as is. I agree we need to automate this one or the other way once we see some pain (hopefully before it hurts too much).

I've automated it already :slightly_smiling_face: 

https://github.com/keycloak/keycloak-benchmark/pull/478

PR #477, opened 2023-08-10
Fix README that wrongly suggests to set AWS_REGION for workflows

### LINKED ISSUE #476: Fix README that wrongly suggests to set AWS_REGION for workflows
README suggests following: Create new variable `AWS_REGION` with value of your preferred AWS region, for example, `eu-central-1`. This produces an error https://github.com/martin-kanis/keycloak-benchmark/actions/runs/5820091558/job/15779748512
Looking into workflow files, it should reference to AWS_DEFAULT_REGION

martin-kanis:
Closes #476


mhajas (approved):
Thanks for the fix @martin-kanis

PR #478, opened 2023-08-10
Create unique machine CIDR per ROSA cluster

ryanemerson:
f


ahus1:
@ryanemerson 

cc: @mhajas - found some docs that indicate that some of the 10.0.x.x ranges are already used, so I reduced it to 10.0.x.x .. 10.63.x.x

ryanemerson:
Good find, thanks!

Issue #464, opened 2023-08-07
Dashboard for external Infinispan metrics


ahus1:
Provide metrics from the external Infinispan to display them in the Grafana dashboard. 

Ideally there should be no changes to the existing dashboard. If we want to add additional metrics, they would appear blank in the standard dashboard, or we copy the dashboard.

ahus1:
@pruivo: Martin and me looked at the setup via the Taskfile, and it already configures the metrics endpoints for Prometheus. 

When looking at the metrics, they are longer than in Keycloak, as Infinispan doesn't use the `names-as-tags`. Can we change that?

Thanks!

pruivo:
> @pruivo: Martin and me looked at the setup via the Taskfile, and it already configures the metrics endpoints for Prometheus.
> 
> When looking at the metrics, they are longer than in Keycloak, as Infinispan doesn't use the `names-as-tags`. Can we change that?
> 
> Thanks!

yes, I can add that.

pruivo:
@ahus1 PR here: https://github.com/keycloak/keycloak-benchmark/pull/465

rmdes:
Wondering how to add User Events to default keycloak metrics, anyone have any clue ?

mhajas:
@rmdes Please use another communication channel to ask your questions. Here are more details: https://www.keycloak.org/community This simplest option for you would be probably https://github.com/keycloak/keycloak/discussions. Also, please add more details on why you need it and what you want to achieve.

PR #465, opened 2023-08-07
Enable namesAsTags for Infinispan cluster

pruivo:
Depends on https://github.com/keycloak/keycloak-benchmark/pull/457

PR #488, opened 2023-08-14
Hardcode JGroups cluster name to match Keycloak embedded settings

ryanemerson:
This is a quick and easy way for us to make the external metric names align with the embedded infinispan names. A better solution would be to explicitly set the JGroups cluster name in the Keycloak EmbeddedCacheManager to a meaningful name associated with a specific Keycloak instance. I'll discuss this with the team and contribute a Keycloak PR as required.

Relates to: https://github.com/keycloak/keycloak-benchmark/issues/464

### DIFF
--- a/provision/infinispan/ispn-helm/templates/infinispan.yaml
+++ b/provision/infinispan/ispn-helm/templates/infinispan.yaml
@@ -38,7 +38,7 @@ spec:
     enabled: false
   replicas: {{ .Values.replicas }}
   container:
-    extraJvmOpts: '-Dorg.infinispan.openssl=false {{.Values.jvmOptions}}'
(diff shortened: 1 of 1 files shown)


ahus1:
@ryanemerson - I like this approach to align the names short term. 

Still, for the mid and long-term, I'd rather like to see "nameAsTags" put the name of the JGroups cluster in a tag.

ryanemerson:
@ahus1 We're already using `nameAsTags` on the external cluster https://github.com/keycloak/keycloak-benchmark/pull/465. The issue is that by default the Infinispan Operator sets the JGroups cluster name to be the same as the `metadata.name` field of the Infinispan CR, in this case "infinispan". Setting this java property overrides this configuration and sets the cluster name to the property value.

martin-kanis (approved):
@ryanemerson Thanks. It works well.

PR #490, opened 2023-08-15
Dashboard for external Infinispan metrics

### LINKED ISSUE #464: Dashboard for external Infinispan metrics

martin-kanis:
Closes #464

This PR adds ability to reuse Keycloak Infinispan Grafana board for external infinispan. To switch to the external infinispan, just select the correct namespace. 
Stats for local caches are empty for external Infinispan as there are no local caches.

### DIFF
--- a/provision/minikube/monitoring/dashboards/keycloak-infinispan.json
+++ b/provision/minikube/monitoring/dashboards/keycloak-infinispan.json
@@ -919,13 +919,10 @@
               {
                 "color": "green",
                 "value": null
-              },
-              {
-                "color": "red",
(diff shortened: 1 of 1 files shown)


mhajas:
Here is what the new metrics look like:
image

mhajas (approved):
Thanks @martin-kanis

PR #493, opened 2023-08-16
fix the missing namespace in widget filters for PostgreSQL DB widgets

kami619:
applied to both minikube and OpenShift deployments and they seem to work as excepted now, even when there are multiple Keycloak namespaces deployed to the same K8s cluster


mhajas (approved):
Thanks @kami619

PR #498, opened 2023-08-22
Documentation about Infinispan and cross-site

pruivo:


mhajas (approved):
Thank you for the docs, I like the changes. I added one comment that adds README files to each scripts/files location so it is simpler to find the correct documentation for each directory. Do you think it makes sense?

pruivo:
Sounds good @mhajas :+1:

PR #501, opened 2023-08-23
Update ROSA scalability workflow to commit the result numbers to github

andyuk1986:
Added Schedule run config for Rosa scalability benchmark

### DIFF
--- a/.github/actions/prometheus-metrics-calc/action.yml
+++ b/.github/actions/prometheus-metrics-calc/action.yml
@@ -25,9 +25,6 @@ inputs:
     default: '3'
   measurementInterval:
     description: 'A specific time period over which we want to calculate the required metrics.'
-  output:
-    description: 'The name of the output to store data in JSON format'
-    default: 'results.json'
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


.github/actions/prometheus-metrics-calc/action.yml
mhajas, line 1: I think we should include some sort of id for each result `json` file in case there are more runs on the same day. Similarly as it is done here: https://github.com/keycloak/keycloak-benchmark/blob/main/benchmark/src/main/content/bin/kcb.sh#L141

.github/workflows/rosa-scaling-benchmark-on-schedule.yml
```diff
@@ -0,0 +1,18 @@
...
+  schedule:
+    - cron: '0 7 * * 1-5' # Runs At 07:00 UTC on every day-of-week from Monday through Friday.
+
+jobs:
+
+  checkout:
+    name: ROSA Scheduled Scaling Benchmark Run
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run AuthorizationCode scalability scenario
```
mhajas, line 13: This is not running only AuthorizationCode

mhajas (commented):
Thank you @andyuk1986! I added a few comments. Otherwise, it looks good to me. 

As a follow-up, I think we should investigate the option of including these numbers in the dashboard within our docs. The dashboard is created here: https://github.com/keycloak/keycloak-benchmark/tree/main/dashbuilder

.github/actions/prometheus-metrics-calc/action.yml
```diff
@@ -87,11 +87,10 @@ runs:
...
-        if [ -f ${{ inputs.output }} ]; then
-          cat ${{ inputs.output }} | jq --arg testName "${{ inputs.performedTestName }}" --arg key "${description}" \
-          --arg val ${CALCULATED_METRIC_VALUE} '. + {($testName): {($key):($val|tonumber)}}' \
-          | tee ${{ inputs.output }}
-        else
-          jq -n --arg testName "${{ inputs.performedTestName }}" --arg key "${description}" \
-          --arg val ${CALCULATED_METRIC_VALUE} '{($testName): {($key):($val|tonumber)}}' > ${{ inputs.output }}
+        cur_date=$(date +"%Y/%m/%d")
+        if [ ! -f ${{ inputs.output }} ]; then
+          jq -n --arg current_date "${cur_date}" '{"Execution date": ($current_date)}' > ${{ inputs.output }}
```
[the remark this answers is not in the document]
mhajas: Looking at the [dashbuilder code](https://github.com/keycloak/keycloak-benchmark/blob/main/dashbuilder/static/dashboard.yaml#L9), it seems it is using `grafana_input.start.iso` for displaying date which maps to the following date format, therefore we should include this one:
```
DATE_END_ISO=$(date --iso-8601=seconds)
``` 

for displaying data. We should include that too.

.github/workflows/rosa-scaling-benchmark-on-schedule.yml
andyuk1986: Oops, result of copy/paste. Resolved.

.github/actions/prometheus-metrics-calc/action.yml
andyuk1986: Done. I have updated the action to add id and timestamp to the result json file. Also the name of the file is updated to include the uuid and timestamp in it (as it is done in keycloak scalability benchmark).

.github/actions/prometheus-metrics-calc/action.yml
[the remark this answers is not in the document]
andyuk1986: As agreed during status meeting, for dashbuilder I will create new task with new PR which will include the corresponding changes. For now, I keep it simple and just include the timestamp when the benchmark started to run.

andyuk1986:
@mhajas thanks for your review. I have updated the code according to the comments. Ready for review.

mhajas (approved):
Thank you @andyuk1986! LGTM. I am merging and I will execute the benchmark for today, to see how it works.

PR #506, opened 2023-08-24
Use synchronize instead of ansible fetch

ryanemerson:
This fixes `MemoryError` errors encountered when transferring large files from the remote EC2 instances.

https://stackoverflow.com/questions/41367278/how-to-fix-memory-error-in-ansible


kami619:
tested locally works as expected for a regular small load test. Thanks for adding this feature @ryanemerson

Issue #495, opened 2023-08-21
A/P failure testing in SYNC mode


PavelVlha:
Test scenarios:

**Manually switch from the active to the passive site under load**
- Document steps needed to do the switch
- Observe the recovery time needed (higher response times and maybe even failures expected as the passive site is not warmed up)

**Make the passive site fail**
- Kill the connection between active and passive sites to simulate a connection disruption. The idea is to remove the gossip router public route (we probably also need to kill/stop the Infinispan operator to make sure the operator does not recreate the route)
- Observe the active site - how long does it take for the active site to detect passive is down, does the active site work correctly without the passive

ryanemerson:
`kc-failover.sh` script, plus docs and associated changes required for failover tests added as part of https://github.com/keycloak/keycloak-benchmark/pull/507

# Deployment

```
AURORA_CLUSTER=ryan-aurora
AURORA_REGION=${REGION}
ROSA_CLUSTER_NAME_1=ryan-active
ROSA_CLUSTER_NAME_2=ryan-passive
DOMAIN="ryan-active-ryan-passive-mtyxmdck.keycloak-benchmark.com"
KC_CLIENT_URL="client.${DOMAIN}"
KC_HEALTH_URL_CLUSTER_1="primary.${DOMAIN}"
KC_HEALTH_URL_CLUSTER_2="backup.${DOMAIN}"
CROSS_DC_IMAGE=quay.io/pruivo/infinispan-server:14.0-patched

KC_DB_POOL_MIN_SIZE=15
KC_DB_POOL_MAX_SIZE=15
KC_DB_POOL_INITIAL_SIZE=15
KC_INSTANCES=6
KC_DISABLE_STICKY_SESSION='true'
KC_MEMORY_REQUESTS_MB=2048
KC_MEMORY_LIMITS_MB=3000
KC_HEAP_MAX_MB=2048
KC_HEAP_INIT_MB=1024
KC_CPU_REQUESTS=8
KC_CPU_LIMITS=8
KC_OTEL='true'
KC_CUSTOM_INFINISPAN_CONFIG_FILE=config/kcb-infinispan-cache-remote-store-config.xml
```

# Benchmark Setup

`./ansible/benchmark.sh` and `kc-failover.sh` script are executed concurrently. The latter is executed manually on one of the EC2 nodes used by `benchmark.sh` to ensure that the Route53 failover latency can be more accurately observed.

Environment:

```
cluster_size: 2
[...]

PR #507, opened 2023-08-29
A/P failure testing #495

ryanemerson:


kami619:
@ryanemerson there seems to be a failure in the Ascii docs tests

kami619 (approved):
overall the changes look good to me. We can work any fixes if need be in subsequent issues and tickets.

ansible/roles/benchmark/tasks/install.yml
```diff
@@ -14,6 +14,16 @@
 - name: Install jq
   package: name=jq state=present
 
+- name: Install dig
+  package: name=bind-utils state=present
+
+- name: Download & install OpenShift client
```
pruivo, line 20: I don't think you are using `oc` anywhere.

pruivo (approved):

ryanemerson:
Docs should be fixed. I just needed to escape `${DOMAIN}` with `$\{DOMAIN}`.

ansible/roles/benchmark/tasks/install.yml
ryanemerson: We need `oc` to login to the Openshift cluster before we execute the `kc-failover.sh` script.

mhajas (approved):
Thanks @ryanemerson

PR #509, opened 2023-08-30
Retry Aurora VPC delete in the event that dependencies are still bein…

ryanemerson:
Hopefully should fix failures similar to https://github.com/keycloak/keycloak-benchmark/actions/runs/6016418146/job/16320388452.

I inspected the AWS console this morning and the VPC had no dependencies, so I believe this is a timing issue.


ahus1:
I would be curious what the dependency would be here ... maybe it is that ROSA cluster which is delete in parallel?

ryanemerson:
I thought it was either the subnet or security-group associated with the Aurora VPC that are removed earlier in the script. However, it would be better if `auto_delete_on_schedule` initiated the `rosa-cluster-delete.yml` inline and waited for it's completion before proceeding to be sure.

ahus1:
Let's wait and see if this re-occurs, and only then try the next approach.

PR #511, opened 2023-08-30
Configure site name in Keycloak

pruivo:
* Added 'jboss.site.name' property
* Removed 'site' attribute from Infinispan configuration

Need to test it in Openshift but I'm expecting to see a log entry like: `[org.keycloak.connections.infinispan.DefaultInfinispanConnectionProviderFactory] (main) Node name: pedro-desktop-30198, Site name: my-site` with the proper site's name.

I saw usage of the site name in the events triggered by the work cache and some sessions/login failures cache.

### DIFF
--- a/provision/minikube/keycloak/config/kcb-infinispan-cache-remote-store-config.xml
+++ b/provision/minikube/keycloak/config/kcb-infinispan-cache-remote-store-config.xml
@@ -24,7 +24,7 @@
 
     <!-- the statistics="true" attribute is not part of the original KC config and was added by Keycloak Benchmark -->
     <cache-container name="keycloak" statistics="true">
-        <transport lock-timeout="60000" site="ISPN"/>
+        <transport lock-timeout="60000"/>
         <metrics names-as-tags="true" />
         <local-cache name="realms" simple-cache="true" statistics="true">
             <encoding>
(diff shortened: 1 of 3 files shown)


ahus1:
This looks like an important change. 

@pruivo - did you see something not working when testing this? 

@mhajas - would this fix some behavior which was broken earlier? How could we tell?

Adding @mhajas as a reviewer.

pruivo:
@ahus1 I got no errors neither in Infinispan or Keycloak logs. Everything seems to be working

mhajas:
> @mhajas - would this fix some behavior which was broken earlier? How could we tell?

@ahus1 What behavior do you have on your mind?

ahus1:
@mhajas - please take this more of an open ended question. I was for example wondering if a misconfigured site name could break some cross site functionality in Keycloak, like processing work cache events which imho contain a site name.

mhajas (approved):
Looking into the code, this should be configured and can avoid some issues. The following places look suspicious with the `ISPN` site name for both sites:

- https://github.com/keycloak/keycloak/blob/main/model/infinispan/src/main/java/org/keycloak/cluster/infinispan/InfinispanNotificationsManager.java#L291-L295
- https://github.com/keycloak/keycloak/blob/main/model/infinispan/src/main/java/org/keycloak/models/sessions/infinispan/events/AbstractUserSessionClusterListener.java#L62-L64

PR #513, opened 2023-08-31
Added basic get scenario

### LINKED ISSUE #512: Add a basic get endpoint scenario
### Description

To be able to easily run the benchmark against any basic GET endpoint it would be nice to have a scenario that can fetch a configurable endpoint. This can for example be used to test the well-known endpoint, or any other "GET" endpoints.

```
bin/kcb.sh --scenario=keycloak.scenario.basic.Get  --basic-url=http://localhost:8080/realms/master/.well-known/openid-configuration
```

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

stianst:
Closes #512


ahus1 (approved):
Hi @stianst - thank you for this PR. I've updated some docs to be aligned with other scenarios.

I also updated the file to be named `Get.scala` instead of `BasicGet.scala` so it is aligned with the other scenarios. As an alternative, it would be to rename the class from `Get` to `BasicGet` inside of the file. 

Please let me know if you agree with this change. This change is then ready to be squash-merged.

PR #537, opened 2023-09-11
Disable liveness and readiness probes to handle high load

ahus1:
With https://github.com/keycloak/keycloak/issues/22109 not being available, I wonder if we should disable those tests by default.


pruivo:
Don't the failover tests require probes to detect when the active site is down?

ahus1:
@pruivo - these probes are just for Kubernetes to skip the pod from the integrated loadbalancer (readiness probe failing) and when to restart a pod (liveness probe failing). 

We could still have probes (the URLs for liveness and readiness probes, or different ones) to make a Keycloak site up or down from the point of the shared external AWS loadbalancer which distributes the traffic between the primary and secondary site. 

Does this answer your questions?

PR #539, opened 2023-09-11
close http connections explicitly

### LINKED ISSUE #518: Enhance the current refresh token scenario to make the simulation closer to an production use case
### Description

enhance the current refresh token scenario to make the simulation closer to an production use case

- detailed yet to be fleshed out.

### Discussion

One approach that we are currently pursuing for this updated token refresh scenario is to.,

- Introduce a new cli argument to the kcb.sh script to `close` the http connection in the refresh token request to make sure we are not re-using the same http connection with the default `keep-alive`

### Motivation

_No response_

### Details

_No response_

kami619:
- Updated the Refresh Token scenario to optionally use `"Connection": "close"` for the refresh token endpoint
- Also adding a new way to look at the http requests sent using external proxy client with `--http-proxy` cli argument

fixes #518

### DIFF
--- a/doc/benchmark/modules/ROOT/pages/configuration.adoc
+++ b/doc/benchmark/modules/ROOT/pages/configuration.adoc
@@ -267,6 +267,11 @@ Used in xref:scenario/authorization-code.adoc[].
 | Number of token refreshes after logging in.
 Used in xref:scenario/authorization-code.adoc[].
 
+| [.nowrap]`--refresh-close-http-connection`
+| `true`
+| Close the HTTP connection after a token refresh.
+Used in xref:scenario/authorization-code.adoc[].
+
 | [[basic-url]][.nowrap]`--basic-url`
 | (not set)
 | URL to be called in the xref:scenario/basic-get.adoc[].
--- a/doc/benchmark/modules/ROOT/pages/configuration.adoc
+++ b/doc/benchmark/modules/ROOT/pages/configuration.adoc
@@ -104,6 +104,12 @@ Will log a lot of information.
 Use this only during development.
 Never use it during a load test as it slows down Gatling.
 
+| [.nowrap]`--http-proxy`
+| (not set)
+| If set, will proxy all HTTP requests to a proxy tool of your choice and the current default proxy host and port are `127.0.0.1`, `8888` respectively. These are currently not configurable.
+
+Never use it during a load test as it slows down Gatling to be able to proxy the requests to a proxy client and comes with the usual proxy nuances that we have to be careful, in general.
+
 | [.nowrap]`--sla-error-percentage`
 | `0`
 | Maximum percentage of requests to fail during a run.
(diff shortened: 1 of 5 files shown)


doc/benchmark/modules/ROOT/pages/configuration.adoc
```diff
@@ -104,6 +104,12 @@ Will log a lot of information.
 Use this only during development.
 Never use it during a load test as it slows down Gatling.
 
+| [.nowrap]`--http-proxy`
+| (not set)
+| If set, will proxy all HTTP requests to a proxy tool of your choice and the current default proxy host and port are `127.0.0.1`, `8888` respectively. These are currently not configurable.
+
+Never use it during a load test as it slows down Gatling to be able to proxy the requests to a proxy client and comes with the usual proxy nuances that we have to be careful, in general.
+
 | [.nowrap]`--sla-error-percentage`
```
pruivo, line 113: isn't `--close-http-connection` missing from the docs?

doc/benchmark/modules/ROOT/pages/scenario/authorization-code.adoc
```diff
@@ -49,7 +49,7 @@ bin/kcb.sh \
 
 To create offline sessions, set the parameter xref:configuration.adoc#scope[`--scope`] to a value including `offline_access`, for example, `openid profile offline_access`.
 
-To test repeated refreshing of tokens between authenticating and logging out, pass `--refresh-token-count=<count>` and `--refresh-token-period=<seconds>`.
+To test repeated refreshing of tokens between authenticating and logging out, pass `--refresh-token-count=<count>` and `--refresh-token-period=<seconds>`. And pass in the additional optional argument `--close-http-connect=true` to test with a new http connection for the refresh token endpoint every time, we also should see the Request name in the metadata to be `RefreshTokenWithNewHTTPConnection` instead of `RefreshToken`, when `--close-http-connect=true` is in use.
```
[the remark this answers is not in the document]
kami619: good catch, fixed the typo.

doc/benchmark/modules/ROOT/pages/configuration.adoc
kami619: I left it out given `--close-http-connection` is only implemented for the refresh token endpoint, similar to the other cli arguments `--refresh-token-period` or `--refresh-token-count` which were relevant only to a specific scenario, at least for now. But if you think, it would add value to expose it in the global config doc, I can add it here. lmk. @pruivo

pruivo: looks good to me :+1: thanks @kami619

pruivo (approved):

ahus1:
@kami619 - I've reviewed the change and update the option: I changed in the last commit from `close-http-connection` to `refresh-close-http-connection`. 

I assume in the future even if we want to add it to other steps as well, we would make it specific to the step. So that's why I prefixed `refresh` to the option to make it refresh specific, and to have room for additional independent options once we add them.

Please review my latest change and let me know if you agree. Thanks!

kami619:
lgtm, I like the introduction of  default Boolean value to close http connection for Refresh Token endpoint. Thanks for the review and edits @ahus1

PR #540, opened 2023-09-12
Make Infinispan failure detection timeout configurable

### LINKED ISSUE #532: Make Infinispan failure detection timeout configurable
When running in SYNC mode, and the Gossip Router between two sites fails, there is currently a timeout of ~40 seconds. 

To make it workable for customers, it would be good to make this configurable to get in the area of 10 to 15 seconds.

pruivo:
Defaults: 10 sec with 2 sec heartbeat interval (suspected after 5 heartbeats lost)

Fixes #532


ahus1 (approved):
Thank you for this enhancement!

PR #552, opened 2023-09-21
Added more task to rosa-cross-dc Taskfile.yaml

pruivo:
* route53 -> creates the Route53 DNS records
* monitoring -> installs grafana in both ROSA clusters
* dataset-import -> triggers the user creation in the primary cluster

### DIFF
--- a/provision/rosa-cross-dc/Taskfile.yaml
+++ b/provision/rosa-cross-dc/Taskfile.yaml
@@ -8,9 +8,10 @@ vars:
   # To compute Infinispan namespace
   KC_NAMESPACE_PREFIX: '{{default "$(whoami)-" .KC_NAMESPACE_PREFIX}}'
   KC_ISPN_NAMESPACE: '{{.KC_NAMESPACE_PREFIX}}keycloak'
-  ISPN_DIR: "../infinispan"
-  RDS_DIR: "../aws/rds"
-  KC_DIR: "../openshift"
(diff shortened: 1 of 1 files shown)


provision/rosa-cross-dc/Taskfile.yaml
```diff
@@ -308,3 +307,185 @@ tasks:
...
+    desc: "Install grafana dashboards in both ROSA clusters"
+    deps:
+      - common:split
+      - common:env
+    requires:
+      vars:
+        - ROSA_CLUSTER_NAME_1
+        - ROSA_CLUSTER_NAME_2
+    cmds:
+      - task: install-helms
```
ahus1, line 419: AFAIK the `helm` commands also target a specific cluster, so running this without a parameter would target only the currently logged in cluster.

ahus1 (changes requested):
Thank you for this addition. I didn't test it, I just reviewed the code. I found one small thing below, please have a look.

provision/rosa-cross-dc/Taskfile.yaml
pruivo: Those are local helm commands to add the Grafana and Jaeger repositories. I changed the task name to make it more clear.

ahus1: I didn't know that those command were local only, so I learned something new today!

ahus1 (approved):

PR #555, opened 2023-09-26
Add a paragraph about building custom image for OpenShift into kubern…

### LINKED ISSUE #554: Add a paragraph about building custom image for OpenShift into kubernetes guide
There is a chapter [Using a custom Keycloak image for deployment in Kubernetes](https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/util/custom-image-for-keycloak). 

It describes building a custom image for minikube. It should be added a paragraph about building custom image for OpenShift as well.

vramik:
…etes guide

Closes #554


ahus1 (approved):
Thank you for the added docs. The Podman examples worked as expected. When running the Docker example, I found that the `docker push` command accepts only one parameter, where there `podman push` command would accept one or two parameters.

I aligned the documentation so both invocations now use the syntax with only one additional parameter. 

Thank you again for finding and fixing this documentation gap.

PR #556, opened 2023-09-26
Increase max-replicas on rosa_create_cluster.sh to avoid error during…

vramik:
… creation of cluster

There is following [error](https://github.com/keycloak/keycloak-benchmark/actions/runs/6308269787/job/17126198580) during creation of cluster. This PR increases --max-replicas value in rosa_create_cluster.sh to avoid the `ERR: Failed to add machine pool to cluster 'gh-keycloak': Attribute 'autoscaling.max_replicas' must be a integer greater than 0.`


mhajas (approved):
Thanks for the PR @vramik. I tried it locally and it seems something changed in ROSA and it is no longer possible to create machinepools with max-replicas set to 0. It makes sense to me to increase it to 10 and rely on the autoscaling feature to scale nodes to 0 when no needed.

PR #569, opened 2023-10-12
Docs: deploying Infinispan with Operator

pruivo:


ahus1 (changes requested):
Hi @pruivo - thank you for these two great step-by-step guides. I moved the includes to a separate folder, this way they are not converted to separated pages by Antora. 

I did some re-rewording in some places to have shorter sentences and to use the active voice where it was previously the passive voice. Please review my changes to see if I slipped in some mistakes doing that. 

Two questions remand open for me: 

* I didn't see the Gossip Router mentioned in here, and that we deploy it only in one datacenter. Deploying only one is AFAIK still the recommendation, so I would expect this to be mentioned in the docs.
* I found that the work cache is in Keycloak a replicated cache, while it seems to be a distributed cache in the external Infinispan. Could you please comment on this / discuss with this the team? We can also handle this in a separate issue if it requires additional investigation / alignment.

ahus1:
@pruivo - one more thing: Please document on how to update the memory settings (requested + limit on Kuberentes level, JVM parameters for heap). A first iteration would IMHO be: Use the calculation we have to calculate the memory for the 3-node Keycloak deployment, and then use the same setting for Infinispan, assuming that the RAM usage for sessions would be similar.

doc/kubernetes/modules/ROOT/pages/running/infinispan-crossdc-deployment.adoc
```diff
@@ -0,0 +1,183 @@
...
+<10> The Truststore key (filename) of the Keystore as defined in the previous step.
+<11> The remote site's name, in this case  `{site-b}`.
+<12> The namespace of the {ispn} cluster from the remote site.
+<13> The {ocp} API URL for the remote site.
+<14> The secret with the access toke to authenticate into the remote site.
++
+For `{site-b}`, the {ispn} CR looks similar to the above.
+Note the differences in point 4, 11 and 13.
++
+.The {ispn} CR for `{site-a}`
```
ryanemerson, line 149: `{site-b}`?

doc/kubernetes/modules/ROOT/pages/running/infinispan-crossdc-deployment.adoc
```diff
@@ -0,0 +1,183 @@
...
+----
+
+. Creating the caches for {kc}.
++
+{kc} requires the following caches to be present: `sessions`, `actionTokens`, `authenticationSessions`, `offlineSessions`, `clientSessions`, `offlineClientSessions`, `loginFailures`, and `work`.
++
+The {ispn} {operator-docs}#creating-caches[Cache CR] allows to deploy the caches in the {ispn} cluster.
+Cross-site needs to be enabled per cache as documented by {xsite-docs}[Cross Site Documentation].
+The documentation contains more details about the options used by this guide.
+In the following example shows the Cache CR for `{site-a}`.
```
ryanemerson, line 162: "The following..."

doc/kubernetes/modules/ROOT/pages/running/infinispan-deployment.adoc
```diff
@@ -0,0 +1,52 @@
...
+== Procedure
+
+include::partial$/running/infinispan-install-operator.adoc[]
+include::partial$/running/infinispan-credentials.adoc[]
+
+. Create an Infinispan cluster.
++
+The https://infinispan.org/docs/infinispan-operator/main/operator.html#creating-clusters[Creating Infinispan clusters] documentation provides all the information on how to create and configure your Infinispan cluster.
++
+For simplicity, we provide the following example as a starting point.
```
ryanemerson, line 27: This sentence doesn't really add anything, I think we can remove it.

doc/kubernetes/modules/ROOT/pages/running/keycloak-with-external-infinispan.adoc
```diff
@@ -15,12 +15,12 @@ See xref:running/index.adoc[] for additional guides.
 
 * OpenShift or Kubernetes cluster running.
 * Existing xref:running/keycloak-deployment.adoc[Basic Keycloak deployment] as it will be extended.
-* Existing deployment of Infinispan with the Operator (TODO).
+* Existing Infinispan deployment, for example, one of xref:running/infinispan-deployment.adoc[] or xref:running/keycloak-with-external-infinispan.adoc[].
```
ryanemerson, line 18: Should "infinispan" be "{ispn}"?

doc/kubernetes/modules/ROOT/partials/running/infinispan-credentials.adoc
```diff
@@ -0,0 +1,37 @@
+. Configure the credential to access the Infinispan cluster.
++
+Keycloak needs this credential to be able to access the cluster.
```
ryanemerson, line 3: "Keycloak needs this credential to be able to authenticate with the {ispn} cluster."?

doc/kubernetes/modules/ROOT/partials/running/infinispan-credentials.adoc
```diff
@@ -0,0 +1,37 @@
...
+By default, the Infinispan Operator generates credentials with a random password and stores it in a secret.
```
ryanemerson, line 4: I think we should just document how to define a pre-configured password and then provide a link to the Infinispan Operator docs as further reading at the end of this section. I don't think a Keycloak user cares about Infinispan's defaults etc, just what they need to do to get started and using pre-configured credentials is the easiest way.

pruivo:
@ahus1 thanks for that! Looks good, just squashed the commits.

I didn't mention the Gossip Router because I'm trying to keep the CRs as minimal as possible and, it is most likely related to how the Openshift is configured. It has been the default to have 1 Gossip Router per site and I haven't heard any reports from customers. In the worst case, we can write some knowledge article or some tunning guide at the end.

The Infinispan memory configuration is documented in our guides, which are linked. If you have some idea of how much memory is required per session, we can try to give some guidance. 

The `work` cache is replicated in Keycloak because it needs to trigger events when it changes. When there is an external Infinispan deployment, the events originate from the external Infinispan. It does not matter (except to prevent data loss) if the cache is distributed or replicated, externally.

doc/kubernetes/modules/ROOT/pages/running/infinispan-crossdc-deployment.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+. Create TLS secrets
++
+In this guide Infinispan uses an {ocp} Route for the cross-site communication.
+It uses the SNI extension of TLS to direct the traffic to the correct Pods.
+To achieve that, JGroups use TLS sockets, which require a Keystore and Truststore with the correct certificates.
++
+For more information check {operator-docs}#securing-cross-site-connections_cross-site[Securing Cross Site Connections] documentation or this https://developers.redhat.com/learn/openshift/cross-site-and-cross-applications-red-hat-openshift-and-red-hat-data-grid[Red Hat Developer Guide].
++
+Upload the Keystore and the Truststore in an {ocp} Secret.
```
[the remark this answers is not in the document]
pruivo: What do you mean? It is not a header.

doc/kubernetes/modules/ROOT/pages/running/keycloak-with-external-infinispan.adoc
pruivo: `ispn` attribute does not exist in this file.

doc/kubernetes/modules/ROOT/pages/running/infinispan-crossdc-deployment.adoc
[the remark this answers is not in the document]
ryanemerson: The statement was so concise, I wasn't sure if it was meant to be a heading for the statements that follow.

Maybe we can expand it slightly so the purpose of this action is clear:

"The Keystore and Truststore must be deployed to {ocp} Secrets so that they can be utilised by the {ispn} deployment."

Then maybe remove the rest of the content before ".Deploy a Keystore".

ahus1 (approved):
Thanks you for the updates. I was concerned with the Gossip Router as I think we did the failure testing with only one Gossip Router and not two. Would it make sense to repeat the failure testing with two Gossip Routers? If yes, please create a follow-up issue.

PR #574, opened 2023-10-24
Docs: Added diagrams for cross-site deployment

pruivo:
Added the diagrams and changed the namespace to `keycloak` to match the documented options in "keycloak with external infinispan" page.


kami619:
@pruivo The diagrams mention that the gossip router pods are present in both active and passive clusters, but currently they are only deployed to the active cluster, is that not correct ?

mhajas:
@ryanemerson Could you please have a look at the database setup there? We were talking at the meeting about how to make the difference between multi-az and multi-region aurora setups visible. Will we have writer and reader instances in multi-az aurora setup too?

pruivo:
> @pruivo The diagrams mention that the gossip router pods are present in both active and passive clusters, but currently they are only deployed to the active cluster, is that not correct ?

By default, and the examples in the docs use the default, it creates a gossip router per site.

ryanemerson:
> Will we have writer and reader instances in multi-az aurora setup too?

Yes, there's only ever a single writer instance. All other Aurora instances are reader instances. During failover, AWS handles the election of a new writer. 

> Could you please have a look at the database setup there? We were talking at the meeting about how to make the difference between multi-az and multi-region aurora setups visible.

For Multi-az I update the diagram to include the "writer" instance in `eu-west-1a` and a "reader" instance in `eu-west-1b` instead of just "Aurora DB". 

For Multi-region, I would make it so that the `eu-west-1` explicitly refers to an "Aurora Primary DB cluster "and `eu-east-1` "Aurora Secondary DB cluster". It's also important that the directional arrow between these two clusters only ever goes Primary -> Secondary.

pruivo:
@ryanemerson updated the diagram. Let me know how it looks now :+1:

kami619:
I think it looks good now and matches @ryanemerson's earlier feedback. Merging it.

PR #575, opened 2023-10-30
updating the jgroups docs for kc22

### LINKED ISSUE #553: Document JGroup Threads for KC22
As a follow-up to #520, find a place where to document: 

* the reduction of Keycloak worker threads to be aligned with the JGroup thread size, as running into retries due to missing threads would still slow down the system
* setting `-Djgroups.thread_dumps_threshold=1` to get notified in the logs when the thread pool is exhausted.
* monitoring the metric `vendor_jgroups_tcp_get_thread_pool_size`

kami619:
Fixes #553 

@ahus1 looks like the first two points in the ticket are already [updated](https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/running/concepts/threads#jgroup-connection-pool) and let me know if they need to be elaborated any further. 

> - the reduction of Keycloak worker threads to be aligned with the JGroup thread size, as running into retries due to missing threads would still slow down the system
> - setting -Djgroups.thread_dumps_threshold=1 to get notified in the logs when the thread pool is exhausted.

I have added a new paragraph for this below item.
> monitoring the metric vendor_jgroups_tcp_get_thread_pool_size

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/running/concepts/threads.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/running/concepts/threads.adoc
@@ -35,6 +35,8 @@ To see the error the first time it happens, the system property `jgroups.thread_
 include::partial$executor-jgroups-thread-calculation.adoc[]
 --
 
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
Thank you @kami619 for these change. I did a small rewording here.

PR #576, opened 2023-10-31
Ensure aurora global subnet groups are correctly removed

ryanemerson:
@ahus1 This should fix the reaper failures


ahus1 (approved):
Thank you for the change. While it looks a bit ugly, still I think I understand why this works. Not sure how to improve this.

PR #581, opened 2023-11-01
Customer Guide for Aurora Multi AZ. Resolves #580

### LINKED ISSUE #580: Customer Guide for Aurora Multi AZ
For customers to prepare a Cross-DC multi-AZ setup, one building block is to prepare an Aurora database and connect it to the two OpenShift clusters. 

For the KCB project, this has been setup as follows:

* There is a guide describing which scripts to call: https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/storage/aurora-regional-postgres
* As the customers won't use the scripts, but instead want to understand the setup, it should look similar to https://www.keycloak.org/keycloak-benchmark/kubernetes-guide/latest/running/infinispan-crossdc-deployment with the Architecture / Preprequisites / Procedure / Verification steps. For the procedure, consider referencing parts of the existing `autora_create.sh` script
* As a pre-requisite, list that each ROSA cluster should have its own machine CIDR so they don't overlap and routing works as expected. AFAIK @ryanemerson experienced connectivity problems otherwise.

ryanemerson:
@ahus1 @vramik @martin-kanis Let me know what you think. I've focused mainly on content, so I'm sure there's many ways we can make this prettier ... my adoc skills are limited, so all tips welcome!

I have split creating an Aurora DB and the peering connections into two distinct guides, as the latter is applicable to other Aurora deployment types and the process must be repeated for every ROSA cluster that requires a DB connection.

Closes #580


ahus1:
Thank you for this documentation, this is very detailed. Some thoughts: 

* There are now two guides, one for the DB and one for the peering. If we expect users to perform always both steps, so IMHO those should be combined into one guide. If we want to have different building blocks later for example for a Aurora global database which would involve different steps, then we should keep it as is.
* The guide is very hands-on and descriptive, as it captures the outputs. An alternative would be be a less descriptive include of parts of our shell script for the Aurora setup, which would give us less duplication. As the Aurora setup is quite stable, I suppose we continue as you outlined it here - I just want to make sure we have discussed it. 
* At the end of setting up the peering connection, there is the - all to obvious step - to add the JDBC URL plus credentials into the Keycloak CR to connect to the database. Even if it is obvious to us, it would be IMHO good to mention where to place it.

ryanemerson:
> * There are now two guides, one for the DB and one for the peering. If we expect users to perform always both steps, so IMHO those should be combined into one guide. If we want to have different building blocks later for example for a Aurora global database which would involve different steps, then we should keep it as is.

Ack. I'll try to re-work the peering connection to be a reusable procedural block that can be nested in multiple guides.

> * The guide is very hands-on and descriptive, as it captures the outputs. An alternative would be be a less descriptive include of parts of our shell script for the Aurora setup, which would give us less duplication. As the Aurora setup is quite stable, I suppose we continue as you outlined it here - I just want to make sure we have discussed it.

I don't think including parts of our Aurora setup scripts would be very understandable for users, as we heavily modify the outputs of many of the commands in order to capture values in variables etc and use 3rd party tools such as `jq`.


> * At the end of setting up the peering connection, there is the - all to obvious step - to add the JDBC URL plus credentials into the Keycloak CR to connect to the database. Even if it is obvious to us, it would be IMHO good to mention where to place it.

Do you think stating the need for this in the "Next Steps" section and linking the relevant guide is sufficient, or do you envisage us explicitly documenting these steps in the same guide? IMO the latter is overkill if we want the documentation to be more modular (maybe that's not the case?).

ryanemerson:
I have added a new commit which combines the two guides using partials, so we can re-use the peering connection procedure in future guides if required.

I have also replaced the "Next Steps" section with a "Keycloak Deployment" section. This isn't exhaustive instructions, but it explains the changes required to the Keycloak CR when deploying with Aurora vs Postgres.

ahus1:
> Do you think stating the need for this in the "Next Steps" section and linking the relevant guide is sufficient

I was thinking to have 1-3 sentences in the form of: "Note the username and password and a JDBC URL to deploy Keycloak as described in the XXX guide". The thing that triggered it for me was me seeing the URL used for the PostgreSQL client is different than the JDBC URL used for Keycloak.

ahus1:
My comment overlapped with your latest commit. That latest change works well - thank you.

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
```diff
@@ -0,0 +1,214 @@
...
+  --filters "Name=tag:AuroraCluster,Values=keycloak-aurora" \
+  --query 'Vpcs[*].VpcId' \
+  --region eu-west-1 \
+  --output text
+----
++
+.Output:
+[source,json]
+----
+vpc-0b40bd7c59dbe4277
```
martin-kanis, line 16: This output is in red colour which seems little bit weird to me. Not sure why it is red, though.

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
```diff
@@ -0,0 +1,214 @@
...
+vpc-0b721449398429559
+----
++
+. Create Peering Connection
++
+.Command:
+[source,bash]
+----
+aws ec2 create-vpc-peering-connection \
+  --vpc-id vpc-0b721449398429559 \# <1>
```
martin-kanis, line 46: These clues are explained below and they are numbered, which creates a nested list which again seems little bit off. If someone knows, how to do it better, that would be great.

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
```diff
@@ -0,0 +1,214 @@
...
+  --peer-vpc-id vpc-0b40bd7c59dbe4277 \# <2>
+  --peer-region eu-west-1 \
+  --region eu-west-1
+----
+<1> ROSA cluster VPC
+<2> Aurora VPC
++
+.Output:
```
martin-kanis, line 54: This and other outputs are very verbose. It seems we need only part of the output for the later commands. Could we perhaps shorten them wherever it is possible?

doc/kubernetes/modules/ROOT/pages/running/aurora-multi-az.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+== Verifying the connection
+
+include::partial$aurora/aurora-verify-peering-connections.adoc[]
+
+== Deploying Keycloak
+
+Now that an Aurora DB has been established and linked with all of your ROSA clusters, the next step is to
+xref:running/keycloak-deployment.adoc[Deploy Keycloak] with the JDBC url configured to use the Aurora DB writer
+endpoint. Todo this we create a `Keycloak` CR outlined in the xref:running/keycloak-deployment.adoc[] guide, however
```
martin-kanis, line 51:
```suggestion
endpoint. To do this we create a `Keycloak` CR outlined in the xref:running/keycloak-deployment.adoc[] guide, however
```

martin-kanis (changes requested):
Thanks for the PR @ryanemerson!

Content-wise it looks good. I have some notes (see inline) the formatting. Maybe someone with good asciidoc expertise could advise us how to address those.


Then, I was trying to build this PR to see the docs locally but I was not able to see it. I did following (maybe I missed some steps):

1.  `mvn clean install`
2. `./build.sh` from antora dir

Then output is in `keycloak-benchmark/antora/_site/keycloak-benchmark/`. However, I don't see anything related to Aurora in `kubernetes-guide/latest/running`. So this PR is probably missing something. One thing I noticed, there is no reference in `doc/kubernetes/modules/ROOT/nav.adoc`

ahus1:
@martin-kanis - for a local preview of the docs, use the script author.sh for an authors preview. The build script will clone the main repo main branch to render the contents (possibly from multiple branches).

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
martin-kanis: Disregard this comment. In the compiled version it looks good!

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
martin-kanis: Disregard this comment. In the compiled version it looks good!

martin-kanis:
> @martin-kanis - for a local preview of the docs, use the script author.sh for an authors preview. The build script will clone the main repo main branch to render the contents (possibly from multiple branches).

Thank you for the advice. @ryanemerson I can know see rendered pages. Only thing missing, is the reference in `doc/kubernetes/modules/ROOT/nav.adoc` to see links in the left navigation menu.

doc/kubernetes/modules/ROOT/partials/aurora/aurora-create-peering-connections.adoc
ryanemerson: That's true. I was torn between providing detailed output for completeness and providing succinct docs. One way we could improve this would be to add `--query` args to some of the commands where only a single value is required, that way the output will be much less verbose.

Issue #534, opened 2023-09-07
Investigate health probe for loadbalancer


ahus1:
When there are two sites, each site can be in multiple states: 

* Down -> no traffic should go to it 
  * "Keycloak is down" (working already)
  * "Network between LB and Keycloak is down" (working already)
  * "External Infinispan is down" (no probe in Keycloak for that currently)
  * "Communication between Keycloak and Infispan is down" (no probe in Keycloak for that currently)

* Up with out-of-date data -> no traffic is allowed go to the site - in A/P this happens only to the passive site
  * SYNC
    * Timeout Exception: Data is either lost or arrives too late, no retry -> Infinispan can be configured that the exception is propagated to Keycloak
    * Suspect exception: assuming data is lost (node to be written to is removed from the JGroup view) as cross-DC doesn't have a retry mechanism in SYNC mode  -> Infinispan can be configured that the exception is propagated to Keycloak
    * Offline triggered by the number of failures and missing positive acks in an amount of time -> polling of metric or a REST endpoint possible
  * ASYNC
    * Secondary site is always out of date
    * All errors will cause a retry
    * Offline triggered by the number of failures and missing positive acks in an amount of time -> polling of metric or a REST endpoint possible
    * Passive site will eventually be up-to-date once the retries succeed and an offline site will receive a state transfer

* Up with up-to-date data -> traffic is allowed to go to the site, but it depends if this is site is considered the active site in an active/passive setup

It might get tricky once there is 

* A failover from a primary to a secondary, and the primary comes up again, doesn't have the data and must not be used
* The primary considering the secondary offline, and then the secondary becoming out-of-sync, and then the secondary must not be used even if the primary becomes offline.

The different strategies might depend on SYNC vs. ASYNC setup. To simplify things, only A/P setups can be considered here.

Out-of-date data means: 
* Keycloak local caches are filled with with old data, for example a user can log-in with an old password, but not their new password
* A user can still use their token in the out-of-date site, while they have been logged-out in the other site.

ahus1:
WIP idea: 

* Have a probe in Keycloak (REST endpoint) that returns the information if the other site is up-to-date
* Have a cloud provider specific pod implementing STONITH: If the other site is not considered up-to-date AND the current site is marked available in the loadbalancer, remove the other site from the loadbalancer so it can't become and active out-of-date site (and using some locking preventing the other site doing this concurrently)

ahus1:
[...]

PR #586, opened 2023-11-13
AWS Global Accelerator Load Balancing

ryanemerson:


ahus1:
@ryanemerson - thank you for this PR. Great to see this working. 

I have been wondering if  `/health/live` is the right endpoint, and I think it is the right one as it is the only one that will be available even under high load as it is non-blocking. I think it is worth highlighting this in the docs. 

For the probes, it seems they are using HTTP and port 80 at least by looking at the docs and scripts - that surprises me. Could you please double check? It would be good to know where TLS is terminated and possibly re-encrypted in that setup. There shouldn't be unencrypted sections in the setup.

ryanemerson:
> I have been wondering if `/health/live` is the right endpoint, and I think it is the right one as it is the only one that will be available even under high load as it is non-blocking. I think it is worth highlighting this in the docs.

+1 I think it's the best option for now. We created https://github.com/keycloak/keycloak-benchmark/issues/534 to investigate possible improvements and I think this would be a good improvement to focus on after the initial release.


> For the probes, it seems they are using HTTP and port 80 at least by looking at the docs and scripts - that surprises me. Could you please double check? It would be good to know where TLS is terminated and possibly re-encrypted in that setup. There shouldn't be unencrypted sections in the setup.

Great spot @ahus1! The global accelerator actually ignores the health probe settings on the Endpoint group when using ALB/NLB endpoints and delegates to the LB's health checks. In this case we were using the default health check settings which simply opens a TCP connection to the k8s service port. I have updated the docs and scripts so that when we create the k8s Service we explicitly configure the health checks on the NLB.

ryanemerson:
When executing the failover benchmark with the Global Accelerator I'm noticing that we continue to encounter exceptions after the failover to the backup cluster has completed. I need to investigate the root cause before we merge this PR.

Screenshot from 2023-11-14 17-15-59

ryanemerson:
Closing in favour of Route53 based approach: https://github.com/keycloak/keycloak-benchmark/pull/594

PR #590, opened 2023-11-15
Infinispan operational guide

pruivo:
I changed existing docs to SYNC cross-site replication and added the manual steps for Infinispan to recovery from some scenarios.

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
@@ -69,15 +69,17 @@ include::partial$running/infinispan-running-batch.adoc[]
 
 ==== Procedures
 
+WARNING: Transferring the full state may impact the {ispn} cluster perform by increasing the response time and/or resources usage.
+
 First procedure is to delete the stale data from the secondary site.
 
-WARNING: Ensure you are connected to the secondary site before executing the clear operation.
+. Login into your secondary site.
 
-. Login into your secondary site
+. Shutdown Keycloak.
 
 . Run the `crossdc-clear-caches` Batch CR (see xref:#running-infinispan-batch[]).
 
-. Delete all Batch CRs created, since they are not required anymore.
+. Delete all Batch CRs created.
 
 Now we are ready to transfer the state from the primary site to the secondary site.
 
@@ -87,21 +89,23 @@ Now we are ready to transfer the state from the primary site to the secondary si
 
 . Check the replication status is `ONLINE` for all caches with `crossdc-status` Batch CR.
 
-. Wait for the state transfer to completed by checking the output of the Batch CR `crossdc-push-state-status`
+. Wait for the state transfer to complete by checking the output of the Batch CR `crossdc-push-state-status`
 
 . If an error is reported in step 4, try to repeat the operations from step 2.
 
-. Finally, clear the state transfer status with Batch CR `crossdc-reset-push-state-status` and delete all Batch CRs created.
+. Clear the state transfer status with Batch CR `crossdc-reset-push-state-status`.
 
-WARNING: Transferring the full state may impact the {ispn} cluster perform by increasing the response time and/or resources usage.
+. Delete all Batch CRs created.
+
+As now the state is available in the secondary datacenter, Keycloak can be started again:
+
+. Login into your secondary site.
+
+. Startup Keycloak.
 
 === AWS Aurora Database
 
-Section under development.
-
-=== Keycloak Cluster
-
-Section under development.
+No action required.
 
 === Route53
 
--- a/doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
@@ -1,0 +1,108 @@
+= Out-of-sync datacenters Recovery
+:description: This guide describes the steps to bring the secondary site up to date to the primary site after a network partition.
+
+include::partial$running/infinispan-attributes.adoc[]
+
+{description}
+
+// Infinispan -> clear caches on secondary site and perform the state transfer
+// Keycloak -> clear caches (how?)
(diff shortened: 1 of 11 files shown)


doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+==== Procedures
+
+WARNING: Transferring the full state may impact the {ispn} cluster perform by increasing the response time and/or resources usage.
+
+First procedure is to delete the stale data from the secondary site.
+
+. Login into your secondary site.
+
+. Shutdown Keycloak.
```
ryanemerson, line 78: How should the user do this?

doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
```diff
@@ -0,0 +1,112 @@
...
+=== {ispn} Cluster
+
+For the context of this guide, `{site-a}` is the primary site and `{site-b}` is the secondary site.
+
+Network partitions may happen between the site and the replication between the {ispn} cluster will stop.
+Manual steps described in this guide bring both site back in sync.
+
+include::partial$running/infinispan-batchcr-intro.adoc[]
+
+The `ConfigMap` with the {ispn} CLI operations are as follows:
```
ryanemerson, line 28: FYI Batch CRs can be created without a ConfigMap, e.g.:
```yaml
apiVersion: infinispan.org/v2alpha1
kind: Batch
metadata:
  name: mybatch
spec:
  cluster: infinispan
  config: |
    create cache --template=org.infinispan.DIST_SYNC mycache
    put --cache=mycache hello world
    put --cache=mycache hola mundo
```
IMO this is simpler for the user, but maybe there was a reason you always created a dedicated ConfigMap?

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+. If an error is reported by step 4, try to repeat the operations from step 2.
+
+. Clear the state transfer status with Batch CR `crossdc-reset-push-state-status`
+
+. Delete all Batch CRs created.
+
+. Login to the primary site.
+
+. Start Keycloak.
```
ryanemerson, line 105: How?

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+. Login to the primary site.
+
+. Shutdown Keycloak. This will clear all Keycloak caches, and it prevents the state of Keycloak from being out-of-sync with Infinispan.
+
+. Run the `crossdc-clear-caches` Batch CR (see xref:#running-infinispan-batch[]).
+
+. Delete all Batch CRs created.
+
+Now we are ready to transfer the state from the secondary site to the secondary site.
```
ryanemerson, line 87: "from the secondary site to the primary site".

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+. Delete all Batch CRs created.
+
+. Login to the primary site.
+
+. Start Keycloak.
+
+Both {ispn} clusters are in sync and the switchover from secondary back to the primary site can be performed.
+
+=== AWS Aurora Database
```
ryanemerson, line 109: Should we state that this is optional?

doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
ahus1: I will add that they should change the number of instances to 0 in the CR.

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
ahus1: See above, will add this. Thanks for mentioning it.

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
ahus1: Thanks, fixed in next commit!

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
ahus1: I'm not sure which paragraph you've pointing your comment to, could you please elaborate?

doc/kubernetes/modules/ROOT/pages/running/network-partition.adoc
ryanemerson: > I had a call with Alexander and we thought about dropping the Batch CR, since we need the output of those commands, and use the CLI tool directly (`bin/cli.sh` from the pod or the `Route` endpoint).

+1, that would be simpler.

doc/kubernetes/modules/ROOT/pages/running/switch-back.adoc
ryanemerson: The switching back of the Aurora DB writer from one AZ to another. Long term there will be a performance benefit of having the writer in the same AZ as the Keycloak primary, however this comes at the expense of more downtime when recovering. So I was thinking this step could be optional as it will improve perf, but not affect data consistency.

Maybe it's simpler to just state that this is necessary so that the docs are more focused.

ahus1: Thank you. I'll add the following sentence in the next commit: 

> SSwitching the writer instance of Aurora will lead to a small downtime, and having the writer instance in the other datacenter with a slightly longer latency might be acceptable for some deployments.
So this might be deferred to a maintenance window or skipped depending on the circumstances of the deployment.

PR #593, opened 2023-11-17
Remove references to the map store

ahus1:
The map store is being removed from Keycloak nightly, the storage option is no longer available.


ahus1:
@kami619 - I've removed the references to the map store. Please review when you have the time. Thanks!

doc/kubernetes/modules/ROOT/pages/storage/embedded-infinispan.adoc
```diff
@@ -1,26 +0,0 @@
-= Configuring Embedded Infinispan
```
kami619, line 1: @ahus1 We would still have the embedded-infinispan as part of the Keycloak store in its current shape, do we want to move this information to a different place instead ?

ahus1:
@kami619 - thanks, re-adding the file. I see that it is not referenced anywhere (?), still that will need to be fixed in another commit.

kcb-ci-runner.sh
```diff
@@ -9,18 +9,8 @@ export RESET_KEYCLOAK=${RESET_KEYCLOAK:-true}
 export KEYCLOAK_STORAGE=${KEYCLOAK_STORAGE:-"JPA-Legacy-PostgreSQL"}
 if $RESET_KEYCLOAK; then
   cd $MINIKUBE_HOME
-  if [ "$KEYCLOAK_STORAGE" = "JPA-Map-PostgreSQL" ]; then
-    task reset-keycloak KC_STORAGE="jpa" KC_DATABASE="postgres"
-  elif [ "$KEYCLOAK_STORAGE" = "JPA-Legacy-PostgreSQL" ]; then
+  if [ "$KEYCLOAK_STORAGE" = "JPA-Legacy-PostgreSQL" ]; then
```
kami619, line 12:
```suggestion
  if [ "$KEYCLOAK_STORAGE" = "JPA-PostgreSQL" ]; then
```

I will adjust the Jenkins Job thats consuming this shell script.

kcb-ci-runner.sh
```diff
@@ -9,18 +9,8 @@ export RESET_KEYCLOAK=${RESET_KEYCLOAK:-true}
 export KEYCLOAK_STORAGE=${KEYCLOAK_STORAGE:-"JPA-Legacy-PostgreSQL"}
```
kami619, line 9:
```suggestion
export KEYCLOAK_STORAGE=${KEYCLOAK_STORAGE:-"JPA-PostgreSQL"}
```

provision/common/Taskfile.yaml
```diff
@@ -8,7 +8,6 @@ vars:
   KC_DB_POOL_INITIAL_SIZE: '{{default "5" .KC_DB_POOL_INITIAL_SIZE}}'
   KC_DB_POOL_MAX_SIZE: '{{default "10" .KC_DB_POOL_MAX_SIZE}}'
   KC_DB_POOL_MIN_SIZE: '{{default "5" .KC_DB_POOL_MIN_SIZE}}'
-  KC_STORAGE: '{{default "" .KC_STORAGE}}'
```
kami619, line 11: so we are effectively removing the `KC_STORAGE` as an `.env` option ? I think that makes sense.

ahus1: Yes, it is being removed.

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -71,27 +65,6 @@ spec:
...
-      value: jpa
-    - name: storage-area-auth-session
-      value: hotrod
-    - name: storage-area-user-session
-      value: hotrod
-    - name: storage-area-single-use-object
-      value: hotrod
-    - name: storage-area-login-failure
-      value: hotrod
-{{ else if ne .Values.storage "" }}
```
kami619, line 87: Would this configuration be not valid anymore ?

doc/kubernetes/modules/ROOT/pages/storage-configurations.adoc
```diff
@@ -3,10 +3,7 @@
 
 {description}
 
-Keycloak provides two storage implementations:
-
-* Legacy Storage
-* New Storage (also called Map Storage)
+This setup supports different databases.
```
kami619, line 6: currently it doesn't, as it only supports `postgres` right @ahus1 ?
```suggestion
This setup supports only PostgreSQL database.
```

provision/minikube/keycloak/templates/keycloak.yaml
ahus1: All storage configurations are removed, so it is no longer valid.

doc/kubernetes/modules/ROOT/pages/storage-configurations.adoc
ahus1: I counted Aurora as a second database :-)

kami619 (approved):
thanks @ahus1 for this extensive change with the map store removal, this should def save us some big chunks of time on PR CI jobs.

PR #594, opened 2023-11-17
Add Route53 LoadBalancing docs

ryanemerson:


ahus1:
Thank you @ryanemerson for this document. I added commit 5f7fc630ce26bed43889bc6c325cc8b46c651a7c with some review additions. 

I added a new guide `fail-over.adoc` which instructs admins to edit the route53 entry to prevent an automatic fall-back once  fail-over has occurred. 

I'm merging this PR now, all further changes would go into a second PR. Let's discuss on Monday if we have still severe gaps that we need to close, or when we will go into consolidating of docs and for example align terms and make them ready for publishing.

PR #596, opened 2023-11-20
Use Infinispan CLI tool instead of BatchCR

pruivo:
A little bit more verbose because I decided to add examples of the expected output. It makes it clearer IMO.


ahus1 (approved):
Hi, I really like the new guide with the CLI and the example output. Thank you!

PR #612, opened 2023-11-29
Allow mutlti-az workflow to create clusters in parallel

ryanemerson:


kami619 (approved):
lgtm, thanks for making this parallel and in theory this should save us 40 mins each time this workflow runs.

ahus1:
There used to be a problem when using inputs directly in the concurrency, therefore prepending `github.event.` like in other places. See 80672e7437ef0fa1081a4a8484f24205405fdad4 and https://github.com/orgs/community/discussions/35341 for more details.

PR #620, opened 2023-12-05
Disable xsite communication before initializing dataset

ryanemerson:
With xsite disabled the benchmark took ~40 minutes to populate the dataset.

https://github.com/ryanemerson/keycloak-benchmark/actions/runs/7101660153/job/19330256218


ahus1 (approved):
Thank you for updating the scripts and doing this test run!

PR #623, opened 2023-12-11
Mask passwords in GHA when executing scripts via go-task

martin-kanis:


provision/aws/rosa_oc_login.sh
```diff
@@ -22,13 +23,6 @@ if [[ "$RUNNER_DEBUG" == "1" ]]; then
   # prevent logging the password in debug mode
   set +x
 fi
-ADMIN_PASSWORD=$(aws secretsmanager get-secret-value --region $SECRET_MANAGER_REGION --secret-id $KEYCLOAK_MASTER_PASSWORD_SECRET_NAME --query SecretString --output text --no-cli-pager)
```
mhajas, line 25: @martin-kanis I think it still makes sense to not print `asw secretmanager` commands in debug mode. Could you please go through all usages of `aws secretmanager` in our script and make sure we use 
```
if [[ "$RUNNER_DEBUG" == "1" ]]; then
  # prevent logging the password in debug mode
  set +x
fi
aws secretmanager
if [[ "$RUNNER_DEBUG" == "1" ]]; then
  set -x
fi
```
everywhere?

mhajas (commented):
Thank you for the change @martin-kanis, could you please have a look at my comment?

provision/aws/rosa_oc_login.sh
martin-kanis: Thanks, added.

mhajas (approved):
Thank you @martin-kanis, LGTM

PR #639, opened 2023-12-14
Schedule CrossDCTest daily

### LINKED ISSUE #629: Schedule CrossDC tests daily

mhajas:
Closes #629

This PR schedules CrossDC functional tests to run daily on our clusters.

Here is a full run: https://github.com/mhajas/keycloak-benchmark/actions/runs/7208095186 
(it failed on committing benchmark results probably because my fork does not have result-data branch?)

It changes also the following:
- It runs workflows one after another instead of scheduling workflows at the time when we expect the previous workflow finished. We execute the following things daily:
  - Create two clusters `a` and `b`
  - Initialize monitoring
  - Create Keycloak deployment
  - Run functional tests
  - Run scalability benchmark
- It seemed clusters `a` and `b` are not created concurrently in our runs, it seems this: https://github.com/keycloak/keycloak-benchmark/blob/22a5a6ab0a1df4a126632a66aa557202e026772a/.github/workflows/rosa-cluster-create.yml#L59 does not work as expected (probably `github.event.inputs` does not return value) which results in running creating clusters sequentially. I changed this to running them in parallel and confirmed it works here: https://github.com/mhajas/keycloak-benchmark/actions/runs/7207592413 (the run failed but clusters were created successfully). You can compare the total time with this: https://github.com/keycloak/keycloak-benchmark/actions/runs/7204684045 which is today's daily run.


martin-kanis (approved):
Thanks @mhajas, looks good.

PR #640, opened 2023-12-14
Dynamically assign Aurora CIDR range based upon existing Aurora DB in…

ryanemerson:
For simplicity this solution assumes that CIDR ranges must be globally unique within the range `192.168.x.0/24`, resulting in a maximum limit of 256 Aurora DBs across all regions. If we ever need more instances than that, I can update the solution as required :stuck_out_tongue_winking_eye:


ahus1:
@ryanemerson - thank you very much for this change, did some small changes and then it looked good to be merged. 

One thing I found: the script `aurora_delete_peering_connection.sh` probably needs a refinement: At the moment it requires the `AURORA_CLUSTER` name to be set. IMHO it should iterate over all peering connections that exist for the given ROSA, determine the Aurora DB from he name, and use the proper region (either the region of the Aurora or the ROSA cluster). Please have a look next week and create an issue as necessary. It might explain dangling VPCs we have been seeing. Thanks!

ryanemerson:
> At the moment it requires the `AURORA_CLUSTER` name to be set. IMHO it should iterate over all peering connections that exist for the given ROSA, determine the Aurora DB from he name, and use the proper region (either the region of the Aurora or the ROSA cluster).

What happens if a ROSA cluster has multiple Aurora DBs associated with it and we only want to remove the peering connection associated with one of them?

ahus1:
> What happens if a ROSA cluster has multiple Aurora DBs associated with it and we only want to remove the peering connection associated with one of them?

Hm, thank you for pointing this out. We might need the script as is for doing what you describe. At the same time we're calling `aurora_delete_peering_connection.sh` from `rosa_delete_cluster.sh`, and for that case someone would need to iterate over all Aurora databases connected to that cluster. 

So this calls for a new script (wrapper to the existing one)?

ryanemerson:
That's a good point! I'll update `rosa_delete_cluster.sh` to make sure all peering connections are removed.

PR #646, opened 2023-12-18
Allow individual/all Aurora peering connections to be removed from a …

ryanemerson:
I've updated the scripts so that if `AURORA_CLUSTER` is provided, then only the peering connection specific to that cluster is removed, otherwise all peering connections associated with the ROSA cluster are removed. To enable this to work, I have also added an `AuroraCluster` tag to the peering connection at creation time. I have updated the existing Aurora peering connections to also contain this tag so that on merge, things should work as expected :crossed_fingers:.


ahus1:
Thank you - will keep an eye on it that the cleanup runs smoothly tonight.

PR #650, opened 2023-12-19
Update Route53 scripts for OCP 4.14.5

ryanemerson:


kami619 (approved):
lgtm

ahus1:
@ryanemerson - I assume the changes done here would need to be reflected in our blueprint here: https://www.keycloak.org/high-availability/deploy-aws-route53-loadbalancer 

Please create two follow-up issue in Keycloak's main repository for this and link them from here:

(a) an issue to update to list which ROSA version this original recipe was tested with in the prerequisites. This would then be backported to KC22 "timely" 
(b) a follow-up issue which updates the blueprint for OCP 4.14.x to use elbv2

If we think we are fast enough with (b) to be ready and backported to KC22, we can do only (b).

ryanemerson:
Done:

- A https://github.com/keycloak/keycloak/issues/25734
- B https://github.com/keycloak/keycloak/issues/25733

PR #659, opened 2023-12-20
Automate removal of Primary cluster from Route53 on site failure

ryanemerson:


provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,122 @@
...
+
+if [ -z "${HEALTH_CHECK_ID}" ]; then
+  echo "HEALTH_CHECK_ID variable must be set"
+  exit 1
+fi
+
+ALARM_NAME=${HEALTH_CHECK_ID}
+TOPIC_NAME=${HEALTH_CHECK_ID}
+FUNCTION_NAME=${HEALTH_CHECK_ID}
+ROUTE53_REGION="us-east-1"
```
kami619, line 19: @ryanemerson isn't Route53 a global resource, just for my understanding should we be still providing a route53 region for it ?

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,122 @@
...
+export AWS_REGION=${ROUTE53_REGION}
+
+TOPIC_ARN=$(aws sns create-topic --name ${TOPIC_NAME} \
+  --query "TopicArn" \
+  --tags "Key=HealthCheckId,Value=${HEALTH_CHECK_ID}" \
+  --output text
+)
+
+# Create CloudWatch alarm
+aws cloudwatch put-metric-alarm \
```
kami619, line 30: what is the outcome of this CloudWatch alarm action ? would it fail the HealthCheckStatus if it doesn't satisfy the threshold ?

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,122 @@
...
+        "Action": "sts:AssumeRole"
+      }
+    ]
+  }' \
+  --query 'Role.Arn' \
+  --output text
+)
+
+# Create a policy with the permissions required by lambda.py
+# TODO update permissions
```
kami619, line 64: is this a future TODO ?

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,122 @@
...
+  --function-name ${FUNCTION_NAME} \
+  --zip-file fileb://${LAMBDA_ZIP} \
+  --handler lambda.handler \
+  --runtime python3.11 \
+  --role ${ROLE_ARN} \
+  --query 'FunctionArn' \
+  --output text
+)
+
+aws lambda add-permission \
```
kami619, line 112: looks like we are adding perms here for the lambda, can we remove the prior TODO ?

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: Just an outdated comment, I have removed.

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: It's a global resource, but it seems CloudWatch alarms associated with Route53 must be created in `us-east-1`. It should be possible to deploy the Lambda function to a different region, as we use an SNS topic, but given that the Lambda only updates Route53 I figured it was easier to just keep it as one region.

https://stackoverflow.com/questions/44787306/can-a-cloudwatch-event-in-one-region-trigger-a-lambda-in-another-region-of-aws

kami619: Thank you, I was not aware of that.

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: > what is the outcome of this CloudWatch alarm action ?

When the alarm is triggered a notification is sent to the SNS topic. A subscription is created on this topic for the AWS lambda, so when a notifcation is received the function is triggered.

> would it fail the HealthCheckStatus if it doesn't satisfy the threshold ?

The alarm is triggered if the `HealthCheckStatus` metric of the Route53 health check fails.

kami619: great, thanks for the clarification Ryan.

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,121 @@
...
+  --role-name ${FUNCTION_NAME} \
+  --policy-arn ${POLICY_ARN}
+
+# Attach the AWSLambdaBasicExecutionRole policy so that the Lambda logs can be written to CloudWatch
+aws iam attach-role-policy \
+  --role-name ${FUNCTION_NAME} \
+  --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+
+# Wait for Role to be updated in order to avoid 'The role defined for the function cannot be assumed by Lambda' errors
+sleep 10
```
kami619, line 94: can we perform a role check if the role is updated or not in addition to the wait or do you think its not necessary ?

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,121 @@
...
+  --query 'Role.Arn' \
+  --output text
+)
+
+# Create a policy with the permissions required by lambda.py
+POLICY_ARN=$(aws iam create-policy \
+  --policy-name ${FUNCTION_NAME} \
+  --policy-document \
+  '{
+      "Version": "2012-10-17",
```
kami619, line 68: why is the version a date here ?

provision/aws/route53/route53_create_failover_lambda.sh
```diff
@@ -0,0 +1,121 @@
...
+aws cloudwatch put-metric-alarm \
+  --alarm-actions ${TOPIC_ARN} \
+  --actions-enabled \
+  --alarm-name ${ALARM_NAME} \
+  --dimensions "Name=HealthCheckId,Value=${HEALTH_CHECK_ID}" \
+  --comparison-operator LessThanThreshold \
+  --evaluation-periods 1 \
+  --metric-name HealthCheckStatus \
+  --namespace AWS/Route53 \
+  --period 60 \
```
kami619, line 39: At what stage the 60 second evaluation period starts for this metric alarm ?

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: I believe this is just a quirk of AWS, as this value was the default when creating policies and is found in all examples. We use the same value when creating policies in `aurora_create_global_db.sh`.

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: I believe it starts as soon as the alarm is created.

provision/aws/route53/route53_create_failover_lambda.sh
ryanemerson: I tried this originally for `aurora_create_global_db.sh` where was also have to wait on the creation, however I was getting strange false positives where the role was seemingly updated when checking via the CLI but hadn't propagated to AWS lambda yet.

kami619: got it, its probably the `Eventual Consistency` quirk of AWS. Thanks Ryan.

kami619 (approved):
LGTM

kami619:
I will merge this PR, once we run it in a failure test scenario.

tkyjovsk (changes requested):
Could we put a short note in the script about the hardcoded `us-east-1` region? Otherwise LGTM.

I've read through the doc and did a manual test of `route53_create_failover_lambda.sh` and `route53_delete_failover_lambda.sh` scripts with the primary health check. With the function present it was necessary to go and edit the healthcheck in order to fail back.

With the function deleted the failback was automatic as soon as the health route appeared again.

The `route53_test_primary_used.sh` script works too.

provision/aws/route53/route53_create_failover_lambda.sh
kami619: i'll add this to the script and merge it

tkyjovsk (approved):

PR #664, opened 2024-01-02
Document oc binary build for OpenShift

ahus1:
I suppose this is simple enough so one wouldn't need an external registry. 

Once we also build the operator, this would be a similar procedure. When building the Operator as well (to match a Keycloak release), only putting the Operator into an external registry alone doesn't solve the problem of also storing the CRD, which depends on the version of the Operator, so using an external registry doesn't solve much on its own for an OpenShit and Operator deployment.


mhajas (approved):
Thank you for this @ahus1, it works as expected. I pushed one small change to the image name and I am merging this now.

ahus1:
Thank you for spotting this!

PR #668, opened 2024-01-03
Clean dangling DHCP Options

mhajas:
This is a workaround for the following issue https://issues.redhat.com/browse/OCPBUGS-1838. We can watch the script output and maybe provide a reproducer for the issue if we find it.

We have around 30 DHCP options still present in our account. I left it present if anyone would like to test the script.


kami619:
would this become moot once we switch to AWS CDK ? @mhajas

mhajas:
Not sure what you mean @kami619. Like I should write this code using AWS CDK? Or that we won't need this when we use AWS CDK?

kami619:
I mean, we might not have to handle the iterative deletes once we implement delete objects with  AWS CDK. But we could still use this for the time being.

doc/kubernetes/modules/ROOT/pages/openshift/cross-site-rosa.adoc
```diff
@@ -99,3 +98,9 @@ AWS provides a https://github.com/awslabs/aws-advanced-jdbc-wrapper[JDBC driver
 To disable the AWS JDBC driver, set the `KC_USE_AWS_JDBC_WRAPPER` variable to `false`.
 
 To specify the version of the AWS JDBC driver, set the `KC_AWS_JDBC_WRAPPER_URL` variable to the URL of corresponding jar file.
+
+=== Warnings / Known issues
+
+* Using `--force` will break the keycloak deployment because it patches the keycloak operator deployment with an incorrect image.
```
mhajas, line 104: I believe this is no longer a problem as we are not patching operator deployment when setting the Keycloak image.

PR #669, opened 2024-01-04
Create non-expirable SA token for cross-site

### LINKED ISSUE #663: Expired X-Site token prevents the work of the Infinispan Operator
### Describe the bug

As described in #607, I was able to reproduce the problem and the error, that the operator didn't restore the deployments on the primary cluster, even without the load. 

I found the following error message in the logs of the operator: 

> ERROR	controllers.Infinispan	could not connect to remote location URL	{"infinispan": "runner-keycloak/infinispan", "URL": "openshift://api.gh-keycloak2-b.wasw.p1.openshiftapps.com:6443", "error": "Unauthorized"}

The root cause is that the token for the other cluster generated in https://www.keycloak.org/high-availability/deploy-infinispan-kubernetes-crossdc is valid for only 60 minutes. 

### Version

main

pruivo:
Fixes #663

### DIFF
--- a/provision/infinispan/sa-token-secret.yaml
+++ b/provision/infinispan/sa-token-secret.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ispn-xsite-sa-token
+  annotations:
+    kubernetes.io/service-account.name: "${SA_ACCOUNT}"
+type: kubernetes.io/service-account-token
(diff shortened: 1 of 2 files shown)


provision/infinispan/sa-token-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: rhdg-xsite-sa-token
```
kami619, line 4: nitpick, do we want this to be `ispn-xsite-sa-token` instead ?

kami619 (approved):
lgtm

PR #670, opened 2024-01-08
Use non-expiring token for Grafana and HTTPS

pruivo:
Related to
https://github.com/keycloak/keycloak-benchmark/commit/0598144ef9f3ef31f2edf48a5bb0accd334a0bc7
https://github.com/keycloak/keycloak-benchmark/commit/5e88c4ab1414b6b106e6d862017cd98a0b69c37e


ahus1:
Thanks, I missed that the code was duplicated there.

PR #671, opened 2024-01-08
Enable JGroups Diagnostics

pruivo:
Not recommended in production since the port is not secure and allows to observe and manipulate JGroups protocols/attributes


pruivo:
@ryanemerson looking a the operator code, it reacts to `JGROUPS_DIAGNOSTICS`. This should work, right? I'll test it tomorrow when I start a new cluster.

pruivo:
Works like a charm :+1:

ryanemerson (approved):

ahus1:
@pruivo - I've updated the script to have the commit comment warning in the shell script, so people will see it if they use something similar in their environment.

PR #672, opened 2024-01-08
update shebang line of all scripts to be more compatible between *nix environments

alice-wondered:
On OSX specifically, the OS ships with a very old version of bash that's missing many features used in various scripts in this project. Installing from brew allows for the env bin to point to a much newer version of bash and resolves this issue

### DIFF
--- a/antora/author.sh
+++ b/antora/author.sh
@@ -1,3 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+set -e
 yarn install
 yarn run author
--- a/ansible/aws_ec2.sh
+++ b/ansible/aws_ec2.sh
@@ -1,6 +1,10 @@
-#!/usr/bin/env bash +x
+#!/usr/bin/env bash
 set -e
 cd $(dirname "${BASH_SOURCE[0]}")
+
+if [[ "$RUNNER_DEBUG" == "1" ]]; then
+  set -x
+fi
 
 OPERATION=$1
 REGION=$2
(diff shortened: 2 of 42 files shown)


ahus1 (changes requested):
Thank you for fixing this. Could you please add the DCO to your commit like we do in the Keycloak main repository? Once that is in, it is good for merge.

kami619:
@ahus1 https://github.com/keycloak/keycloak-benchmark/actions/runs/7453510299/job/20328347373?pr=672#step:4:5 Please note that the Documentation CI job failed with the recent change from @alice-wondered

antora/author.sh
```diff
@@ -1,3 +1,3 @@
-#!/bin/bash -e
+#!/usr/bin/env bash -e
```
ahus1, line 1: With the GitHub action failing, and when looking at it in my IDE (IntelliJ) which Shell Check enabled, it provided the following link that explains the problem and the remedy: 

https://github.com/koalaman/shellcheck/wiki/SC2096

Please update all occurrences like the one below.

```suggestion
#!/usr/bin/env bash 
set -e
```

ahus1 (changes requested):
Apparently there was a problem lurking in this - could you please have a look and fix the scripts, so they work well with the parameter? Thanks!

antora/author.sh
alice-wondered: No problem, I'll run through and correct these

alice-wondered: Sorry about the oversight! Should be resolved now (and I also picked up that shell check plugin 😄)

ahus1 (approved):
Thank you for this change!

PR #677, opened 2024-01-15
Fix delete-aurora job in rosa-multi-az-cluster-delete.yml

ryanemerson:


ahus1:
@ryanemerson - can you please provide more context when this happens? 

Do we really want to pass in the region as a parameter, or can our current scripts cope without?

ryanemerson:
None of our automations use this workflow currently, which is why the issue(s) have gone undetected. Without these changes attempts to call the `rosa-multi-az-cluster-delete.yml` workflow will fail.

> Do we really want to pass in the region as a parameter, or can our current scripts cope without?

Currently this field is mandatory when calling `aurora_delete.sh`.

PR #678, opened 2024-01-15
Automatic failover lambda round 2

ryanemerson:
I have restored the original two commits and then added a third commit with the required changes.

This PR adds the following tasks to the rosa-cross-dc Taskfile:

- deploy-route53-failover-lambda
- undeploy-route53-failover-lambda
- `route53-failback`: Reverts changes to the healthcheck so that the Primary can once again become healthy

When deploying a multi-az cluster using the taskfile, the Lambda is now created once the Keycloak deployment is up and healthy. Similarly, when undeploying the Lambda is removed and the healthcheck is updated to use the normal resource path `/lb-check` so that a new deployment can proceed using the existing Route53 domains.

The `route53-failback` task can be used with failover test automation, or the `route53_failback.sh` script can be used directly.  AFAICT the `kc-failover.sh` script is not executed as part of the CI testsuite... please let me know if it is and I've just missed it :slightly_smiling_face:.


kami619:
@ryanemerson this PR seems to be out of sync with main branch, would you fix the conflicts and we shall try and test it again ?

ahus1:
Rebased due to conflicts which appeared after changes to `main`

mhajas (changes requested):
I didn't go through this fully but it seems cross-dc test is not updated with the new approach to failover recovery. Please do not merge until this is there.

ryanemerson:
> I didn't go through this fully but it seems cross-dc test is not updated with the new approach to failover recovery. Please do not merge until this is there.

Do you mean the `logoutUserWithFailoverTest`? Looking at the current code, I don't think there's any need for it to be updated from a Keycloak function POV as failover from the client perspective is the same as before.

Or do you mean to update the code to verify that the Route53 health check has been updated on the primary by the Lambda? That should be possible, but we will need to add AWS java libraries.

ryanemerson:
> However, after running the `kc-failover.sh` script the site didnt failover for some reason, and if I understand it correctly only after the primary site is marked failed, the lambda would be triggered based on the metric alarm, need some guidance in debugging this, I will try to repeat the tests on a fresh install and see if it makes a difference.

My steps for deploying and testing are as follows:

1. `cd provision/rosa-cross-dc`

2. `ROSA_CLUSTER_NAME_1=  ROSA_CLUSTER_NAME_2= task route53`

3. `DOMAIN= KC_CLIENT_URL=client.$DOMAIN KC_HEALTH_URL_CLUSTER_1=primary.$DOMAIN KC_HEALTH_URL_CLUSTER_2=backup.$DOMAIN ROSA_CLUSTER_NAME_1= ROSA_CLUSTER_NAME_2= AURORA_CLUSTER=_REGION= task`

4. Login to primary Openshift cluster

5. `PROJECT= DOMAIN= FAILOVER_MODE=HEALTH_PROBE FAILOVER_DELAY=0 ./benchmark/src/main/content/bin/kc-failover.sh`

Once the health route has been deleted, the `kc-failover.sh` script may take a while to failover due to DNS caching. You can verify that the Lambda has been executed quicker by inspecting the health checks in the Route53 console. Once failover has occurred, it should be possible to see the primary health check using `/lb-check-failed-over` and it should also say "1 of 1 in ALARM" in the alarms column.

kami619:
@ryanemerson thanks for the details, based on what we went through together, the functionality seems to be now working as expected. Once we address the testsuite dependency on the `lb-check` endpoint we should be good to merge this one.

ryanemerson:
[AWSClient.java](https://github.com/keycloak/keycloak-benchmark/pull/678/commits/a68a7a958bdeabb4db9e3b78541680331bbe8ab6#diff-fd59188fa4d0b16186ede0606dcedbf51acae4a480bca3430b6e19f9f80b713d) added to abstract Route53 logic.Successful functional test run on my fork: https://github.com/ryanemerson/keycloak-benchmark/actions/runs/8051617435

kami619:
thanks @ryanemerson I have some time today, I will test this and merge it if it looks good.

kami619:
@mhajas This change to make the testsuite run with the change to update the lb-check endpoint has been addressed. I have reviewed the change and ran the tests locally and it works fine for me, the same it works for Ryan. Take a look when you have a chance and we can merge this probably.

Note: This does create a dependency on the AWS load balancer check, however Ryan added it in such a way, it should be easier to manipulate with a different type of implementation.

kami619 (approved):
lgtm

ahus1:
@mhajas - with Kamesh' review in, are you ok to merge or would you like to re-review based on your previous review which requested changes?

mhajas:
We should be good to merge this based on the review from @kami619. Just a question, do we plan to document this for A/P deployment? As far as I understand this is not necessary for A/A so I am thinking whether it is worth it to merge.

ahus1:
@mhajas - the reason we created this PR was to make it simpler/safer for users of A/P to prevent a fail-back when the primary site isn't good to be switched back to. 

* If we think this automation makes it more complicated, we should not merge this PR. 
* If we think this automation makes it simpler (both for ourselves and users of Keycloak), we should merge this PR and update the Keycloak HA documentation

We can pick this up in today's meeting.

ahus1:
We discussed that with A/A we might have more of those lambdas to allow fencing of A/A site, so let's merge this and learn in practice about benefits and shortcomings of this approach.

Documenting in the main Keycloak HA docs is deferred until we get some practical experience, and possibly some feedback from the community after a blog post or email to the mailing list.

Issue #637, opened 2023-12-13
Horreum: Include environment related information


andyuk1986:
- Implement logic for including environment related information in JSON file, i.e. the used database type, Multi AZ vs single-AZ, with or without external Infinispan, Keycloak version, etc.
- Update the schema Labels in Horreum correspondingly.

andyuk1986:
This task also includes the update to the existing JSON structure.

andyuk1986:
Environment data is properly into Horreum report using configmap. Please find attached the example of JSON report which was generated yesterday during test run.

[result-20240114-231327-4ec405b5-28f5-4444-9c4e-f8176394d287.json](https://github.com/keycloak/keycloak-benchmark/files/13939777/result-20240114-231327-4ec405b5-28f5-4444-9c4e-f8176394d287.json)

andyuk1986:
Here is the updated one updated with additional aurora database information:


[result-20240115-161429-2cbca6d0-d864-4893-a68e-ad7c38d15f80.json](https://github.com/keycloak/keycloak-benchmark/files/13942192/result-20240115-161429-2cbca6d0-d864-4893-a68e-ad7c38d15f80.json)

andyuk1986:
Sent a PR: https://github.com/keycloak/keycloak-benchmark/pull/679

PR #679, opened 2024-01-16
Rosa Scaling Benchmark: JSON Update with environment data

andyuk1986:
#637

### DIFF
--- a/.github/actions/prepare-horreum-report/action.yml
+++ b/.github/actions/prepare-horreum-report/action.yml
@@ -1,0 +1,33 @@
+name: Prepare Horreum Report
+description: Prepares JSON report before further processing, i.e. creates the JSON with the env data which gets from the configmap.
+
+inputs:
+  project:
+    description: 'The project namespace.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - id: prepare-report-file
+      name: Create and Prepare Report File
+      shell: bash
+      # language=bash
+      run: |
+        output_file_prefix="result-"
+        cur_date=$(date)
+        cur_date_iso=$(date -d "$cur_date" --iso-8601=seconds)
+        cur_date_iso_compressed=$(date -d "$cur_date" '+%Y%m%d-%H%M%S')
+        uuid=$(uuidgen)
+        OUTPUT_FILE_NAME="${output_file_prefix}${cur_date_iso_compressed}-${uuid}.json"
+        echo "HORREUM_OUTPUT_FILE_NAME=$OUTPUT_FILE_NAME" >> $GITHUB_ENV
+        jq -n --arg current_date "${cur_date_iso}" --arg id "${uuid}" \
+        '{"$schema": "urn:keycloak-benchmark:0.1", "uuid": ($id), "name": "ROSA Scalability Benchmark Run Results",
+        "start_dttm": ($current_date)}' > ${OUTPUT_FILE_NAME}
+        #Reading configmap with environment data
+        configJson=$(oc get configmap env-config -n ${{ env.PROJECT }} -o "jsonpath={ .data['environment_data\.json']}'" | rev | cut -d\' -f2- | rev | jq)
+        jq '. + {"context":{}}' ${OUTPUT_FILE_NAME} > tmp.json && \
+        mv tmp.json ${OUTPUT_FILE_NAME}
+        #Putting environment parameters into JSON
+        jq --argjson configJson "${configJson}" '.context = ($configJson)' ${OUTPUT_FILE_NAME} > tmp.json && \
+        mv tmp.json ${OUTPUT_FILE_NAME}
--- a/provision/rosa-cross-dc/Taskfile.yaml
+++ b/provision/rosa-cross-dc/Taskfile.yaml
@@ -16,6 +16,7 @@ vars:
   ANSIBLE_DIR: "{{.ROOT_DIR}}/../../ansible"
   PYTHON_DIR: "{{.ROOT_DIR}}/../../benchmark/src/main/python"
   BENCHMARK_DIR: "{{.ROOT_DIR}}/../../benchmark/src/main/content/bin"
+  ENV_DATA_JSON_PATH: "{{.ROOT_DIR}}/../environment_data.json"
   KEYCLOAK_MASTER_PASSWORD:
     sh: aws secretsmanager get-secret-value --region eu-central-1 --secret-id keycloak-master-password --query SecretString --output text --no-cli-pager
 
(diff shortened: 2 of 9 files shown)


andyuk1986:
Hey, during testing I have found an issue in configmap deletion part. Please look into this PR after the PR is updated with the fix.

kami619:
@andyuk1986 I am reviewing this PR, thanks for finding and fixing the issue with the configMap 🚀

kami619:
@andyuk1986 are there any runs that you did with these GH workflow files ?

.github/actions/prepare-horreum-report/action.yml
```diff
@@ -0,0 +1,32 @@
...
+  using: composite
+  steps:
+    - id: prepare-report-file
+      name: Create and Prepare Report File
+      shell: bash
+      # language=bash
+      run: |
+        output_file_prefix="result-"
+        cur_date_iso=$(date --iso-8601=seconds)
+        cur_date_iso_compressed=$(date '+%Y%m%d-%H%M%S')
```
kami619, line 19: just a nit pick, the `cur_date_iso` and `cur_date_iso_compressed` would be slightly different, calling it out just in case.

kami619:
@andyuk1986 perfect, I think those runs great and love the new structure of the environmental data resolved json

```
{
  "$schema": "urn:keycloak-benchmark:0.1",
  "uuid": "b2e22919-0316-4136-9d8a-35075a7e1bbb",
  "name": "ROSA Scalability Benchmark Run Results",
  "start_dttm": "2024-01-17T12:02:28+00:00",
  "context": {
    "num_of_pods": 3,
    "cpu_limit_per_pod": 6,
    "sticky_session_disabled": "true",
    "mem_requests_per_pod": 4000,
    "mem_limit_per_pod": 4000,
    "heap": {
      "init": 64,
      "max": 2048
    },
    "db_pool": {
      "init": 30,
      "min": 30,
      "max": "30"
    },
    "metaspace": {
      "init": 96,
      "max": "256"
    },
    "is_multi_az": "true",
    "keycloak_database": "aurora-postgres",
    "database_mode": "single",
    "ispn_deployment": "ispn:crossdc"
  },
  "memory-usage-test": {
    "active-sessions-per-500Mb-per-pod": 16307
  },
  "cpu-usage-for-logins-test": {
    "user-logins-per-sec-per-1vCpu-per-pod": 36
  },
  "cpu-usage-for-credential-grants-test": {
    "credential-grants-per-sec-per-1vCpu": 321
  }
}

```

.github/workflows/rosa-multi-az-cluster-clear.yml
```diff
@@ -0,0 +1,66 @@
+name: Multi-AZ Clusters - Clear
```
kami619, line 1: @andyuk1986 my suggestion would be to name this workflow to be `rosa-multi-az-cluster-undeploy.yml` and the name to be `Multi-AZ Clusters - Undeploy` that would make the workflow consistent with the downstream tasks such as `task undeploy. Its a great Idea by the way to have the steps sorted this way, thanks for doing that :)

provision/rosa-cross-dc/Taskfile.yaml
```diff
@@ -420,6 +434,12 @@ tasks:
           ROSA_CLUSTER_NAME: "{{.ROSA_CLUSTER_NAME_2}}"
           KC_HOSTNAME_OVERRIDE: "{{.KC_CLIENT_URL}}"
           KC_HEALTH_HOSTNAME: "{{.KC_HEALTH_URL_CLUSTER_2}}"
+      - task: create-env-configmap
```
kami619, line 437: Is this environment config map per cluster or per namespace in the ROSA cluster ? Would it uniquely identify a particular deployment on the cluster ?

kami619: or is our assumption to have only one deployment per cluster during these benchmark tests ?

andyuk1986: the env config map is created on the cluster per namespace. So it is unique for each *-keycloak namespace, and it is possible to run the benchmark tests on different namespaces and we will have the generated JSON with corresponding data.

.github/workflows/rosa-multi-az-cluster-clear.yml
andyuk1986: thanks a lot for the suggestion, I have renamed the job and the file correspondingly.

andyuk1986:
@kami619 so I have fixed the issue with configmap deletion (the job was failing if the configmap was not available in the cluster). 
Also during testing I have noticed that in multi-az delete and in my undeploy workflows, SUBDOMAIN parameter is passed in a wrong way to route53 deletion script, so the SUBDOMAIN is not passed and the script deletes all route53 records instead of the only ones which are necessary. So I have fixed the route53_delete.sh script calls with proper parameter passing. 
The same refers to aurora database - I have updated the undeploy workflow, to have check-mark asking if to skip Aurora Database delete or no -> this is necessary in case there are more than 1 keycloak namespaces on the cluster. The process is the following: while undeploying the cluster, the workflow undeploys the default deployment, i.e. runner-keycloak.  But it is possible to pass KC_NAMESPACE_PREFIX with env variables, and in that case the specific deployment will be undeployed. But aurora database is unique for the cluster (is not connected to namespace), so I have added a checkbox for workflow for keeping aurora if needed. 
Another issue that I have found is in our daily scheduled deletion job, the clusters' delete considers "keepalive" namespace. This doesn't refer to route53 records and aurora database deletion scripts. So we mark the cluster to keep alive, but route53 and aurora database of that cluster are deleted, which leads later to miss-functioning of that cluster (need to run again aurora create, route53 create, for being able to use the cluster). If you don't mind I will send the fix for this in separate PR.

kami619:
Awesome, thanks so much @andyuk1986 

> Another issue that I have found is in our daily scheduled deletion job, the clusters' delete considers "keepalive" namespace. This doesn't refer to route53 records and aurora database deletion scripts. So we mark the cluster to keep alive, but route53 and aurora database of that cluster are deleted, which leads later to miss-functioning of that cluster (need to run again aurora create, route53 create, for being able to use the cluster). If you don't mind I will send the fix for this in separate PR.

Sure thing, this has been a longstanding issue, fixing this would definitely save 15 mins of every developer's time. Kudos to you ! 🥇

provision/rosa-cross-dc/Taskfile.yaml
kami619: perfect, thanks for letting me know.

kami619 (approved):

PR #681, opened 2024-01-16
Fine tune Infinispan timeouts

pruivo:
Fine tunning timeouts in Infinispan and JGroups

* Tunnel Heartbeats timeout must be lower than Infinispan cross-site timeout to allow the offline gossip router to be detected and retransmissions to be forwarded to the online gossip router without timing out the write operation at Infinispan
* Failure detection timeout must be higher than Tunnel heartbeat timeout to prevent a node from being suspected in case a gossip router is down
* Reduce timeout in Infinispan server to prevent Keycloak timeouts; although, I'm not sure if it is a good idea or not at this point. 

Edit: fixed the grafana snapshot python script to use HTTPS instead of HTTP.

### DIFF
--- a/benchmark/src/main/python/snapGrafana.py
+++ b/benchmark/src/main/python/snapGrafana.py
@@ -53,7 +53,7 @@ async def generate_pdf(
 async def generate_grafana_api_token(dashboard_domain, admin_password):
     print(f"CREATE GRAFANA API KEY")
     dttm = get_current_time()
-    url =  f'http://admin:{admin_password}@{dashboard_domain}/api/auth/keys'
+    url =  f'https://admin:{admin_password}@{dashboard_domain}/api/auth/keys'
     reqBody = {'name':f'dashboard_pdf_{dttm}','role':'Editor','secondsToLive':86400}
     auth_token_id = ""
     auth_token = ""
@@ -75,7 +75,7 @@ async def generate_grafana_api_token(dashboard_domain, admin_password):
 
 async def delete_grafana_api_key(dashboard_domain, admin_password, auth_token_id):
     print(f"DELETE GRAFANA API KEY")
-    url =  f'http://admin:{admin_password}@{dashboard_domain}/api/auth/keys/{auth_token_id}'
+    url =  f'https://admin:{admin_password}@{dashboard_domain}/api/auth/keys/{auth_token_id}'
     try:
         response = requests.delete(url)
         response.raise_for_status()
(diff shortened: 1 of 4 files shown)


pruivo:
Gossip Router failing scenario with patched Infinispan/JGroups image.
Default timeouts:
default-timeout

This pull request:
pr-timeout

ahus1:
Thank you for these optimizations. Maybe @ryanemerson can review them from an Infinispan point of view. 

Looking at future code changes and also to the users of Keycloak, I'd like to see the rationales in the description in the PR present in the code, so users of Keycloak Benchmark as well as users of Keycloak don't change it in a wrong way in the future. 

Possible ways for this could be: 
* add the constrains you described in the ticket above as comments to the different fields in `infinispan.yaml`, so they appear in the generated documentation
* add the constraints as comments to the Infinispan `values.yaml`, and add a more descriptive text to the Keycloak docs (which could evolve to its own concepts page around Infinispan timeouts/configurations similar to  https://www.keycloak.org/high-availability/concepts-threads or https://www.keycloak.org/high-availability/concepts-database-connections

Let me know which way you prefer, or if you see a different way to achieve a similar thing.

ryanemerson (approved):
The changes LGTM.

+1 on adding some form of documentation to explain the rationale behind these values.

pruivo:
I assume the docs changes will be in the Keycloak project and not in this PR. In any case, this would fit better in a tunning guide.

ahus1:
Ok, added a backlog item to our project board "Add Infinispan performance tuning guide". Once I merge this, the next time we're copying over the generated YAMLs, I assume those values will show up in the docs on the Keycloak homepage.

pruivo:
The heartbeats/fd timeout won't show in the guide. The cache timeout will.

I've talked to Bela and I'm having second thoughts about it. We are optimizing for an edge case that may or may not happen and 99% of the time, everything will be fine. It may be a waste of resources sending heartbeats every second just to save some seconds from the detection time (during my test, the default timeouts took ~40s longer to detect and correct itself). 

I agree we should have a (tunning) guide explaining all these concepts so users can make decisions on how they want to tolerate faults. Just remember, there is no such thing as a free lunch, there is always a cost :+1: 

Another concern is if we want to do the same tunning in Keycloak cluster. I do not want to do 1-on-1 mapping of JGroups attributes into CLI options :scream: :fearful: 

cc @ahus1 @ryanemerson

PR #683, opened 2024-01-17
Add Maven enforcer plugin

ryanemerson:


ahus1:
@ryanemerson - looks good in general, but it seems to me the parameter needs a bit of tuning as it currently fails. 
I addition, I'd like to see a code comment why this restriction was put in place (that was Scala found not working JDK21).

PR #684, opened 2024-01-17
Add missing KUBECONFIG to kubectl commands

pruivo:
* Added the missing `KUBECONFIG` otherwise it uses the one defined in your machine
* Set `defer` after the creation of the `Batch` command.

### DIFF
--- a/provision/infinispan/Utils.yaml
+++ b/provision/infinispan/Utils.yaml
@@ -461,9 +461,8 @@ tasks:
         - NAMESPACE
     cmds:
       - task: rosa-oc-login
-
       - |
-        cat << EOF | kubectl -n {{ .NAMESPACE }} apply -f -
+        cat << EOF | KUBECONFIG=".task/kubecfg/{{.ROSA_CLUSTER_NAME}}" kubectl -n {{ .NAMESPACE }} apply -f -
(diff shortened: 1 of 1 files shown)


provision/infinispan/Utils.yaml
```diff
@@ -472,10 +471,9 @@ tasks:
           cluster: infinispan
           configMap: crossdc-connect
         EOF
-      - kubectl -n {{ .NAMESPACE }} wait --for=jsonpath='{.status.phase}'=Succeeded Batch/bring-online
-      - defer: |
-          kubectl -n {{ .NAMESPACE }} logs job/bring-online || true
-          kubectl -n {{ .NAMESPACE }} delete Batch/bring-online || true
+      - defer: KUBECONFIG=".task/kubecfg/{{.ROSA_CLUSTER_NAME}}" kubectl -n {{ .NAMESPACE }} delete Batch/bring-online || true
```
ryanemerson, line 474: We're no longer printing the logs in the defer. Won't this mean that if the `kubectl wait` condition fails we'll never get the logs to debug what went wrong?

pruivo: fixed

ryanemerson (approved):

PR #685, opened 2024-01-17
Use JGroups MTLS setup with certmanager generated certificates

ahus1:
Related to: keycloak/keycloak#25703

Until the PR listed above is merged and the code is available in a nightly build, a locally built image will need to be used.

### DIFF
--- a/provision/openshift/Taskfile.yaml
+++ b/provision/openshift/Taskfile.yaml
@@ -6,6 +6,7 @@ vars:
   KC_HOSTNAME_SUFFIX: '{{default "$(kubectl get route/console -n openshift-console -o jsonpath=\u0027{.spec.host}\u0027 | cut -d . -f 2-)" .KC_HOSTNAME_SUFFIX}}'
   KC_NAMESPACE_PREFIX: '{{default "$(whoami)-" .KC_NAMESPACE_PREFIX}}'
   KC_ADMIN_PASSWORD: '{{default "$(aws secretsmanager get-secret-value --region eu-central-1 --secret-id keycloak-master-password --query SecretString --output text --no-cli-pager || echo admin)" .KC_ADMIN_PASSWORD}}'
+  KC_JGROUPS_TLS: '{{default "true" .KC_JGROUPS_TLS}}'
 
 output: prefixed
 
@@ -251,6 +252,7 @@ tasks:
         --set heapMaxMB={{ .KC_HEAP_MAX_MB }}
         --set metaspaceInitMB={{ .KC_METASPACE_INIT_MB }}
         --set metaspaceMaxMB={{ .KC_METASPACE_MAX_MB }}
+        --set infinispan.jgroupsTls={{ .KC_JGROUPS_TLS }}
         --set infinispan.customConfig={{ .KC_CUSTOM_INFINISPAN_CONFIG }}
         --set infinispan.configFile={{ .KC_CUSTOM_INFINISPAN_CONFIG_FILE }}
         --set infinispan.remoteStore.enabled=$(cat .task/remote-store-enabled)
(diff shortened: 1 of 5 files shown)


provision/openshift/Taskfile.yaml
```diff
@@ -6,6 +6,7 @@ vars:
   KC_HOSTNAME_SUFFIX: '{{default "$(kubectl get route/console -n openshift-console -o jsonpath=\u0027{.spec.host}\u0027 | cut -d . -f 2-)" .KC_HOSTNAME_SUFFIX}}'
   KC_NAMESPACE_PREFIX: '{{default "$(whoami)-" .KC_NAMESPACE_PREFIX}}'
   KC_ADMIN_PASSWORD: '{{default "$(aws secretsmanager get-secret-value --region eu-central-1 --secret-id keycloak-master-password --query SecretString --output text --no-cli-pager || echo admin)" .KC_ADMIN_PASSWORD}}'
+  KC_JGROUPS_TLS: '{{default "true" .KC_ADMIN_PASSWORD}}'
```
pruivo, line 9: typo?
```suggestion
  KC_JGROUPS_TLS: '{{default "true" .KC_JGROUPS_TLS}}'
```

ahus1: Nice find!

pruivo:
`Taskfile.yaml` in `rosa-cross-dc` directory needs to be updated too.

My suggestion is to add the following line `--set infinispan.jgroupsTls={{ .KC_JGROUPS_TLS | default false }}` around this line: 
https://github.com/keycloak/keycloak-benchmark/blob/9451873f69b244c4825d6b164783d376bff54668/provision/rosa-cross-dc/Taskfile.yaml#L280

pruivo:
I've tested this using `openssl s_client -connect <keycloak-pod>:<jgroups-port>` and it shows a certificate:

```
CN = keycloak-jgroups
Requested Signature Algorithms: ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:Ed25519:Ed448:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA+SHA256:RSA+SHA384:RSA+SHA512
Shared Requested Signature Algorithms: ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:Ed25519:Ed448:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA+SHA256:RSA+SHA384:RSA+SHA512
Peer signing digest: SHA256
Peer signature type: RSA-PSS
Server Temp Key: X25519, 253 bits
```

The JGroups view was formed properly too.

ahus1:
I've incorporated your suggestions.

> The JGroups view was formed properly too.

Could you think of a good check to add to a readiness probe where we could detect a mis-configuration? 
Something like "the discovery returned some IP addresses, but I wasn't able to connect to them"? It should be a check that would still succeed when there is just a single node, but that would fail once there are two nodes and they are unable to connect to each other especially in situations where the encryption can't be established.

provision/rosa-cross-dc/Taskfile.yaml
```diff
@@ -17,6 +17,7 @@ vars:
   BENCHMARK_DIR: "{{.ROOT_DIR}}/../../benchmark/src/main/content/bin"
   KEYCLOAK_MASTER_PASSWORD:
     sh: aws secretsmanager get-secret-value --region eu-central-1 --secret-id keycloak-master-password --query SecretString --output text --no-cli-pager
+  KC_JGROUPS_TLS: '{{default "true" .KC_JGROUPS_TLS}}'
```
pruivo, line 20: The global variable is not required. If you want to keep it, the line 280 can be simplified:
`--set infinispan.jgroupsTls={{ .KC_JGROUPS_TLS }}`

ahus1: Thanks. I would like to keep it as it serves as some kind of reference at the top of the file. And I'd like to keep it `true`. 

Simplified the setup below as you suggested.

pruivo (approved):

ahus1:
> Could you think of a good check to add to a readiness probe where we could detect a mis-configuration?

Created a follow-up issue #692

PR #688, opened 2024-01-19
Add support for more benchmark options

pruivo:
* Added user-per-sec and duration
* Added refresh token count and period


pruivo:
Probably I'm the only one using these tasks :)

kami619 (approved):
LGTM

PR #697, opened 2024-01-25
Fix required verify user profile failure

mhajas:


provision/rosa-cross-dc/keycloak-benchmark-crossdc-tests/src/test/java/org/keycloak/benchmark/crossdc/client/KeycloakClient.java
```diff
@@ -137,7 +139,10 @@ public Map<String, Object> refreshToken(String realmName, String refreshToken, S
     }
 
     public String usernamePasswordLogin(String realmName, String username, String password, String clientId) throws IOException, URISyntaxException, InterruptedException {
-        String formUrl = openLoginForm(realmName, clientId);
+        HttpResponse<String> loginFormResponse = openLoginForm(realmName, clientId);
+        assertEquals(200, loginFormResponse.statusCode(), "Failed to open login form for realm " + realmName + " with response body " + loginFormResponse.body());
+        String formUrl = getLoginFormActionURL(loginFormResponse);
```
kami619, line 144: we are returning `null` in one of the cases for `getLoginFormActionURL`, should we handle that when we create this `formUrl` variable ?

kami619:
@mhajas apart from the query I had about the returned `null` value, the changes look good. Its working with the current Keycloak build which had the verify profile changes. Should we wait for the fixed build with reverted changes also to be available for tests and then merge this ?

mhajas:
> @mhajas apart from the query I had about the returned `null` value, the changes look good. Its working with the current Keycloak build which had the verify profile changes, I have verified in my local. Should we wait for the fixed build with reverted changes also to be available for tests and then merge this ?

@kami619 I updated the PR to check for `null` when obtaining the form action URL. I would not wait for the build with the reverted change. I tested this already and this PR passes without the commit (I tested with this branch: https://github.com/mhajas/keycloak/tree/revert-user-profile-commit).  Also, if I understand correctly the commit will be reintroduced later and these changes will be again necessary so it is good we are prepared for that.

kami619 (approved):
lgtm

PR #705, opened 2024-02-01
update ROSA Benchmark key results docs with new hashing changes

kami619:


kami619:
@ahus1 

I realized for us to be able to update the below we need to run this test manually as this is not part of the automated runs yet. I shall add a task in the background and see if I can run this from my local and update the text, hope thats enough.

```
* For each 350 refresh token requests per second, 1 vCPU per Pod in a three node cluster (tested with up to 435 refresh token requests per second).
```

ahus1:
> I realized for us to be able to update the below we need to run this test manually as this is not part of the automated runs yet. I shall add a task in the background and see if I can run this from my local and update the text, hope thats enough.

@kami619 - I'd be ok to leave those number for refresh tokens as is (given that the other numbers on client credential grants didn't change), and invest the time into automating refresh tokens in our nightly runs.

doc/benchmark/modules/ROOT/pages/report/rosa-benchmark-key-results.adoc
```diff
@@ -67,12 +67,12 @@ Observations:
...
 * Leave 1 GB extra head-room for spikes of RAM.
 
-* For each 100,000 active user sessions, add 500 MB per Pod in a three-node cluster (tested with up to 200,000 sessions).
+* For each 100,000 active user sessions, add 600 MB per Pod in a three-node cluster (tested with up to 200,000 sessions).
 +
 This assumes that each user connects to only one client.
 Memory requirements increase with the number of client sessions per user session (not tested yet).
 
-* For each 40 user logins per second, 1 vCPU per Pod in a three-node cluster (tested with up to 300 per second).
+* For each 40 user logins per second, 1.5 vCPU per Pod in a three-node cluster (tested with up to 300 per second).
```
ahus1, line 75: Please stick to the numbers as we calculate them in the run to avoid confusion. The base unit should be 1vCPU.
I would also round the number of 29 to 30, to have simpler numbers. This is an estimate only anyway.

```suggestion
* For each 30 user logins per second, 1 vCPU per Pod in a three-node cluster (tested with up to 300 per second).
```

To make the point of not mixing things: While the reduction from 40 to 30 logins per second per vCPU indicates a 25% drop of performance, you added 50% to the CPU usage, which is doubling the effect. 

Same for for the above 500 MB. I also see that the number is "116556" in the January 26 run. As I remember it fluctuates, we can keep it at 100000, or increase it to 110000 per 500 MB.

ahus1 (changes requested):
Hi @kami619 - thank you for the update. Please keep the base units 1 vCPU and 500 MB from the report. I used them this way as I expect people to figure out what their current settings of the Pod will be able to achieve.

doc/benchmark/modules/ROOT/pages/report/rosa-benchmark-key-results.adoc
kami619: makes sense, I had it initially this way and modified it to match with the limits we had for calculations. So based on this, should I modify "Calculation example" with the 30 logins ?

kami619:
@ahus1 I think it should be all good now.

ahus1 (approved):
LGTM, thank you!

ahus1:
> I can open a PR for keycloak docs as well

Thank you @kami619  - please go ahead and create a PR for the Keycloak main repository as well.

PR #706, opened 2024-02-01
Use new Infinispan config options

mhajas:
Please don't merge this PR yet because it doesn't work with the current nightly build.

Two things are necessary before merging:
1. This PR: https://github.com/keycloak/keycloak/pull/26682 or any equivalent fix needs to be merged in Keycloak main and it must be merged at the time when no-one works today as this will break deployments with current KC nightly. However, it needs to be merged before tomorrow morning because otherwise, tomorrow's scheduled run will fail
2. This PR https://github.com/keycloak/keycloak/pull/26666 needs to be merged


kami619 (approved):
lgtm

PR #710, opened 2024-02-07
Use `http-pool-max-threads` instead of `QUARKUS_THREAD_POOL_MAX_THREADS`

mhajas:
As per the recent changes in Keycloak https://github.com/keycloak/keycloak/issues/26455 we need to consume the same in the keycloak-benchmark deployments.

### DIFF
--- a/provision/minikube/keycloak/templates/keycloak.yaml
+++ b/provision/minikube/keycloak/templates/keycloak.yaml
@@ -67,7 +67,7 @@ spec:
     - name: storage-hotrod-password
       value: admin
 {{ end }}
-{{ if .Values.disableCaches }}
+{{- if .Values.disableCaches }}
     - name: cache-realm-enabled
       value: 'false'
     - name: cache-user-enabled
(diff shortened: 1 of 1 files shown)


provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -67,7 +67,7 @@ spec:
     - name: storage-hotrod-password
       value: admin
 {{ end }}
-{{ if .Values.disableCaches }}
+{{- if .Values.disableCaches }}
```
kami619, line 70: so was this never working before ? @mhajas

mhajas: It worked. 

I added `-` because there is no empty line in the resulting yaml. I added it because it didn't look good with the new line, see: https://www.keycloak.org/high-availability/deploy-keycloak-kubernetes there is an empty line below `spec:`

Search for `{{-` here: https://helm.sh/docs/chart_template_guide/control_structures/  for more details

kami619: thanks @mhajas !

kami619 (approved):
lgtm, thanks for fixing the ISPN conflict issue.

PR #711, opened 2024-02-08
Use the new remote-store options

pruivo:
Depends on https://github.com/keycloak/keycloak/pull/26756

I also changed the `provision/rosa-cross-dc/Taskfile.yaml` to deploy Keycloak in parallel in both ROSA clusters.


pruivo:
Custom Keycloak image available here: `quay.io/pruivo/keycloak:gh-25676`
Contains https://github.com/keycloak/keycloak/pull/26756 + https://github.com/infinispan/infinispan/pull/11942 (kind of)

mhajas:
We probably can't merge this until this: 
https://github.com/keycloak/keycloak-benchmark/blob/867dbd8a03410c3b2f10520e8b78fbb9cdd524ad/provision/minikube/keycloak/config/kcb-infinispan-cache-remote-store-config.xml#L56-L58

is needed

ahus1:
I know ... and we wouldn't merge this before KC24 is release as we would also need to update the documentation at the same time? 

I was planning to create a PR to add this to the Java code that it adds the workaround as well. But not today, only tomorrow.

pruivo:
This should be ready for review as well.

ahus1:
Tested with openshift / minikube / rosss cross dc setup, smoke tests work. Thanks!

PR #714, opened 2024-02-12
JGroups failure detection configuration

pruivo:
Needs to be moved to a proper place...

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
@@ -1,0 +1,181 @@
+= Understanding failure detection in Infinispan and JGroups
+:description: This guide describes how JGroups detects failed instances and how to configure it.
+
+{description}
+It explains the configuration attributes available for customization and its trade-off to achieve the desired detection rate.
+
+== Where JGroups is used, and why failure detection is important
+
+{project_name} uses Infinispan to store session related information in distributed caches both within {project_name}, and in external instances of Infinispan.
+{project_name} instances as well as external Infinispan instances form clusters, and the communication between the nodes of a cluster is handled by JGroups.
+The communication between different external Infinispan clusters in a multi-site setup is also handled by JGroups.
+
+Each node in such a cluster is serving session data, which is important for {project_name} to serve requests.
+An unresponsive node leads to blocked requests, and possibly timeouts and errors on the caller's side.
+If a node fails, requests need to be redirected to the remaining nodes, and its data needs to be re-distributed from backup nodes in the cluster.
+
+== Broken TCP connections causing communication failures
+
+A broken TCP connection has a performance impact on the cluster, and it is necessary to detect it, so the connection can be closed and new ones may be established.
+Note that TCP is a blocking/synchronous protocol, and if the connection is not working properly, threads will be blocked in the `Socket::write` operation.
+The kernel parameter `tcp_retries2` configures the amount failed retries until the Kernel force closes the connection.
+Quoting the https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt[Kernel Documentation]:
+
+[quote]
+--
+This value influences the timeout of an alive TCP connection, when RTO retransmissions remain unacknowledged.
+The default value of 15 yields a hypothetical timeout of 924.6 seconds and is a lower bound for the effective timeout.
+--
+
+During this period, the cluster performance is degraded.
+Changing Kernel parameters in an image or container is not desired, but JGroups has attributes that can be configured to detect and force-close these broken connections.
+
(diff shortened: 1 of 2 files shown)


doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
```diff
@@ -0,0 +1,103 @@
...
+:description: This guide describes how JGroups detects failed instances and its configuration.
+
+{description}
+
+== Audience
+
+This guide describe the failure detection mechanism that JGroups enforces to detect crashed pods inside the Infinispan cluster.
+It explains the configuration attributes available for customization and its trade-off to achieve the desired detection rate.
+
+== Broken TCP connections
```
pruivo, line 11: @belaban when you get a chance, can you take a look at this section

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
```diff
@@ -0,0 +1,103 @@
...
+However, it also increases the network usage due to the increase frequency of heartbeats.
+Note that all Infinispan pods in both cluster will send heartbeats to all the configured Gossip Router pods.
+
+TIP: The defaults should be work for 99% of the uses cases.
+
+For intra cluster communication in Infinispan or Keycloak cluster, broken TCP connection are detected and closed by the failure detection.
+See section <<fd>> for more information.
+
+[[fd]]
+== Failure Detection
```
pruivo, line 57: @belaban when you get a chance, can you take a look at this section

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
belaban: * "have a performance impact" -> "has a ..."
* "estabilished"
* "The timeouts are tuning..."?

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
belaban: * FD_ALL3: perhaps mention that doesn't send out heartbeats when regular messages have been sent

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
belaban: LGTM!

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
pruivo: We use mostly distributed caches (use unicast only) and heartbeats are suppressed only with multicast:

https://github.com/belaban/JGroups/blob/901197b381b6228eb3d2fde862972839812dfce6/src/org/jgroups/protocols/FailureDetection.java#L338

Is it worth mentioning?

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
pruivo: Thanks for the comments!

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
belaban: 'Multicast' in this respect means that we're sending a heartbeat to *everyone*, so in TCP this is an N-1 operation

ahus1:
@pruivo - I did some editing for style and grammar, and also added a paragraph here and there describing the "why" of failure detection, and the impacts the different failures have. 

I read the tip you're providing in the docs:

> The defaults should work for 99% of the use cases.

I agree that this is a level a regular admin should not touch in those setups we want to support. The installation we describe should come with good defaults for a low latency within the cluster, as well as <10 ms latencies between the data centers. With all operations being in-memory, the only erratic behavior would be a garbage collection during normal operations.

I am hesitant to put this level of detail to the main Keycloak documentation, to avoid people changing those defaults when they shouldn't.  Therefore I want to keep them within the KCB documentation site for now. We might then reference and refine it when we have questions from the community around this. 

So once you think the documentation is ready from your end, switch this PR to ready for review and I'll merge it after a final review. 

Something we might need for customer and the community is something similar, but with a different perspective: What should be metrics or log messages that around timeouts that I should watch out for, and what would be possible causes and remedies. Still, this might be a different page and too much for this page, and it might grow as people ask questions around that. 

Please let me know you think differently.

pruivo:
I've reviewed and squashed the commits. I'm fine keeping this outside the main Keycloak documentation. It should be available somewhere since it may help solution architects or support. 

Timeouts are usually a signal of trouble:

* Long GC pauses (miss configuration of the heap? memory leak?)
* Thread pool exhaustion (higher workload than expected?)
* Faulty network leading to dropped packages
* Pods crashing, requests will timeout until JGroups sends a new view change event
* Issues in the Keycloak side, because the `ClientListeners` are invoked in sync mode and with the lock acquired.

If we start thinking about it, we have material to write a book :smile:

PR #717, opened 2024-02-15
Add possibility to configure AWS JDBC Wrapper for Keycloak driver

### LINKED ISSUE #698: Use AWS JDBC Wrapper in Keycloak Benchmark
When deploying Keycloak with Aurora AWS, use the JDBC Wrapper provided by AWS to ensure that DB connections don't point to a read-only DB server after a failover.

mhajas:
Closes #698

This is WIP PR to use the AWS JDBC wrapper for. 

Checklist:
- [x]  Build image containing AWS JDBC wrapper jar
- [x] Test in CrossDC environment (I tested only in single OCP with ephemeral postgres and checked the logs)
- [x] Test with custom image build
- [x] Sort out the conflict between the image containing jar files in `/keycloak/providers` and mounting within helm chart (two options: make Keycloak look for jar files recursively (suggested by @ryanemerson in issue) or add all jar files when building image). I prefer 1st but it requires a change in Keycloak
- [x] Document new options
- [x] Optional: test database failover does not break Keycloak connections

I am testing this from `keycloak-tasks` directory with the following `.env` file:
```
KC_INSTANCES=1
KC_HOSTNAME_SUFFIX=<cluster-hostname-suffix>

ROSA_CLUSTER_NAME=gh-keycloak-a
KUBECONFIG=/path/to/kubeconfig
NAMESPACE=mhajas-keycloak

```

### DIFF
--- a/provision/minikube/keycloak/templates/keycloak.yaml
+++ b/provision/minikube/keycloak/templates/keycloak.yaml
@@ -27,7 +27,7 @@ spec:
 {{ if or (eq .Values.database "aurora-postgres") (eq .Values.database "postgres") (eq .Values.database "postgres+infinispan") }}
     vendor: postgres
 {{- if .Values.useAWSJDBCWrapper}}
-    url: jdbc:aws-wrapper:postgresql://postgres:5432/keycloak
+    url: jdbc:aws-wrapper:postgresql://{{ .Values.dbUrl }}:5432/keycloak
 {{- else }}
     url: jdbc:postgresql://postgres:5432/keycloak
 {{ end }}
--- a/provision/minikube/keycloak/templates/keycloak.yaml
+++ b/provision/minikube/keycloak/templates/keycloak.yaml
@@ -65,6 +65,10 @@ spec:
         key: {{ base .Values.infinispan.configFile }} # <1>
   # end::keycloak-ispn[]
 {{- end }}
+{{ if .Values.useAWSJDBCWrapper }}
+  transaction:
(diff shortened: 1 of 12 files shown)


ryanemerson:
I tried the PR on `gh-keycloak-a`. It's building the custom image with the AWS driver and deploying said image as expected :tada:

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -26,7 +26,11 @@ spec:
   db:
 {{ if or (eq .Values.database "aurora-postgres") (eq .Values.database "postgres") (eq .Values.database "postgres+infinispan") }}
     vendor: postgres
+{{- if .Values.useAWSJDBCWrapper}}
+    url: jdbc:aws-wrapper:postgresql://postgres:5432/keycloak
```
ahus1, line 30: I'm not sure if we should recommend using the external service to connect to Aurora. Instead, I'd prefer to use the direct DNS name of the database. This would prevent potential DNS obscurities due to indirection, and the JDBC wrapper might have direct access to multiple DNS entries. When doing this, `aurora-service.yaml` would no longer be needed.
 

```suggestion
    url: jdbc:aws-wrapper:postgresql://{{ .Values.dbUrl }}:5432/keycloak
```

ryanemerson (approved):

ryanemerson:
Other than Alex's comment about the JDBC url LGTM :slightly_smiling_face:

PR #718, opened 2024-02-16
Reduce Infinispan timeouts

pruivo:
* remote-timeout set to 5 seconds
* lock acquire-timeout set to 4 seconds
* cross-site timeout set to 4.5 seconds


ahus1:
I'd like to record the "why" so that my future self know why this has been changed. From our previous chat I recall: The retry mechanism in Keycloak have an incremental back-off parameterized starting with 10 milliseconds and 10 retried, which results in a maximum wait time of 10230 ms. 

I also recall that you added some debug statements when analyzing this. Feel free to create a PR to add those on the trace or debug level as well to help future investigations.  

Please add anything I have missed. I will merge it in a second.

PR #719, opened 2024-02-16
Keycloak derives memory settings from container limit

ahus1:
Relates to keycloak/keycloak#27078

The 300 MB no-heap usage are a preliminary value with manual tests on OpenShift. We'll need to watch the nightly tests to confirm that value. 
In my tests I've seen values up to 240 MB, and 300 MB is assumed to be on the safe side.


doc/benchmark/modules/ROOT/pages/report/rosa-benchmark-key-results.adoc
```diff
@@ -63,24 +62,25 @@ Summary:
...
-* Leave 1 GB extra head-room for spikes of RAM.
+* The base memory usage for an inactive Pod is 1000 MB of RAM.
 
 * For each 100,000 active user sessions, add 500 MB per Pod in a three-node cluster (tested with up to 200,000 sessions).
 +
 This assumes that each user connects to only one client.
 Memory requirements increase with the number of client sessions per user session (not tested yet).
 
+* In containers, Keycloak allocates 70% of the memory limit for heap based memory. It will also use approximately 300 MB of non-heap-based memory.
+To calculate the requested memory, use the calculated above. As limit, subtract the non-heap memory from the value above and divide the result by 0.7.
```
kami619, line 73:
```suggestion
To calculate the requested memory, use the calculation above. As limit, subtract the non-heap memory from the value above and divide the result by 0.7.
```

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -236,23 +233,6 @@ spec:
...
-# tag::keycloak[]
-            resources:
-              requests:
-                {{ if .Values.cpuRequests }}cpu: "{{ .Values.cpuRequests }}"{{end}}
-                {{ if .Values.memoryRequestsMB }}memory: "{{ .Values.memoryRequestsMB }}M"{{end}}
-              limits:
-                {{ if .Values.cpuLimits }}cpu: "{{ .Values.cpuLimits }}"{{end}}
-                {{ if .Values.memoryLimitsMB }}memory: "{{ .Values.memoryLimitsMB }}M"{{end}}
-            # end::keycloak[]
-            # readinessProbe:
```
kami619, line 248: do we want to leave the trace for probes for future use ?

kami619 (approved):
apart from the comments I made, this looks good to me.

provision/minikube/keycloak/templates/keycloak.yaml
ahus1: sure, we can keep that.

PR #732, opened 2024-03-06
first steps on upgrading gatling to the latest

kami619:
I started off taking the leap of faith by pushing the gatling to the latest version, as the earlier versions and scala dependencies have some CVE's listed.

I was able to fix one library refactoring, but now stuck with an issue around `java.nio.file.Path`'s `ancestor` method being phased out to fetch the root directory of the project, modifying that to `getRoot()` was returning more downstream issues. Would love to tag team with someone on this.

```
[INFO] --- scala-maven-plugin:4.8.1:compile (compile) @ keycloak-benchmark ---
[INFO] Compiler bridge file: /Users/kameswararaoakella/.sbt/1.0/zinc/org.scala-sbt/org.scala-sbt-compiler-bridge_2.13-1.8.0-bin_2.13.13__61.0-1.8.0_20221110T195421.jar
[INFO] compiling 29 Scala sources and 6 Java sources to /Users/kameswararaoakella/code/redhat/keycloak-benchmark/benchmark/target/classes ...
[WARNING] : -target is deprecated: Use -release instead to compile against the correct platform API.
[ERROR] /Users/kameswararaoakella/code/redhat/keycloak-benchmark/benchmark/src/main/scala/IDEPathHelper.scala:9: value ancestor is not a member of java.nio.file.Path
[WARNING] /Users/kameswararaoakella/code/redhat/keycloak-benchmark/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala:291: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method logout,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function. [quickfixable]
[WARNING] /Users/kameswararaoakella/code/redhat/keycloak-benchmark/benchmark/src/main/scala/keycloak/scenario/KeycloakScenarioBuilder.scala:310: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method logout,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function. [quickfixable]
[WARNING] three warnings found
[ERROR] one error found

```

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -7,12 +7,12 @@
     <version>0.12-SNAPSHOT</version>
 
     <properties>
-        <scala.version>2.12.10</scala.version>
-        <gatling.version>3.4.2</gatling.version>
+        <scala.version>2.13.13</scala.version>
+        <gatling.version>3.10.4</gatling.version>
         <janino.version>3.1.7</janino.version>
-        <scala-maven-plugin.version>3.2.2</scala-maven-plugin.version>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <scala-maven-plugin.version>4.8.1</scala-maven-plugin.version>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
         <encoding>UTF-8</encoding>
     </properties>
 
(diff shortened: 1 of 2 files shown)


kami619:
Couple of things to note:

- The target JVM would be JDK 17 with this upgrade. We need to make downstream changes wherever necessary.
- We would be using Scala 2.13 with this upgrade and a newer maven scala plugin.

ahus1 (approved):
Thank you for this upgrade. JDK17 is used for Keycloak as well, so we're good here.

PR #743, opened 2024-03-24
Fix after upstream change.

andyuk1986:


dataset/src/main/java/org/keycloak/benchmark/dataset/config/DatasetConfig.java
```diff
@@ -154,7 +152,7 @@ public class DatasetConfig {
     private Integer clientRolesPerUser;
 
     // Password policy with the number of password hash iterations. It is 27500 by default
```
kami619, line 154: I think comment would change as well to 210000.
```suggestion
    // Password policy with the number of password hash iterations. It is 210000 by default
```

PR #748, opened 2024-03-26
Use Hosted Control Planes for ROSA to speed up cluster creation

### LINKED ISSUE #673: Use Hosted Control Planes for ROSA to speed up cluster creation
### Description

Hosted Control Planes will speed up the creation of clusters.

### Discussion

_No response_

### Motivation

With hosted control planes, it takes only 10 minutes instead of 40 to create a ROSA cluster 

### Details

The Red Hat and OpenShift accounts are properly set up to work with the HCP options.

See https://docs.openshift.com/rosa/rosa_hcp/rosa-hcp-sts-creating-a-cluster-quickly.html for a guide

### LINKED ISSUE #750: GH action aurora-delete-database incorrectly sets environment variables for aurora_delete.sh
### Describe the bug

Workflows relying on the `aurora-delete-database` will fail as the action incorrectly sets the `AWS_REGION` env var when it should set `AURORA_REGION`

### Version

24

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ryanemerson:
I have removed the multi-az cluster options associated with creating rosa clusters as we haven't been using these and we don't have a need for this functionality, at least in the short term. If there are any objections I can attempt to reinstate these.

Resolves #673
Resolves #750

Remaining actions:
- [x] Test GH action integration for creating clusters
- [x] Provision crossdc setup across multiple HCP clusters

### DIFF
--- a/.github/actions/aurora-delete-database/action.yaml
+++ b/.github/actions/aurora-delete-database/action.yaml
@@ -18,4 +18,4 @@ runs:
       working-directory: provision/aws/rds
       env:
         AURORA_CLUSTER: ${{ inputs.name }}
-        AWS_REGION: ${{ inputs.region }}
+        AURORA_REGION: ${{ inputs.region }}


ryanemerson:
Multi-az clusters created and deployed as expected on inspection: https://github.com/ryanemerson/keycloak-benchmark/actions/runs/8452406289

.gitignore
```diff
@@ -96,3 +96,8 @@ quarkus/data/*.db
 # Horreum #
 ###########
 provision/environment_data.json
+
+# OpenTofu / Terraform
+**/*.tfstate*
+**/*.terraform*
+**/.terraform.lock*
```
ahus1, line 103: IMHO we should keep the `.terraform.lock.hcl` files in the Git repository, as it prevents accidental / unintended updates of providers. 

Once that file exists, only a `opentofu init -upgrade` would upgrade to newer versions as allowed by the constraints given in the configurations.

provision/aws/rosa_create_cluster.sh
```diff
@@ -5,96 +5,53 @@ if [ -f ./.env ]; then
...
+cd ${SCRIPT_DIR}/../opentofu/modules/rosa/hcp
+tofu init
+tofu workspace new ${CLUSTER_NAME} || true
+export TF_WORKSPACE=${CLUSTER_NAME}
+tofu apply -auto-approve \
+  -var cluster_name=${CLUSTER_NAME} \
+  -var instance_type=${COMPUTE_MACHINE_TYPE} \
+  -var openshift_version=${VERSION} \
+  -var region=${REGION} \
+  -var replicas=${REPLICAS}
```
ahus1, line 46: This script used to be idempotent if a following step failed, so we could re-run it via GitHub actions when for example installing of the logging failed. 

I tried to run the following, and even after setting `TF_VAR_token` get an error. 

Ideally it should continue where it stopped. If this is not trivial, at least re-adding the snippet from before would do most of the trick: 

```
if rosa describe cluster --cluster="${CLUSTER_NAME}"; then
  echo "Cluster ${CLUSTER_NAME} already exists."
else
#  ...
``` 

See below for the command I ran, and the output it provided. 

```
tofu workspace select gh-ryan-us-a
tofu apply -var cluster_name=gh-ryan-us-a -var instance_type=m5.xlarge -var openshift_version=4.14.5 -var region=us-east-2 -var replicas=2
``` 

Error received:

```
╷
│ Error: Attempt to index null value
│ 
│   on ../account-roles/main.tf line 7, in locals:
│    7:       policy_details       = data.rhcs_hcp_policies.all_policies.account_role_policies["sts_hcp_installer_permission_policy"]
│     ├────────────────
│     │ data.rhcs_hcp_policies.all_policies.account_role_policies is null
│ 
│ This value is null, so it does not have any indices.
╵
```

provision/opentofu/modules/rosa/hcp/scripts/rosa_verify_network.sh
```diff
@@ -0,0 +1,14 @@
...
+  set -x
+fi
+
+if [ -z "$CLUSTER_NAME" ]; then echo "Variable CLUSTER_NAME needs to be set."; exit 1; fi
+
+CLUSTER=$(rosa describe cluster -c ${CLUSTER_NAME} -o json)
+REGION=$(echo ${CLUSTER} | jq -r '.region.id')
+SUBNETS=$(echo ${CLUSTER} | jq -r '.aws.subnet_ids | join(",")')
+rosa verify network --cluster ${CLUSTER_NAME}
+rosa verify network --watch --status-only --region ${REGION} --subnet-ids ${SUBNETS}
```
ahus1, line 14: I see that this script spends 3 minutes, and it even spends 3 minutes when I run it now locally. 

To understand why this is necessary, it would be good to have a comment here in this script. Overall, I would hope one could skip it and safe us 3 minutes :-)

provision/opentofu/modules/rosa/hcp/variables.tf
```diff
@@ -0,0 +1,82 @@
...
+
+variable "instance_type" {
+  type    = string
+  default = "m5.4xlarge"
+}
+
+variable "replicas" {
+  type    = number
+  default = 2
+}
```
ahus1, line 82: Nitpick: As we now have a good place here for the defaults, I would tend to not keep them in other places and just maintain them here. 

So we would have fewer parameters on our workflows. It would also be ok to remove some and keep some from the list if you think we're about to use them now and then.

As this is a nitpick, feel free to keep for another future PR.

provision/opentofu/modules/rosa/README.md
```diff
@@ -0,0 +1,6 @@
+# ROSA OpenTofu Modules
+In order to use the modules in this directory, it's necessary for the var `token` to be set. This is an OpenShift Cluster
+Manager token, which can be accessed in the [Red Hat Hybrid Cloud Console](https://console.redhat.com/openshift/token).
+
+Instead of providing the token as a variable everytime `tofu` is called, it's possible to set the value via the
+`TF_VAR_token` environment variable.
```
ahus1, line 6: `TF_VAR_token` is such a non-descriptive name :-( I don't know if this can be changed. Again, a nitpick.

ahus1 (changes requested):

ahus1:
@ryanemerson - great job, happy to see the first working cluster created via OpenTofu!

provision/aws/rosa_create_cluster.sh
ryanemerson: > I tried to run the following, and even after setting `TF_VAR_token` get an error.

What error are you getting when running `rosa_create_cluster.sh` multiple times?

I previously executed this script multiple times once the cluster had been created (but the later steps failed) and it recognised the tofu state was as expected.

provision/opentofu/modules/rosa/hcp/scripts/rosa_verify_network.sh
ryanemerson: Yeah this is painful. For unknown reasons the "Inflight Check" fails for the egress when the HCP cluster is first created, however explicitly running the verification again always passes. I need to raise this with the Red Hat terraform team.

provision/aws/rosa_create_cluster.sh
ahus1: Ok, then it might have been my fault as I ran just the tofu command manually. Let's disregard it as you mention you tested that you were able to rerun it.

provision/opentofu/modules/rosa/hcp/scripts/rosa_verify_network.sh
ahus1: OK, then please add a comment to the script why it is there, and keep the rest for a follow-up.

provision/opentofu/modules/rosa/README.md
ryanemerson: Good point. I've changed the variable to `rhcs_token` so it becomes `TF_VAR_rhcs_token`.

provision/opentofu/modules/rosa/hcp/variables.tf
ryanemerson: I've removed the default values from the GH rosa-cluster-create workflow, but left the inputs so that the default can still be overridden. The `rosa_create_cluster.sh` script then adds the variables as `-var ` args to `tofu` if a value is specified.

ryanemerson:
I've addressed all comments and have executed a new GH run to verify: https://github.com/ryanemerson/keycloak-benchmark/actions/runs/8465601858

provision/opentofu/modules/rosa/hcp/scripts/rosa_verify_network.sh
ryanemerson: I asked the Terraform team and they say this is due to the VPC still being initialised when the cluster comes up. They may need to add an additional dependency check for the VPC on the `rosa_hcp_cluster` resource.

In the mean time they say it's possible to add a delay between the VPC and cluster creation. IMO the current solution works well enough and we should wait for this to be rectified in the Terraform provider.

ahus1 (approved):
Thank you for the change, it is good to see it will shave off 2x20 min for building our ROSA clusters.

I'm approving it today, and will merge it on Monday or Tuesday, so we'll not run into trouble over the long weekend.

mhajas (approved):
Nice work @ryanemerson! This is a huge step forward towards using OpenTofu! I expected creating ROSA clusters would be simpler in tofu though. 

Anyway, I think this is ready for merging. I added two comments but even if it will require some changes I am ok to do it later.

provision/aws/rosa_create_cluster.sh
```diff
@@ -1,100 +1,74 @@
...
 
-mkdir -p "logs/${CLUSTER_NAME}"
+echo "Installing ROSA cluster ${CLUSTER_NAME}"
 
-function custom_date() {
-    date '+%Y%m%d-%H%M%S'
-}
+cd ${SCRIPT_DIR}/../opentofu/modules/rosa/hcp
+tofu init
+tofu workspace new ${CLUSTER_NAME} || true
```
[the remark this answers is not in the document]
ryanemerson: I don't think it's possible for us to use a single workspace for two clusters as we're effectively using the same module but with different configurations in  order to create the two clusters. I think on executing apply for the second time with a different `cluster_name` opentofu will attempt to update the first cluster's resources.

This is all new to me as well though, so maybe I'm misunderstanding :slightly_smiling_face:

opentofu/remote-state/main.tf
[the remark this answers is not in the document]
ryanemerson: Good point, updated :+1:

provision/aws/rosa_create_cluster.sh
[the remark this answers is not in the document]
mhajas: It is new to me as well but using the same module more than once should be possible. 

For example, I used two modules here: https://github.com/mhajas/keycloak-benchmark/blob/opentofu-poc/provision/opentofu/main.tf#L7-L14

You can also specify aliases for providers and use more provider configurations within one tf file like this:
https://github.com/mhajas/keycloak-benchmark/blob/opentofu-poc/provision/opentofu/infinispan/main.tf#L7
and
https://github.com/mhajas/keycloak-benchmark/blob/opentofu-poc/provision/opentofu/infinispan/main.tf#L92

But as I said, it is ok for me to merge this as is and play with enhancements later.

mhajas:
I just realized one more thing. I probably know the answer as we are using the same script for choosing CIDR but I think it is worth double-checking. Can we run cluster creation in parallel with the new approach? https://github.com/keycloak/keycloak-benchmark/blob/main/.github/workflows/rosa-multi-az-cluster-create.yml#L61-L62

ahus1:
> Can we run cluster creation in parallel with the new approach?

Not yet, AFAIK, as we're still using the same script. Still it is a worthy next step. At the moment the script checks only the available IP address, but doesn't allocate it. Maybe we can find a mechanism to reserve the CIDR by setting some expiring information somewhere (maybe a S3 bucket entry with a timeout?) No idea what the most lightweight approach would be. cc: @ryanemerson

provision/aws/rosa_create_cluster.sh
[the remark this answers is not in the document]
ryanemerson: > For example, I used two modules here: https://github.com/mhajas/keycloak-benchmark/blob/opentofu-poc/provision/opentofu/main.tf#L7-L14

I hadn't considered referencing two modules from the root like that :slightly_smiling_face:

I think the downside to ^ approach is that we would have to make more changes to our various *.sh scripts and GH actions as we would either need to replace the use of `rosa_create_cluster.sh` or adapt it to support creating multiple clusters at once.

The flip side is that it would definitely make it easier to create multiple clusters in parallel, as the first stage of the module could be to determine CIDR ranges for both clusters before provisioning them.

ryanemerson:
I think the simplest way, i.e. no additional dependencies, for allowing parallel creation would be to utilise a terraform module that creates multiple clusters in a single workspace as suggested by @mhajas https://github.com/keycloak/keycloak-benchmark/pull/748#discussion_r1547461490.

Issue #636, opened 2023-12-13
Horreum: Include latencies from Infinispan Metrics


andyuk1986:
- Implement logic for retrieving the latencies information from Infinispan metrics.
- Update the schema in Horreum correspondingly.

andyuk1986:
PR with ISPN metrics inclusion is sent #749 . 
Separate PR will be sent with changes for resetting ISPN statistics over JMX.

andyuk1986:
I have updated the ispn metric implementation according to our's Monday discussion. Sent a PR #794

PR #749, opened 2024-03-27
Horreum: Include latencies from Infinispan Metrics

andyuk1986:
The PR includes the retrieval and inclusion for ISPN "vendor_rpc_manager_average_xsite_replication_time" metric. With this PR jmx is already enabled on ISPN servers, but the statistics is not yet reseted over JMX operation. That will be done with separate PR.

A successful run with generated json file where you can see ispn metrics can be found here: https://github.com/andyuk1986/keycloak-benchmark/actions/runs/8448435155

Relates to #636


ahus1:
Thank you for this PR!

PR #764, opened 2024-04-05
updates from @ahus1 for dashboard changes to support JVM debugging

kami619:


ahus1:
@kami619 - thank you for updating the diagrams, I was only providing hints. You did a lot more work for this PR than me.

PR #767, opened 2024-04-09
Upgrade to Openshift 4.15.5

ryanemerson:


ahus1 (approved):

ahus1:
Thank you for this housekeeping. I usually appreciate a short comment if this just housekeeping, or if we want to benefit from a a change. And I prefer a link to a GitHub action run with the new version.

PR #768, opened 2024-04-11
Support for installing Keycloak Operator via Operator Lifecycle Manager

tkyjovsk:
Added a new task `install-keycloak-operator-olm` to `provisioning/keycloak-tasks/Utils.yaml` as an alternative for the currently used task `install-keycloak-operator`.

Depending on a boolean parameter `KC_OPERATOR_OLM` either `install-keycloak-operator` or `install-keycloak-operator-olm` will be used in `provision/openshift/Taskfile.yaml`.

Additional parameters for the OLM task are:
- `KC_OPERATOR_CATALOG_SOURCE_NS` - defaults to `openshift-marketplace`
- `KC_OPERATOR_CATALOG_SOURCE` - defaults to `community-operators`
- `KC_OPERATOR_NAME` - defaults to `keycloak-operator`
- `KC_OPERATOR_CHANNEL` - defaults to the default channel of the selected product
- `KC_OPERATOR_VERSION` - defaults to the latest version in the selected channel

There is no documentation yet. I wanted to collect some feedback from the team before adding it.

### DIFF
--- a/doc/kubernetes/modules/ROOT/pages/openshift/installation-openshift.adoc
+++ b/doc/kubernetes/modules/ROOT/pages/openshift/installation-openshift.adoc
@@ -44,6 +44,30 @@ Cryostat depends on the Cryostat Operator to be installed, which is the default 
 
 To configure the deployment, see xref:customizing-deployment.adoc[] for details.
 
+=== Installing Keycloak from the Operator Hub
+
+By default the operator is installed directly via https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/refs/tags/nightly/kubernetes/kubernetes.yml[`keycloak-k8s-resources`].
+
+To install the operator from the _Operator Hub_ using the _Operator Lifecycle Manager_ set:
+[source,shell]
+----
+KC_OPERATOR_OLM=true
+----
+
+Additionally it is possible to customize the following parameters:
+
+KC_OPERATOR_CATALOG_SOURCE_NS:: Namespace of the OLM catalog source. Defaults to `openshift-marketplace`.
+KC_OPERATOR_CATALOG_SOURCE:: OLM catalog source. Defaults to `community-operators`.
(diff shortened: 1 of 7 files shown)


tkyjovsk:
Note that there is currently an issue with the Keycloak Operator (when installed from OLM) which prevents it from reading the custom Infinispan config map. https://github.com/keycloak/keycloak/issues/28638

Also note that the current keycloak Helm template is not backwards-compatible, i.e. it's not possible to install Keycloak 23 and lower because the CRDs are incompatible.

Additionally when installing the RHBK variant of the operator the `isup.sh` script will timeout when looking up `keycloak-operator` deployment (instead of `rhbk-operator`). I wanted to handle these issues separately from this PR.

tkyjovsk:
I have a workaround for [keycloak/keycloak#28638.](https://github.com/keycloak/keycloak/issues/28638). I can make it part of this PR or add it [separately](https://github.com/keycloak/keycloak-benchmark/commit/2d6e17e8869453e68e047feba310921441f5d3f3) after.

ahus1:
@tkyjovsk - thank you for this PR. 

> Also note that the current keycloak Helm template is not backwards-compatible

This is expected - the main branch is deploying only KC25. Branch `0.12-SNAPSHOT` is targeting KC24 - see GitHub releases for the details. As usual, PRs are first merged to main, and then eventually backported if we need them.

It is good that you found the bug when setting the config map for the ISPN configuration. Please add the workaround as you suggested in this PR as I understand it is essential for this to work and as it seems you have a workaround at hand already. 

It is ok to to have the other workaround as a separate PR if you don't have a workaround for it at hand.

tkyjovsk:
I added the documentation.

I included a workaround for the `isup.sh` script.

The workaround for https://github.com/keycloak/keycloak/issues/28638 is included in the `olm.sh` script. Ideally it should be separate but the procedure depends on some parameter autodetection in the `olm.sh` script which I didn't want to duplicate.

PR #774, opened 2024-04-16
Improve workaround for issue #28638.

tkyjovsk:
Use `kubectl patch` instead of the `kubectl get | jq | kubectl apply` approach.
Also patch the role definition in the operator's Cluster Service Version (CSV) in case OLM decides to recreate the role.


ahus1:
Thank you for this improvement. For future commits, please add a sign-off like the following as we also require it for the main repository: 

```
Signed-off-by: Author Name <authoremail@example.com>
```

PR #776, opened 2024-04-16
restart pods after dataset is created in rosa scaling benchmark for scheduled and on demand runs

kami619:
- I tried to do this with Keycloak CRD and annotating the restart time, but it was not working, hence fell back to using the reliable delete all pods method.

### DIFF
--- a/.github/workflows/rosa-scaling-benchmark.yml
+++ b/.github/workflows/rosa-scaling-benchmark.yml
@@ -193,6 +193,11 @@ jobs:
           ROSA_CLUSTER_NAME: ${{ inputs.clusterName || format('gh-{0}', github.repository_owner) }}
           NAMESPACE: ${{ env.PROJECT }}
 
+      - name: Restart Keycloak Pods after dataset creation
+        uses: ./.github/actions/keycloak-restart-pods
+        with:
+          project: ${{ env.PROJECT }}
+
       - name: Create AWS EC2 instances
         id: create_aws_ec2_instances
         uses: ./.github/actions/ec2-create-instances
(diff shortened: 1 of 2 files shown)


kami619:
it works : https://github.com/kami619/keycloak-benchmark/actions/runs/8709296473/job/23888618968#step:15:1

.github/workflows/rosa-scaling-benchmark.yml
```diff
@@ -193,6 +193,9 @@ jobs:
           ROSA_CLUSTER_NAME: ${{ inputs.clusterName || format('gh-{0}', github.repository_owner) }}
           NAMESPACE: ${{ env.PROJECT }}
 
+      - name: Restart Keycloak Pods after dataset creation
+        run: kubectl delete pods -n ${{ env.PROJECT }} -l app=keycloak
```
ahus1, line 197: I assume this works, as the creation of the EC2 instance takes longer than restarting the Keycloak instances. Still, for the sake of consistency, I'd like to wait using `kubectl wait` like we do it in the isup.sh script, or even use the isup.sh script itself if it would apply here.

kami619:
@ahus1 this should be good now for another review.

ahus1:
@kami619 - thank you for the update. Do you have an example run in your repository with this change that you can link to?

kami619:
> @kami619 - thank you for the update. Do you have an example run in your repository with this change that you can link to?

https://github.com/kami619/keycloak-benchmark/actions/runs/8739686884/job/23981556406#step:15:46

that annoying thing with github where you cannot run with the branch on your fork but have to backport the changes to main to get a GHA run.

PR #751, opened 2024-03-27
remove password-hash-iterations from dataset module config

kami619:
Fixes #744

PR #778, opened 2024-04-17
Added dataset options for setting password hashing algorithm and iterations.

tkyjovsk:
Added options `-g ALGORITHM` and `-i ITERATIONS` to the `dataset-import.sh` script for adjusting the realm password policy. Both are optional. The `DatasetResourceProvider` will only apply the options which were provided.

### DIFF
--- a/dataset/dataset-import.sh
+++ b/dataset/dataset-import.sh
@@ -25,7 +25,7 @@ set_environment_variables () {
   CREATE_TIMEOUT="3600"
   THREADS="-1"
 
-  while getopts ":a:r:n:c:u:e:o:i:p:l:t:C:T:" OPT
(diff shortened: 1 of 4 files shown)


kami619:
@tkyjovsk thanks a lot for adding this back in a proper way. I think we have to update the docs.

Ref., from last time when we modified them - https://github.com/keycloak/keycloak-benchmark/pull/751/files#diff-cdf53b413cfe9e2a53e09baeed0645f6744d883049fb1f458d6d8e4155921496L34

kami619 (changes requested):
we need to update the docs.

kami619:
@tkyjovsk I am going to add the docs and test it.

kami619 (approved):

PR #796, opened 2024-05-02
Improved workaround for issue 28638. Added waiting logic for operator pod.

tkyjovsk:
Improved workaround for issue keycloak/keycloak#28638

Added waiting loop for operator pod which was not yet created in some % of runs, exitting the OLM script with error.


ahus1:
@tkyjovsk - RHBK 24.0.3 has been released earlier this week which should include the fix for the original problem.

Therefore I assume that the workaround is no longer needed. If this is the case, could you please instead remove the workaround in this PR?

kami619:
@tkyjovsk can you please look into the suggestion from @ahus1 and take the appropriate next step ?

kami619:
closing this PR due to inactivity will resume the bits if the situation repeats itself.

PR #797, opened 2024-05-03
Use Graviton instances on AWS to save money using ARM

ahus1:
I've tested the changes for the EC2 instances. I didn't test it for the Aurora global database, as we're not using it at the moment. IMHO we would test this once we start using it again.


ryanemerson (approved):
LGTM. I agree, no need to verify the Aurora global deployment now.

kami619 (approved):

PR #798, opened 2024-05-03
persistent sessions load test added to the workflow

### LINKED ISSUE #785: enable persistent sessions in the provisioning and CI benchmarks
### Description

enable persistent sessions in the provisioning and CI benchmarks

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
Fixes #785 

- [x] Updated the Multi-AZ Create workflow to suit our current needs
- [x] Trimmed the ROSA Scaling Benchmark workflow to only perform dataset creation and benchmark and reporting
- [x] Updated the scheduled workflow to accommodate two deployments and benchmarks
- [x] Added filterable fields to Horreum report based on the deployment type (Additional work would be needed here to filter the reports in Horreum)
- [x] Local tests 

Note: this will allow the test results from persistent sessions to goto horreum now, but with some additional work @andyuk1986 confirmed we would be able to filter them out based on the deployment context we are updating the upcoming horreum report schema changes.

### DIFF
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -9,7 +9,7 @@ on:
 #   CLUSTER_PREFIX: gh-keycloak
 
 jobs:
-  cluster-deploy:
+  cluster-create-keycloak-deploy:
     name: ROSA Scheduled Create cluster
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
@@ -18,7 +18,7 @@ jobs:
     secrets: inherit
 
   run-functional-tests:
-    needs: cluster-deploy
+    needs: cluster-create-keycloak-deploy
     uses: ./.github/workflows/rosa-run-crossdc-func-tests.yml
     with:
       clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
@@ -31,9 +31,9 @@ jobs:
       clusterName: gh-keycloak-a # ${{ env.CLUSTER_PREFIX }}-a -- unfortunately 'env.' doesn't work here ${{ env.CLUSTER_PREFIX }}-a
     secrets: inherit
 
-  cluster-undeploy:
+  keycloak-undeploy:
     needs: run-scaling-benchmark
-    name: ROSA Scheduled Create cluster
+    name: Undeploy Keycloak deployment on the multi-az cluster
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-undeploy.yml
     with:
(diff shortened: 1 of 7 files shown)


kami619:
Test Runs:

- Deploy job : https://github.com/kami619/keycloak-benchmark/actions/runs/8940986385
- Benchmark job : https://github.com/kami619/keycloak-benchmark/actions/runs/8941255089

kami619:
@ahus1 sure thing, I meant `rows` when I said `writes` in my previous comment. Yup, its brining in lot more rows, would be interesting to see the load on CPU when this query runs.

kami619:
@andyuk1986 thanks for the review and providing with the fixes for the dependency issues that you uncovered.

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -29,6 +29,32 @@ jobs:
     uses: ./.github/workflows/rosa-scaling-benchmark.yml
     with:
       clusterName: gh-keycloak-a # ${{ env.CLUSTER_PREFIX }}-a -- unfortunately 'env.' doesn't work here ${{ env.CLUSTER_PREFIX }}-a
-      skipCreateDeployment: true
-      skipDeleteProject: true
+    secrets: inherit
+
+  cluster-undeploy:
```
mhajas, line 34: Should we call it `keycloak-undeploy`?

.github/workflows/rosa-scaling-benchmark.yml
```diff
@@ -346,25 +306,30 @@ jobs:
         with:
           createReportFile: false
 
+      - name: Generate timestamp for archive name
+        run: |
+          cur_epoch_time=$(date +%s)
+          echo "ARCHIVE_EPOCH=$cur_epoch_time" >> $GITHUB_ENV
```
mhajas, line 312: Would it makes sense to add some name parameter to the workflow name? Can be harder now to distinguish between non-persistent-sessions and persistent-session when there is only a different timestamp. WDYT?

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -29,6 +29,32 @@ jobs:
...
+    name: ROSA Scheduled Create cluster with Persistent Sessions
+    if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+    uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
+    with:
+      clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
+      enablePersistentSessions: true
+      createCluster: false
+    secrets: inherit
+
+  run-scaling-benchmark-with-peristent-sessions:
```
mhajas, line 55: Would it make sense to run functional tests before this step? It can be a follow-up as I am not sure the tests will pass for persistent-sessions.

mhajas (approved):
Thanks @kami619 , LGTM, I added a few comments but these are not blockers for merging so I am approving anyway.

.github/workflows/rosa-scaling-benchmark.yml
kami619: That was a question from Anna as well, but it would mean introducing a variable which we have to remove once we switch to persistent sessions only based runs. If it makes it easier, I can work on that change. lmk @mhajas

mhajas: I think it would be cleaner to have the name there. So for me it is +1 to change this but I will leave it to you

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
kami619: I can run it from my local and then enable it. good idea to run it after the persistent session based deploy.

andyuk1986 (approved):
@kami619 thanks a lot for all your changes. Looks good to me as well :)

.github/workflows/rosa-scaling-benchmark.yml
kami619: I will fix this in a subsequent PR along with any trailing issues.

PR #800, opened 2024-05-07
Use config value instead of console route value

mhajas:


ahus1:
LGTM, I ran the new command isolated on the CLI and it returns the same value.

PR #816, opened 2024-05-15
Ensure that rosa_delete_cluster.sh waits until completion before retu…

ryanemerson:


provision/aws/rosa_delete_cluster.sh
```diff
@@ -23,22 +23,23 @@ if [ -z "$REGION" ]; then echo "Variable REGION needs to be set."; exit 1; fi
...
-./../../../destroy.sh ${WORKSPACE} &
-DESTROY_PROC_ID=$!
+timeout -k 30m 30m bash -c "./../../../destroy.sh ${WORKSPACE}" &
+pids+=($!)
 
-timeout 30m bash -c 'wait $DESTROY_PROC_ID $LOGS_SAVING_PROC_ID'
-if [[ $? -eq 124 ]]; then
-    echo "Timeout occurred after 30 minutes."
-else
+if wait "${pids[@]}"; then
```
andyuk1986, line 41: @ryanemerson I have tried to run a simple script on my local using timeout / and `if wait`, as it is done here. So `wait "${pids[@]}"` returns false only when both processes exited with timeout. If one of them finished with timeout and other one normally exited, then the `wait` will return true and will print that everything is OK.

ryanemerson: Good spot @andyuk1986. I've updated it so that we only wait on the destroy process, as once that's completed the `rosa logs` command will terminate due to a 404 anyway.

andyuk1986: @ryanemerson thank a lot. Now it looks good.

andyuk1986 (approved):

andyuk1986:
Looks good to me. But I don't have access to merge the PR.

PR #818, opened 2024-05-16
Deploy Infinispan with maxCount=10000 and num_owners=1

### LINKED ISSUE #805: For persistent sessions, deploy Infinispan with maxCount memory setting in user/client sessions cache
### Description

Once persistent sessions are enabled, not all entries need to be in memory for Keycloak.

The same should be done for the external Infinispan, where also only a subset should be kept in memory.

### Discussion

https://github.com/keycloak/keycloak/pull/29347

### Motivation

Don't exhaust Infinispan's memory.

This is already in the docs. To make sure that it works this should be part of the scripts in the KCB project.

### Details

_No response_

ryanemerson:
Resolves #805

### DIFF
--- a/provision/infinispan/Utils.yaml
+++ b/provision/infinispan/Utils.yaml
@@ -104,6 +104,7 @@ tasks:
         --set image={{.CROSS_DC_IMAGE}}
         --set fd.interval={{.CROSS_DC_FD_INTERVAL}}
         --set fd.timeout={{.CROSS_DC_FD_TIMEOUT}}
+        {{if eq .KC_PERSISTENT_SESSIONS "true"}}--values ispn-helm/persistent-session-caches.yaml{{end}}
         ./ispn-helm
     preconditions:
       - test -f ".task/kubecfg/{{.ROSA_CLUSTER_NAME}}"
--- a/provision/infinispan/ispn-helm/persistent-session-caches.yaml
+++ b/provision/infinispan/ispn-helm/persistent-session-caches.yaml
@@ -1,0 +1,17 @@
+caches:
+  sessions:
+    owners: 1
+    memory:
+      maxCount: 10000
+  offlineSessions:
+    owners: 1
+    memory:
+      maxCount: 10000
+  clientSessions:
+    owners: 1
+    memory:
+      maxCount: 10000
+  offlineClientSessions:
+    owners: 1
+    memory:
+      maxCount: 10000
--- a/provision/infinispan/ispn-helm/templates/infinispan.yaml
+++ b/provision/infinispan/ispn-helm/templates/infinispan.yaml
@@ -143,6 +143,10 @@ spec:
       remoteTimeout: {{ $config.remoteTimeout | default $.Values.cacheDefaults.remoteTimeout }}
       locking:
         acquireTimeout: {{ $config.lockTimeout | default $.Values.cacheDefaults.lockTimeout }}
+      {{- if and $config.memory $config.memory.maxCount }}
+      memory:
+        maxCount: {{ $config.memory.maxCount }}
+      {{- end }}
       stateTransfer:
         chunkSize: 16
       {{ if $.Values.crossdc.enabled }}


doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
```diff
@@ -166,7 +168,8 @@ Change the cache configuration as show and add the attributes (in bold) you want
 
 [source,xml,subs=+quotes]
 ----
-<distributed-cache name="sessions" owners="2" statistics="true" *remote-timeout="15000"*> <!--1-->
+<distributed-cache name="sessions" owners="1" statistics="true" *remote-timeout="15000"*> <!--1-->
+    <memory max-count="10000" />
```
ahus1, line 172: IMHO the max-count needs a call-out describing for which caches (sessions, clientSessions, offlineSessions, offlineClientSessions) and in what scenario (persistent sessions) this should be configured.

This would also apply for the for the YAML configuration above.

provision/minikube/keycloak/config/kcb-infinispan-cache-config.xml
```diff
@@ -37,19 +37,23 @@
             </encoding>
             <memory max-count="10000"/>
         </local-cache>
-        <distributed-cache name="sessions" owners="2">
```
ahus1, line 40: The changes about owners and max-count should be active only when persistent sessions are enabled. At the moment Keycloak does those settings automatically in the `CacheManagerFactory`. Although this would prints an INFO message, I'd rather have the message in the log than maintaining two different configurations. So I'd ask you to revert the changes in this file.

ahus1 (dismissed):
Hi @ryanemerson - see below for my comments. One overall requirement I'd see: Those changes should only apply when persistent sessions are enabled. 

Unfortunately a remedy to the bug we're handling in https://github.com/keycloak/keycloak/issues/29592 might be not to apply this change in the external Infinispan. 

With that recent development, I'd suggest to put this issue on hold until the other issue resolved. We would then revisit it again once the remote-store only changes from #28745 are in. But even then I'm not sure how an cache with maxCount and amnd xsite setup can maintain entries that are not outdated, as a remove in the primary site for a key that exists only in the secondary site will never reach the secondary site. 

cc: @pruivo @mhajas

ahus1:
@ryanemerson - update to my previous comment: The PR https://github.com/keycloak/keycloak/pull/29597 might still allow for an external Infinispan with maxCount, let's continue the discussion next week.

mhajas:
@ryanemerson  Is ISPN operator able to update existing caches with `maxCount` and `owners`? I remember that some time ago when I tried to change something in CR Operator refused to change it.

ryanemerson:
> @ryanemerson Is ISPN operator able to update existing caches with `maxCount` and `owners`? I remember that some time ago when I tried to change something in CR Operator refused to change it.

Yes this is possible when the default strategy of `spec.updates.strategy: recreate` is used on the Cache CR. The Operator will attempt to update the cache configuration on the server to equal that of the CR, however the operation will fail if the old and new configuration are incompatible for a runtime update.

ahus1:
@ryanemerson - the issues to handle the expiry of sessions is now resolved as the other PR is merged. 

Could you please now attend to the other issue: Your changes should only apply when persistent sessions are enabled, as limiting the number of entries is otherwise not supported. 

Thanks!

doc/kubernetes/modules/ROOT/pages/running/timeout_tunning.adoc
ahus1: agreed, let's leave it out here.

ryanemerson:
> Your changes should only apply when persistent sessions are enabled, as limiting the number of entries is otherwise not supported.

It should be good now. If `KC_PERSISTENT_SESSIONS` is true when provisioning the Infinispan cluster then cache configs are updated accordingly.

pruivo (approved):

kami619:
merging it, as its approved by @pruivo

PR #822, opened 2024-05-20
Only remove all Route53 subdomains if REAPER env variable set

ryanemerson:


mhajas (approved):
LGTM thanks for the quick fix @ryanemerson!

PR #823, opened 2024-05-20
Test delete entry in remote Infinispan and check removals are replica…

mhajas:
…ted to the second site anyway

- [x] Test logouts
- [ ] Test refreshes

### DIFF
--- a/provision/rosa-cross-dc/keycloak-benchmark-crossdc-tests/src/test/java/org/keycloak/benchmark/crossdc/LoginLogoutTest.java
+++ b/provision/rosa-cross-dc/keycloak-benchmark-crossdc-tests/src/test/java/org/keycloak/benchmark/crossdc/LoginLogoutTest.java
@@ -1,5 +1,6 @@
 package org.keycloak.benchmark.crossdc;
 
+import org.jboss.logging.Logger;
 import org.junit.jupiter.api.Test;
 import org.keycloak.benchmark.crossdc.util.InfinispanUtils;
 
@@ -16,6 +17,9 @@ import static org.keycloak.benchmark.crossdc.util.InfinispanUtils.SESSIONS;
 
 
 public class LoginLogoutTest extends AbstractCrossDCTest {
+
+    protected static final Logger LOG = Logger.getLogger(LoginLogoutTest.class);
+
     @Test
     public void loginLogoutTest() throws URISyntaxException, IOException, InterruptedException {
         //Login and exchange code in DC1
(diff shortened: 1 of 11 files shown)


ahus1:
@mhajas - I tested this a bit today, and it works nicely for the logouts. 

I verified this also with my local setup where I tested last week: The remove event is propagated to the client for a remote Infinispan 15.0.3, even if the key doesn't exist any more. 

When using 14.0.27 which I did locally, I found that the remove events was propagated to the caller only when the entries existed. So it seems that there is a change in behavior between the major versions. Good that we now have a test for this. 

So I assume we want people to use ISPN 15.x when running with KC 25? 

cc: @ryanemerson @pruivo

ryanemerson:
> So I assume we want people to use ISPN 15.x when running with KC 25?

100%

Let's keep the support matrix as small as possible. This will allow us to spend less time on verifying different combinations and more time enhancing the Keycloak experience for users with new improvements/features.

ahus1:
@ryanemerson / @pruivo - I agree with your comments. Could please also think in the direction if the change in behavior could be seen as a regression by others, and that it would be fixed (to the disadvantage of Keycloak) in an upcoming ISPN release?

ryanemerson:
@ahus1 I believe the change in behaviour was introduced by [ISPN-15211](https://issues.redhat.com/browse/ISPN-15211) and is documented in our upgrading guide.

provision/rosa-cross-dc/keycloak-benchmark-crossdc-tests/src/test/java/org/keycloak/benchmark/crossdc/LoginLogoutTest.java
```diff
@@ -80,4 +86,113 @@ public void testRefreshTokenRevocation() throws Exception {
...
+        Map<String, Object> tokensMap = LOAD_BALANCER_KEYCLOAK.passwordGrant(REALM_NAME, CLIENTID, USERNAME, MAIN_PASSWORD);
+
+        // Make sure all ISPNs can see the entry in the cache
+        assertTrue(DC_1.kc().embeddedIspn().cache(SESSIONS).contains((String) tokensMap.get("session_state")));
+        assertTrue(DC_1.ispn().cache(SESSIONS).contains((String) tokensMap.get("session_state")));
+        assertTrue(DC_2.kc().embeddedIspn().cache(SESSIONS).contains((String) tokensMap.get("session_state")));
+        assertTrue(DC_2.ispn().cache(SESSIONS).contains((String) tokensMap.get("session_state")));
+
+        // Remove the session from the remote store in DC1 only
+        try (var close = InfinispanUtils.withBackupDisabled(DC_1.ispn().cache(SESSIONS), DC_2.ispn().siteName())) {
```
ryanemerson, line 102: `var ignored = ` will stop Intellij complaining about the var being unused.

ahus1:
@ryanemerson - thank you for the note that this is an expected behavior. 

Given that we now depend on ISPN 15 specific behavior, this should be reflected in the docs and the migration guide. I created an issue for that: https://github.com/keycloak/keycloak/issues/29750

Could you please pick that one up and create a [minimal] docs PR for this? Thanks!

mhajas:
As agreed we will not provide a test for updates currently so this PR should be ready for reviewing and merging. @ryanemerson @ahus1 Could you please re-review?

ryanemerson (approved):

ahus1:
Thank you for these tests, they really helped us to understand Infinispan's behavior around this!

PR #824, opened 2024-05-21
Upgrade to Infinispan 15.0.4.Final

ryanemerson:


ryanemerson:
The current operator release isn't aware of the 15.0.4.Final Operand, which is why the image has a different version to the specified image. It should be safe to upgrade to a custom image in this way, as there has been no changes to the server that should impact the Operator functionality.

I created this PR now, instead of waiting for the Operator release which might be another week or so yet, as 15.0.4.Final is required by https://github.com/keycloak/keycloak-benchmark/pull/819

If we would rather wait for the Operator release, then I'm happy to close this PR.

ahus1:
@ryanemerson - thank you for adding this context why you chose to change one file and not the other. Works for me!

PR #828, opened 2024-05-22
Exclude the Highcharts dependency and provide a minimal replacement

### LINKED ISSUE #809: Resolve licensing issues for Gatling Highcharts
### Description

As Keycloak seeks to complete the incubating onboarding items, it must align with the CNCF license policy, which mandates board approval for non-Apache 2.0 dependencies. 

Gatling, which incorporates code from Akka that adopted the "source available" Business Source License (BSL) v1.1 a few years ago, raises licensing concerns. Consequently, we must transition to Hyperfoil, which is governed by the ASL 2 license, in our upcoming releases. Additionally, we should remove Gatling from the current binary distribution if required to maintain compliance until the integration of Hyperfoil is finalized.

### Acceptance Criteria
1. Confirm that replacing Gatling with Hyperfoil resolves any existing licensing issues.
2. Evaluate the technical implications and compatibility of integrating Hyperfoil.
3. If required for compliance, temporarily stop the distribution of Gatling in current binary releases.
5. Inform users about the change and any temporary measures taken.

### Discussion

_No response_

### Motivation

Adhere to the CNCF license policies.

### Details

- https://issues.redhat.com/browse/RHBK-1406

ahus1:
Closes #809


ahus1:
This is an example out, all charts will be gone: 

overview

This minimal replacement will continue to provide a `js/stats.json` file which we can use for our reports. 

See the ZIP file of a small example run here: [result.zip](https://github.com/keycloak/keycloak-benchmark/files/15404495/result.zip)

ahus1:
@kami619 - I managed to remove the problematic library. The output is not so nice, but sufficient for our automatic nightly runs. I assume we would then need to rely more on the Grafana dashboards :-/

benchmark/src/main/resources/META-INF/services/io.gatling.charts.component.ComponentLibrary
```diff
@@ -0,0 +1 @@
+org.keycloak.benchmark.gatling.chart.ComponentLibraryDummy
```
ryanemerson, line 1: Instead of providing dummy implementations can we just add the `io.gatling:gatling-charts` dependency? It's released under the Apache 2.0 license in the main Gatling repository https://github.com/gatling/gatling/tree/main/gatling-charts

ahus1: Despite a naming that might suggest otherwise, it is IMHO as follows: 

* `gatling-charts` provides the base charting functionality, and is still included in the PR
* `gatling-charts` depends on an implementation of the `ComponentLibrary` which is then either the original Highcharts one, or my dummy implementation

kami619:
the changes look good to me, and thanks Alex for providing a functional replacement to high charts. I think we could live without the graphs, in the end even in the html charts, I was mostly following the numbers, which are still present.

kami619 (approved):

kami619:
@ryanemerson feel free to review this further and merge it if you are satisfied as well.

ryanemerson:
Thanks @ahus1

PR #833, opened 2024-05-28
[Horreum] Infinispan metric calculation sometimes results to division to 0 #832

### LINKED ISSUE #832: [Horreum] Infinispan metric calculation sometimes results to division to 0
### Describe the bug

Sometimes it happens that for specific test run the ISPN metric calculation ends with division to 0. Need to investigate and fix.

### Version

main

### Expected behavior

The ispn metrics should be calculated properly and stored in json report file.

### Actual behavior

The issue with division to 0 end with missing data in json file.

### How to Reproduce?

Run the rosa scaling benchmark workflow.

### Anything else?

_No response_

andyuk1986:
Fixes #832

### DIFF
--- a/.github/actions/prometheus-metrics-calc/action.yml
+++ b/.github/actions/prometheus-metrics-calc/action.yml
@@ -122,6 +122,7 @@ runs:
         mv tmp.json ${HORREUM_OUTPUT_FILE_NAME}
         #Removing ispn metrics file for next execution
         rm "${{ inputs.ispnCacheName}}_ispn_metrics_file_count"
+        rm "${{ inputs.ispnCacheName}}_ispn_metrics_file_sum"
 
     - id: gatling-report-parse-and-store
       name: Parses gatling report and stores got information in JSON file.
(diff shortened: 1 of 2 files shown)


andyuk1986:
You can find a successful run on my fork here: https://github.com/andyuk1986/keycloak-benchmark/actions/runs/9267087730

andyuk1986:
So, I am afraid that the generated report data so far was only correct for memoryUsageTest, for other 2 test runs (cpuUsage and cpuUsageForGrants was incorrectly calculated).

ahus1:
@andyuk1986 - thank you for figuring out the problem!

To help me understand the root cause: Can you please describe it to me? Looking at the diff, the `rm` seems to do the trick?

andyuk1986:
> @andyuk1986 - thank you for figuring out the problem!
> 
> To help me understand the root cause: Can you please describe it to me? Looking at the diff, the `rm` seems to do the trick?
@ahus1 actually there were 2 problems: 
1. The floating point math operation in bash sometimes were throwing error which was causing that data was completely missing for specific test.
2. The 'rm' part was the reason of devision to 0 issue. When the statistic is read from prometheus the data is stored in the file. This is done before and after the benchmark ran and is written on new line. Then the metric calculator reads the data and makes math operations to calculate that and theoretically should remove the file for next bechmark iteration as the next benchmark uses the same file name. I was deleting one of the files and missed the second, as for ispn metric calculation I am picking 2 time sub-metrics (seconds sum and count) and store them in different files for later processing. So as one of the files was not deleted, the later benchmarks were adding the new statistics to already existing file where already data was from previous benchmark run, and after that the whole calculation was going wrong.

ahus1:
Thank you for the explanation and the fix!

Issue #789, opened 2024-04-29
Simplify HCP deployment with RH Terraform modules


ryanemerson:
### Description

When we migrated to ROSA HCP clusters it was necessary to include many Terraform modules in our repository to provision HCP prerequisite resources such as operator roles, VPCs etc. Red Hat now provides dedicated modules for this purpose https://github.com/terraform-redhat/terraform-rhcs-rosa-hcp

We should simplify our current scripts by incorporating ^ as appropriate.

ryanemerson:
The first stable release [v1.6.2](https://github.com/terraform-redhat/terraform-rhcs-rosa-hcp/releases/tag/v1.6.2) is now available.

PR #842, opened 2024-06-05
 Unable to remove ROSA cluster due to osd-network-verifier SecurityGroup

### LINKED ISSUE #841: Unable to remove ROSA cluster due to osd-network-verifier SecurityGroup
### Describe the bug

The nightly cluster reaper is occasionally failing due to the presence of a "osd-network-verifier" SecurityGroup being associated with some ROSA clusters. This is created by ROSA itself as part of the network verification step and AFAIK can't be prevented, however it may not be necessary once https://github.com/keycloak/keycloak-benchmark/issues/789 has been fixed.

A simple workaround for now is to explicitly delete the SecurityGroup as part of the `rosa_delete_cluster.sh` script if present.

### Version

N/A

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ryanemerson:
Closes #841

### DIFF
--- a/provision/aws/rosa_delete_cluster.sh
+++ b/provision/aws/rosa_delete_cluster.sh
@@ -23,6 +23,20 @@ if [ -z "$REGION" ]; then echo "Variable REGION needs to be set."; exit 1; fi
 ./rds/aurora_delete_peering_connection.sh || true
 ./rosa_efs_delete.sh || true
 
+# Explicitly delete OSD Network Verifier that's sometimes created as it prevents VPC being deleted
+OSD_VERIFIER_SG=$(aws ec2 describe-security-groups \
+  --filters Name=tag:Name,Values=osd-network-verifier \
+  --filters Name=tag:cluster,Values=${CLUSTER_NAME} \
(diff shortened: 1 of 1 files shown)


mhajas (approved):
@ryanemerson LGTM, however, is there any issue reported in ROSA for this? Sounds like something they should have a look.

ahus1:
+1 for Michal's comment. Ryan, Feel free to merge when you think this PR is ready.

ryanemerson:
I'm not sure if it's ROSA specifically doing something wrong, or if it's due to us explicitly calling [rosa verify network](https://github.com/keycloak/keycloak-benchmark/blob/main/provision/opentofu/modules/rosa/hcp/main.tf#L97-L107). The explicit verification was added, as without it we were getting errors saying that the initial "Inflight check failed" for the cluster. When I discussed the need for this with the terraform team, they said it was due to a timing issue between the VPC creation and the ROSA provisioning. The official Red Hat HCP modules now have a stable release as of last week, so hopefully #789 will resolve the underlying issue.

PR #846, opened 2024-06-12
Fix Active/Passive functional tests

ryanemerson:


mhajas (approved):
Thanks for the fix @ryanemerson

PR #852, opened 2024-06-14
initial docs for JVM tuning and observability

### LINKED ISSUE #837: Document relevant metrics and JVM options that we used in determining the Keycloak JVM heap behavior
### Description

Document relevant metrics and JVM options that we used in determining the Keycloak JVM heap behavior. The objective is to standardize the way we think and understand about the JVM and GC characteristics of Keycloak application across different platforms (standalone, container etc.,)

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
fixes #837


kami619:
> Thank you for this PR, the contents look good to me. Still, as it is now, those new pages will not show in any navigation - please add them to the appropriate navigation.

good catch @ahus1, oversight on my end. It should be addressed in my latest commit now.

ahus1 (approved):
Thank you for this PR. I added also the missing entries to the `nav.adoc`, so they'll appear as an expandable menu on the left side of the screen.

PR #853, opened 2024-06-14
Enable protostream

### LINKED ISSUE #628: CrossDCTest: Unable to load keys for caches that use UUID as key

pruivo:
* Enables protostream encoding in the external Infinispan
* Changes the testsuite to use the Hot Rod client

Closes #628

I have the following error with status code 500. I haven't figured out where they come from. I couldn't find any exceptions in the pod's log. 

```
[ERROR]   EntityReplicationTest.keycloakEntityReplicationOverCacheTest:25 » InternalServerError HTTP 500 Internal Server Error
[ERROR]   SessionExpirationTest.sessionExpirationTest:19 » InternalServerError HTTP 500 Internal Server Error
```

We probably need to fork the `main` branch before merging this one. The forked branch will keep compatibility with jboss-marshalling and KC25.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,7 @@
 
   <properties>
     <keycloak.version>999.0.0-SNAPSHOT</keycloak.version>
+    <infinispan.version>15.0.4.Final</infinispan.version>
     <junit5.version>5.10.1</junit5.version>
     <httpclient.version>4.5.14</httpclient.version>
     <maven.enforcer.plugin.version>3.4.1</maven.enforcer.plugin.version>
@@ -86,6 +87,12 @@
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpclient</artifactId>
         <version>${httpclient.version}</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-client-rest</artifactId>
+        <version>${infinispan.version}</version>
         <scope>test</scope>
       </dependency>
     </dependencies>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -11,6 +11,7 @@
 
   <properties>
     <keycloak.version>999.0.0-SNAPSHOT</keycloak.version>
+    <infinispan.version>15.0.4.Final</infinispan.version>
```
mhajas, line 14: The deployment is currently using 15.0.x Infinispan image, I believe this can break some API compatibility. Is there some nightly maven release?

mhajas (commented):
Thank @pruivo, this PR looks good to me. 

To the errors you see. What were you testing this against? If you tested against `runner-keycloak` namespace created by the nightly run, it is probably caused by the fact the testsuite currently does not work with A/A. We need to test this against A/P deployment where the testsuite is passing.

Here are the errors: https://github.com/keycloak/keycloak-benchmark/actions/runs/9559243516/job/26354830531

I will be looking at the A/A failures now.

pom.xml
pruivo: That version is only used by the REST client since the Hot Rod client will use the version shipped with Keycloak. It is fine and only the major version matters.

pruivo:
> Thank @pruivo, this PR looks good to me.
> 
> To the errors you see. What were you testing this against? If you tested against `runner-keycloak` namespace created by the nightly run, it is probably caused by the fact the testsuite currently does not work with A/A. We need to test this against A/P deployment where the testsuite is passing.
> 
> Here are the errors: https://github.com/keycloak/keycloak-benchmark/actions/runs/9559243516/job/26354830531
> 
> I will be looking at the A/A failures now.

I created a new cluster in a different namespace (keycloak + infinispan) with Route53. I was using the existing aurora DB instead of creating and populating another one. I assume it wouldn't have any impact.

mhajas:
I suspect it can be caused by the database. It seems A/A tests leave the database in a state where some requests end up with 500. Can you please test with a different database cluster?

mhajas (approved):
Thank you @pruivo!

PR #875, opened 2024-07-01
Testsuite for External Infinispan Feature

### LINKED ISSUE #874: Testsuite for External Infinispan Feature
Run the test suite with active/active and active/passive with the External Infinispan feature enabled.

pruivo:
Add option to skip the embedded cache assertions and update the GH jobs to run the test suite

Fixes #874

### DIFF
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -93,6 +93,15 @@ jobs:
       outputArchiveSuffix: 'external-infinispan'
     secrets: inherit
 
+  run-ap-functional-tests-with-external-infinispan:
+    needs: cluster-create-keycloak-deploy
+    name: Runs the functional test suite in Active/Passive with the External Infinispan feature
+    uses: ./.github/workflows/rosa-run-crossdc-func-tests.yml
+    with:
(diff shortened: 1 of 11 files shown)


ahus1:
@pruivo - I've updated the GitHub action a bit for the added step as it seems to me we always run the functional test first, and then the load test. I also updated the `needs` to match the previous step.

Do you have a run in your GH repo that show a green outcome of the scripts?

pruivo:
@ahus1 if I enable the scripts in my fork, wouldn't that conflict with the nightly job from this repo? I didn't run the GH scripts, just the tests locally.

ahus1:
@pruivo - The main question is what I need to test, and what has already been tested. You can enable actions in your repo, and the nightly job in your repo shouldn't conflict with the one in the main repo. 

Unfortunately you can't pass a cluster prefix to a manually scheduled run as of today. 

Ok, I'll continue testing. BTW, the DCO bot complains that a signoff fails. I'll ignore it here for now.

pruivo:
It is just the GH actions yaml files that aren't tested. I was hope to use the nightly for it :P

ahus1 (approved):
Thank you for this PR. I ran a test here for the changes that were done, see https://github.com/ahus1/keycloak-benchmark/actions/runs/9760760106/job/26946547537

During the review, I ensured that the `needs` always points to the previous step. I also found out that new parameters are needed for `workflow_call` as well if it is called from another workflow.

pruivo:
thanks @ahus1 !

PR #907, opened 2024-07-27
optimize rosa scheduled workflow to run reduced combinations

### LINKED ISSUE #900: Make the daily test run on A/A + persistent sessions like we plan this for KC26

kami619:
Fixes #900 

Trim the ROSA Scheduled workflow to only two combinations

- ROSA Scheduled Create Active/Active cluster with Volatile Sessions and External Infinispan
- ROSA Scheduled Create Active/Active cluster with Persistent Sessions and External Infinispan

Working run: https://github.com/kami619/keycloak-benchmark/actions/runs/10123478369

### DIFF
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -36,7 +36,6 @@ jobs:
     with:
       clusterName: gh-keycloak-a # ${{ env.CLUSTER_PREFIX }}-a -- unfortunately 'env.' doesn't work here
       outputArchiveSuffix: 'active-active-external-infinispan'
-      skipCreateDataset: true
     secrets: inherit
 
   keycloak-undeploy-active-active-with-external-infinispan:
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -9,111 +9,49 @@ on:
 #   CLUSTER_PREFIX: gh-keycloak
 
 jobs:
-  cluster-create-keycloak-deploy:
-    name: ROSA Scheduled Create cluster A/P volatile sessions
+  keycloak-deploy-active-active-with-external-infinispan:
+    name: ROSA Scheduled Create Active/Active cluster with External Infinispan
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
     with:
       clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
-    secrets: inherit
-
-  run-functional-tests:
-    needs: cluster-create-keycloak-deploy
-    uses: ./.github/workflows/rosa-run-crossdc-func-tests.yml
-    with:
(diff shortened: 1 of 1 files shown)


mhajas (approved):
Thank you @kami619, I pushed one more commit, otherwise LGTM

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -138,62 +75,18 @@ jobs:
...
-    with:
-      clusterName: gh-keycloak-a # ${{ env.CLUSTER_PREFIX }}-a -- unfortunately 'env.' doesn't work here ${{ env.CLUSTER_PREFIX }}-a
-      outputArchiveSuffix: 'active-active-external-infinispan'
-      skipCreateDataset: true
-    secrets: inherit
-
-#  keycloak-undeploy-active-active-with-external-infinispan:
-#    needs: run-functional-tests-active-active-with-external-infinispan
-#    name: Undeploy Keycloak deployment on the multi-az cluster
+#  keycloak-undeploy-active-active:
```
ryanemerson, line 83: Is there any reason to keep this commented code around, or should we just remove as we can retrieve from git if required?

ryanemerson:
Open question to discuss in the meeting today. Should we remove the old A/P code to reduce the things we need to maintains as it's still available in the `0.13` branch if required?

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -9,111 +9,48 @@ on:
...
       skipAuroraDeletion: true
+      activeActive: true
     secrets: inherit
 
   keycloak-deploy-active-active:
-    needs: keycloak-undeploy-with-external-infinispan
+    needs: keycloak-undeploy-active-active-with-external-infinispan
     name: ROSA Scheduled Create Active/Active cluster with Persistent Sessions
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
```
pruivo, line 56: isn't `enableExternalInfinispanFeature: true` missing from this config?

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -9,111 +9,48 @@ on:
...
     name: Undeploy Keycloak deployment on the multi-az cluster
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-undeploy.yml
     with:
       clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
       skipAuroraDeletion: true
+      activeActive: true
     secrets: inherit
 
   keycloak-deploy-active-active:
```
pruivo, line 52: nitpicking
```suggestion
  keycloak-deploy-active-active-with-persistent-sessions:
```

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
mhajas: Hmm. I think you are correct. I thought it is not necessary and `MULTI_SITE` + `PERISTENT_SESSIONS` should be enough to enable the Remote implementation factories for the entities other than sessions. However, it seems not to be true at the moment so we should add it here as well.

mhajas:
@ryanemerson @pruivo Thank you for the reviews, I addressed your comments and will merge this so we have a run tomorrow

mhajas (approved):

PR #913, opened 2024-08-06
Throw an error when Cross-DC operations fail

### LINKED ISSUE #531: Throw an error when Cross-DC operations fail (in SYNC mode)
When running in SYNC mode in Cross-DC, a write to the other DC can fail. 

The current (default) behavior: 
* The request will time out
* The call returns to the caller where the data is the current site, but not the remote site

Due to that, the current and the remote site are then out of sync, so a session might not exist in the remote DC, or might not have been deleted in the remote DC. 

When discussion this with @pruivo, I understand that this behavior can be changed, so that there is an exception. When there is an exception, the expectation is that no changes written to the local Infinispan (both the internal and the external Infinispan). 

With this change, there will be no longer the need to issue a state transfer when single writes fail. It will only be necessary to do a state transfer when the other side becomes offline.

This issue is to 

* Check if the understanding of the current and target state is correct
* Reconfigure Infinispan and Keycloak to match the target state
* Run tests to verify the new target behavior

Possible edge cases: 

* The writes to Infinispan are in a second phase of the transaction which is `enlistAfterCompletion` - due to that, a write to the database might complete including a commit, and then the write to Infinispan fails. This might happen when changing a client or user, but shouldn't happen when logging in, refreshing tokens and logging out as those are read-only for the database.

pruivo:
Closes #531

### DIFF
--- a/provision/infinispan/Utils.yaml
+++ b/provision/infinispan/Utils.yaml
@@ -101,6 +101,9 @@ tasks:
         --set hotrodPassword={{.CROSS_DC_HOT_ROD_PASSWORD}}
         --set cacheDefaults.crossSiteMode={{.CROSS_DC_MODE}}
         --set cacheDefaults.stateTransferMode={{.CROSS_DC_STATE_TRANSFER_MODE}}
+        --set cacheDefaults.xsiteFailurePolicy={{.CROSS_DC_XSITE_FAIL_POLICY | default "FAIL" }}
+        --set cacheDefaults.txMode={{.CROSS_DC_TX_MODE | default "NON_XA" }}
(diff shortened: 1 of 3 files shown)


ahus1 (approved):
Thank you for this change. AFAIK those changes will show up in the docs once we do the next of the generated templates in the Keycloak main repository.

ahus1:
@ryanemerson - added you in the loop here so you are aware of the change. Please merge once you are happy with the change.

pruivo:
> Thank you for this change. AFAIK those changes will show up in the docs once we do the next of the generated templates in the Keycloak main repository.

That's correct.

ryanemerson:
Thanks for the heads up, I'll update the HA guide PR.

PR #917, opened 2024-08-07
Java 21

ryanemerson:


ryanemerson:
Closing as we either need Scala 2.3.15 or we need to migrate to Scala 3 before we can upgrade: https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

ahus1:
Reading the table, I see that Scala 2.13.11 support JDK 21. Looking at Gatling 3.10.5, it uses Scala 2.13.13, which should work. I'll push a commit.

ryanemerson:
It seems my reading comprehension deteriorates dramatically as we approach the end of the day :sweat_smile: 

Thanks for fixing this up @ahus1. Merging as the minikube failure is also present on main.

PR #921, opened 2024-08-09
Using Keycloak's new built-in OTEL tracing

ahus1:
This simplifies our setup and we won't need EFS any more. 

It will be a preview feature in KC26, and we will be able to add more items to it step-by-step.


mhajas (commented):
Thank you for the PR @ahus1! This simplifies things a lot! 

I have two questions:
1. The dashboard for [auth code and client credentials grant](https://github.com/keycloak/keycloak-benchmark/tree/main/provision/minikube/monitoring/dashboards) are relying on OTEL, are the dashboards still working the same way as before? Do we need to upgrade some metric names? 
2. Otel is now disabled by default and as far as I know we don't enable it fo nightly run, should we enable it now since the setup is much simpler? https://github.com/ahus1/keycloak-benchmark/blob/9b9ecabc622eea908d508a21fe84e335d3983f63/provision/common/Taskfile.yaml#L5

ahus1:
> The dashboards for auth code and client credentials grant are relying on OTEL, are the dashboards still working the same way as before? Do we need to upgrade some metric names?

Thank you for raising that question. They used to rely on OTEL. In Keyloak 25 (?) we were able to finally enable Keycloak's native HTTP metrics including histograms, and now they work independently of OTEL. In fact, they are now available for any nightly run. Talking about that, I can now revert part of the 4992a276b9a6c2a7398c56d2cab3c0393c04dc5b change as `REDIRECT` is no longer used. I'll push a change.

> Should we enable it now since the setup is much simpler?

I'd love to. Still I'd like to have this as a follow-up issue once this is agreed upon:  I'd like to do a side-by-side analysis how much performance impact enabling OTEL has. On our last run we tracked down 3 ms regression, so it seems we are are spotting even small regressions here.  

Does this answer your questions, and it is ok to merge?

provision/rosa-cross-dc/Taskfile.yaml
```diff
@@ -576,20 +576,17 @@ tasks:
...
         KUBECONFIG="{{.ISPN_DIR}}/.task/kubecfg/{{.ROSA_CLUSTER_NAME}}"
         helm upgrade --install jaeger jaegertracing/jaeger --version {{.JAEGER_VERSION}} -n monitoring -f "{{.ROOT_DIR}}/../minikube/jaeger/values.yaml"
-      - envsubst < "{{.ROOT_DIR}}/../minikube/jaeger/deployment-patch.yaml" > .task/jaeger-{{.ROSA_CLUSTER_NAME}}-patchfile.yaml
-      - >
-        KUBECONFIG="{{.ISPN_DIR}}/.task/kubecfg/{{.ROSA_CLUSTER_NAME}}"
-        oc patch deployment jaeger -n monitoring --patch-file .task/jaeger-{{.ROSA_CLUSTER_NAME}}-patchfile.yaml
+        --set allInOne.podSecurityContext.runAsUser=$(cat .task/monitoring-uids-{{.ROSA_CLUSTER_NAME}})
+        --set allInOne.podSecurityContext.runAsGroup=$(cat .task/monitoring-uids-{{.ROSA_CLUSTER_NAME}})
+        --set allInOne.podSecurityContext.fsGroup=$(cat .task/monitoring-uids-{{.ROSA_CLUSTER_NAME}})
+        --set allInOne.extraEnv[0].value={{.RETENTION}}
```
mhajas, line 587: I tried this locally and I got docker.io rate limit error. We can add 
```
--set allInOne.image.registry=quay.io
```
to load the image from quay. This can then be applied also to other jaeger helm chart installations.

mhajas (commented):
Yes, this answers my questions. Thank you.

I added one more comment related to docker.io rate limits.

provision/rosa-cross-dc/Taskfile.yaml
ahus1: Thanks, added it to the values.yaml instead.

PR #922, opened 2024-08-12
Testsuite fixes

ryanemerson:
Reinstate WARN on backup failures until take-offline configuration in place, plus misc fixes to deployment scripts/actions.

This will fix the testsuite for deployments with persistent sessions, however a 401 is still thrown when testing with volatile sessions.

### DIFF
--- a/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
+++ b/.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   keycloak-deploy-active-active-with-external-infinispan:
-    name: ROSA Scheduled Create Active/Active cluster with External Infinispan
+    name: ROSA Scheduled Create Active/Active cluster with External Infinispan and volatile sessions
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
     with:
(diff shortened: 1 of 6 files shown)


.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -52,14 +52,13 @@ jobs:
 
   keycloak-deploy-active-active:
     needs: keycloak-undeploy-active-active-with-external-infinispan
-    name: ROSA Scheduled Create Active/Active cluster with Persistent Sessions
+    name: ROSA Scheduled Create Active/Active cluster with External Infinispan and Persistent Sessions
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-multi-az-cluster-create.yml
     with:
       clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
-      enablePersistentSessionsFeature: true
```
pruivo, line 60: How do we enable persistent sessions?

ahus1: @ryanemerson - thank you for the explanation. Assuming this is now resolve, I'll go a head and merge this. 

@ryanemerson / @mhajas - it would be good to figure out how this would be communicated (both within the team, and then to users who want to try out this experimental feature).

ahus1 (approved):
LGTM, thank you for identifying the root cause, and also aligning the workflow files with Keycloak's behavior around that.

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
mhajas: Persistent sessions can still be enabled using `persistent-user-session` feature. Multi-site is just another option how to enable it.

mhajas (approved):
Thank you @ryanemerson LGTM

PR #928, opened 2024-08-13
Take sites offline on split-brain

### LINKED ISSUE #930: Add a mechanism to automatically take Infinispan backup sites offline on split-brain
### Description

# Problem

If an exception is thrown on xsite replication errors, as introduced by #531, it's no longer possible for a Keycloak cluster to make progress once a split-brain has occurred as an exception will be thrown on every Infinispan cache operation until the Infinispan backup site has explicitly been taken offline.

Infinispan provides an automatic take-offline configuration for individual caches, however this problematic in practice as:

1. A site is only taken offline after a configured number of failures and/or min-wait time has passed between exceptions
2. The offline status is updated per Infinispan node only when an entry hosted by that node is affected by a failure. There's currently no mechanism to update the status across all nodes for a given cache.

Given that Keycloak uses a number of different caches and each cache is associated with specific Keycloak functionality, it's possible that it could be a significant period of time after split-brain has occurred before all caches are taken offline.

# Solution

Instead of relying on Infinispan's take-offline capabilities, we can enhance the STONITH Lambda required by multi-site deployments so that it explicitly takes offline all caches when a split-brain is detected. The advantage of this approach is that it ensures backup sites are never taken offline until after the Global Accelerator has been updated.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
STONITH Lambda updated so the backup site of the "winning" AZ is explicitly taken offline. This is necessary in order for us to utilise a FAIL xsite strategy, as without it, the winning AZ would not be able to continue to serve user requests.

This approach requires the user to perform additional configuration steps:

1. A "infinispan" label must be added to the `PrometheusAlert` definition in each site. This label contains the exposed URL of the Infinispan cluster in that site.
2. An Infinispan username and password must be configured in the Lambda. The password should be retrieved via the AWS SecretsManager the same as the Keycloak password.

In order to allow 1. in our scripts, I have created an additional task "deploy-xsite-infinispan-monitoring".

Closes #930

### DIFF
--- a/provision/opentofu/modules/aws/accelerator/src/stonith_lambda.py
+++ b/provision/opentofu/modules/aws/accelerator/src/stonith_lambda.py
@@ -1,10 +1,31 @@
-# tag::stonith-start[]
+# tag::fencing[]
+from urllib.error import HTTPError
+
 import boto3
 import jmespath
 import json
+import os
+import urllib3
 
 from base64 import b64decode
 from urllib.parse import unquote
+
+# Prevent unverified HTTPS connection warning
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+class MissingEnvironmentVariable(Exception):
+    pass
+
+
+class MissingSiteUrl(Exception):
+    pass
+
+
+def env(name):
+    if name in os.environ:
+        return os.environ[name]
+    raise MissingEnvironmentVariable(f"Environment Variable '{name}' must be set")
 
 
 def handle_site_offline(labels):
@@ -40,6 +61,9 @@ def handle_site_offline(labels):
             EndpointConfigurations=endpoints
         )
         print(f"Removed site={offline_site} from Accelerator EndpointGroup")
+
+        take_infinispan_site_offline(reporter, offline_site)
+        print(f"Backup site={offline_site} caches taken offline")
     else:
         print("Ignoring SiteOffline alert only one Endpoint defined in the EndpointGroup")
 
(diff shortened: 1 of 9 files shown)


provision/opentofu/modules/aws/accelerator/src/stonith_lambda.py
```diff
@@ -40,6 +53,9 @@ def handle_site_offline(labels):
             EndpointConfigurations=endpoints
         )
         print(f"Removed site={offline_site} from Accelerator EndpointGroup")
+
```
pruivo, line 56: not related to this line...
where is the label `site` defined? In line 48, you have `offline_site = labels['site']` but in `infinispan-alerts.yaml` no label with the name `site` exists.

pruivo (commented):
Just have a couple questions, but it looks good to me :+1:

provision/opentofu/modules/aws/accelerator/src/stonith_lambda.py
ryanemerson: The label is automatically included because we use it in the metric query, the same is true of `namespace`

pruivo (approved):

ryanemerson:
Marking as draft as I need to verify the "Fix endpoint selection in Lambda" commit works as expected.

ryanemerson:
Approach refactored so the PrometheusRule no longer be updated, but users must now provide the following Env variables to configure the Lambda:

```bash
INFINISPAN_USER= # The username used to authenticate calls to the Infinispan cluster
INFINISPAN_USER_SECRET= #  name of the aws secret containing the password used to authenticate calls to the Infinispan cluster
WEBHOOK_USER= # The username used to authenticate calls to the webhook
WEBHOOK_USER_SECRET= # The name of the aws secret containing the password used to authenticate calls to the webhook
SECRETS_REGION= # The region containing the aws secrets
```

ahus1 (approved):
Thank you for this change. I'll open a follow-up issue to track that we update the docs that this lambda or an equivalent is required once you choose FAIL.

PR #929, opened 2024-08-14
Endpoint in global accelerator is missing after functional tests are run

### LINKED ISSUE #916: Endpoint in global accelerator is missing after functional tests are run

ryanemerson:
Closes #916

### DIFF
--- a/provision/infinispan/ispn-helm/templates/infinispan-alerts.yaml
+++ b/provision/infinispan/ispn-helm/templates/infinispan-alerts.yaml
@@ -21,7 +21,7 @@ spec:
     groupBy:
       - accelerator
     groupInterval: 90s
-    groupWait: 0s
+    groupWait: 60s
     matchers:
       - matchType: =
         name: alertname


ahus1:
Ryan wrote: 
> The root cause seems to be that the Lambda is triggered twice per site for every functional test run. I think this is to do with the fact we're using group_by: 0 in the AlertManagerConfig. Originally I set such a low value to help local testing, as we wouldn't want old alerts being delayed for too long, but I believe we have to increase this to at least 60s in order for the CI to play nice

I did some research and didn't find so much on any min values for group_by.
In our OpenShift setup, there are two user prometheus instances for the user workload, and two alert manager instances.

The two prometheus instances will scrape in parallel (for HA), and send out alerts to the alert manager independently. This could trigger the two alerts.

The alertmanagers form a cluster, and the second will send out the alerts only if the first one didn't send out the alert for another X seconds. Again, this is a HA setup. 

Unfortunately this doesn't fully explain to me why there are two notifications. Maybe this just doesn't work for a group_by of 0 seconds. So let's use the value as you proposed here.

ryanemerson:
@ahus1 I found [this issue](https://github.com/prometheus/alertmanager/issues/2647) very informative about how the prometheus AlertManager handles notifications and the impact of the various configuration toggles.

My understanding is that the issue isn't that the alert is fired multiple times, as this should be expected, but it's how the AlertManager reacts to multiple events and whether it fires one or more notifications. By setting `group_by: 60s` we're affectively ensuring that all SiteOffline events within a 60s period only fire a single AlertManager notification, i.e. calling the STONITH Lambda webhook. Previously this value was `0s`, which would mean that no grouping was happening and every time a SiteOffline alert was raised the webhook would be invoked.

I came to the value of 60s based upon the AlertManager debug logs after the daily scheduled run, which showed the second SiteOffline alert was processed by the AlertManager within a minute.

ahus1:
Thank you for referencing the issue. Reading it, I think the following was happening before:

* First alert received
* First notification sent out immediately
* Second alert received
* Second notification sent out 90 seconds after the first

What is happening now: 

* First alert received, now in group waiting state for 60 seconds
* Second alert received
* First notification sent out

Maybe the root cause is the way we're currently restoring the state of the system in our test cases: 

* Once we restore the xsite connection, we should wait for Prometheus to scrape the Infinispan metrics and to recover the alert
* Only then it would make sense to restore the LB state

At the same time, this is what I see we are doing in the `failbackLoadBalancers()`. So I'm still confused why this happens.

ryanemerson:
> Reading it, I think the following was happening before:
> 
>     * First alert received
> 
>     * First notification sent out immediately
> 
>     * Second alert received
> 
>     * Second notification sent out 90 seconds after the first
> 
> 
> What is happening now:
> 
>     * First alert received, now in group waiting state for 60 seconds
> 
>     * Second alert received
> 
>     * First notification sent out

That's a nice succinct description of what I believe to be occurring :+1: 

> At the same time, this is what I see we are doing in the `failbackLoadBalancers()`. So I'm still confused why this happens.

My understanding is that Prometheus Alert firing and AlertManager notification firing are independent, with the latter governed by the `group_interval`, `group_by` parameters etc, which is why waiting for the prometheus alert to no longer be present is not sufficient.

AFAIK there's no way for us to query the AlertManager instances or force notifications to be dropped/flushed.

PR #935, opened 2024-08-15
Reverting to an old setup where keep all sessions in memory on the Infinispan side of things

### LINKED ISSUE #934: Revert change to have only 10_000 entries in Infinispan caches
### Describe the bug

We always allows only 10_000 entries in remote Infinispan for sessions. This was done when we first introduced persistent sessions. This is now obsolete, as with persistent sessions we don't use the external caches, and it is wrong when we use them for the external Infinispan only setups.

### Version

main

### Expected behavior

All valid entries should be kept in the cache.

### Actual behavior

10_000 entries max in the external cache.

### How to Reproduce?

Look at the nightly run.

### Anything else?

_No response_

ahus1:
Closes #934

### DIFF
--- a/provision/infinispan/ispn-helm/values.yaml
+++ b/provision/infinispan/ispn-helm/values.yaml
@@ -21,27 +21,15 @@ cacheDefaults:
   txLockMode: PESSIMISTIC
 caches:
   sessions:
-    owners: 1
-    memory:
-      maxCount: 10000
     mergePolicy: ALWAYS_REMOVE
   actionTokens: {}
   authenticationSessions:
     mergePolicy: ALWAYS_REMOVE
   offlineSessions:
-    owners: 1
-    memory:
-      maxCount: 10000
     mergePolicy: ALWAYS_REMOVE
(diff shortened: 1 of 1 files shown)


kami619:
@ahus1 do we also need to update the following locations? 

https://github.com/keycloak/keycloak-benchmark/blob/a1a40e761604c493e7deada0726780ea0a77dd17/provision/minikube/keycloak/config/kcb-infinispan-cache-config.xml#L31

https://github.com/keycloak/keycloak-benchmark/blob/a1a40e761604c493e7deada0726780ea0a77dd17/provision/minikube/keycloak/config/kcb-infinispan-cache-remote-store-config.xml#L32

ahus1:
@kami619 - no, those locations are local embedded caches where we we cache entries which we retrieved from the database earlier. Those are expected to be limited. Everything that is evicted there is later re-loaded from the database when needed.

kami619:
thanks @ahus1.

kami619 (approved):
lgtm

PR #967, opened 2024-09-05
Limit Infinispan cluster resources

pruivo:
CPU: 2 requested, max of 4
Memory: 1GB requested, max of 2GB

offtopic, updated benchmark defaults to match the GH actions nightly runs.


ahus1 (approved):

ahus1:
Just wondering: What have been the defaults before?

pruivo:
> Just wondering: What have been the defaults before?

The memory was limited to 1GB (more than enough even with volatile sessions) and the CPU was unlimited (in other words, limited to the VM cores).

PR #968, opened 2024-09-05
modify rate to irate in grafana dashboards

kami619:
This allows us to identify more precisely in time when the system behavior changes.


ahus1 (approved):

ahus1:
LGTM, added to the commit comment and the description "why" we did the change.

PR #970, opened 2024-09-06
Updated Horreum schema version to 0.2

andyuk1986:


kami619:
@andyuk1986 Thanks Anna for this change, Does the schema itself it have any changes? or are we going to do that retroactively for the old datasets?

andyuk1986:
Hey Kamesh, the schema 0.2 doesn't contain the `persistentSessionsEnabled` variable, based on which I was filtering data in horreum. I have also updated the result jsons generated singe August 13 to have schema 0.2 version and updated that data in Horreum. So this PR is needed for newly uploads.

PR #979, opened 2024-09-13
Changed the ispn cache name for which the metric should be retrieved.

andyuk1986:
Updated the ISPN cache name as the sessions are stored in DB now. Chosen `authenticationSessions` cache, as it is used while login. After KC works mainly with DB.


pruivo:
merged! thanks @andyuk1986 !

PR #987, opened 2024-09-23
enable active passive tests in daily runs

### LINKED ISSUE #986: re-enable A/P nightly tests in ROSA scheduled run
```[tasklist]
### Tasks
- [ ] re-enable the workflow in the daily run for A/P tests, which should include, deploy, functional tests and benchmarks
```

kami619:
Fixes #986 

I am keeping it as a draft, as I am figuring out the health checks to run for both deployments.


.github/actions/get-keycloak-url/action.yml
```diff
@@ -14,11 +14,6 @@ runs:
       run: |
         KEYCLOAK_URL=https://$(kubectl get routes -n "${{ inputs.project }}" -l app=keycloak -o jsonpath='{.items[*].spec.host}')
         echo "KEYCLOAK_URL=$KEYCLOAK_URL" >> "$GITHUB_ENV"
-    - id: get-keycloak-site-url
```
kami619, line 17: moving the fetching of site url portion out of this action, as we would need to conditionally fetch this based on the LB setup.

.github/workflows/keycloak-multi-site-health-check.yml
```diff
@@ -59,6 +59,12 @@ jobs:
         with:
           project: ${{ inputs.project }}
 
+      - id: get-keycloak-site-a-url
```
kami619, line 62: My idea is to conditionally run this based on the deployment time from within this workflow, which is probably more flexible.

.github/workflows/rosa-cluster-auto-provision-on-schedule.yml
```diff
@@ -70,3 +70,42 @@ jobs:
...
+    name: Undeploy Keycloak A/A deployment on the multi-az cluster
+    if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+    uses: ./.github/workflows/rosa-multi-az-cluster-undeploy.yml
+    with:
+      clusterPrefix: gh-keycloak # ${{ env.CLUSTER_PREFIX }} -- unfortunately 'env.' doesn't work here
+      skipAuroraDeletion: true
+      activeActive: true
+    secrets: inherit
+
+  keycloak-deploy-active-passive:
```
kami619, line 85: Still need to add health checks once I fix the above mentioned comments.

PR #993, opened 2024-09-30
update health checks to work for both A/A and A/P deployments

kami619:
This PR addresses,
- Addition of script to fetch the Route53 domain dynamically and then construct the Keycloak Site Url.
- Add the health checks for Route 53 to the scheduled workflow.

Example Output of the script,

```
> DOMAIN=$(./route53_fetch_domain.sh "keycloak-benchmark.com.")
> echo $DOMAIN
gh-keycloak-a-gh-keycloak-b-mjqxmdyk.keycloak-benchmark.com
```


mhajas (approved):
Thank you @kami619

PR #994, opened 2024-10-01
Increase number clients-per-realm for all kcb scenarios

mhajas:


.github/workflows/rosa-scaling-benchmark.yml
```diff
@@ -219,6 +219,7 @@ jobs:
           --logout-percentage=0
           --measurement=${{ inputs.measurement }}
           --users-per-realm=${{ inputs.usersPerRealm }}
+          --clients-per-realm=${{ env.CLIENTS_PER_REALM }}
```
ryanemerson, line 222: Looking at `main`, it seems the default is `20000`. Does that make sense still?

mhajas: We had clients and users set to the same number 20k. Then based on the recent limitation document we changed users to 100k. I decided to not increase clients as well so I kept it to be 20k by default. Since the workflow already has too many inputs, I could not add another for clients so I decided to hardcode 20k for clients. I would say it is a reasonable default as the `realms` cache has 10k count limit. We can set it to the same number as users though. 

On the other hand, I think 20k is reasonable default to detect performance changes in entities cached in `realms` cache.

ryanemerson: Thanks for explaining, that seems reasonable to me. Let's merge as is, see the performance results tomorrow and then we can modify the value if required.

ryanemerson:
Thanks @mhajas

PR #999, opened 2024-10-03
Feedback on Metrics

pruivo:
* Hide Hot Rod metrics (reason: not released yet)
* Fix missplaced endif
* Add largest JGroups thread pool size metric
* Add hit and success ratios example to Hot Rod metrics


ryanemerson (approved):

ryanemerson:
Thanks @pruivo

Issue #945, opened 2024-08-20
Handling secure cookies for localhost URLs 


ahus1:
### Describe the bug

Starting with https://github.com/keycloak/keycloak/issues/16770 Keycloak changed its handling of secure cookies. While this works with browsers, it doesn't work with Gatling: https://github.com/gatling/gatling/issues/4578

### Version

main

### Expected behavior

Gatling should work with URLs "localhost" when logging in using the UI. 

### Actual behavior

The authentication flow won't work with "localhost" addresses.

### How to Reproduce?

Run the scenario `keycloak.scenario.authentication.LoginUserPassword` against a URL like `http://localhost:8080`

```
./bin/kcb.sh --scenario=keycloak.scenario.authentication.LoginUserPassword --server-url=http://localhost:8080 --realm-name=master --username=admin --user-password=admin --users-per-sec=30 --ramp-up=1 --client-id=test --client-secret=*** --client-redirect-uri=http://localhost:8080 --measurement=600 --log-http-on-failureLoginUserPassword
```

### Anything else?

This is waiting for Gatling upstream to fix the issue. In the meantime, a safety switch is in place which will case an early failure and will show a descriptive error message.

ahus1:
The change has been reverved as Gatling stopped combining multiple Gatling Logs in version 3.12.0. Although it might have happend unintendedly, they consider this to be a feature of the enterprise version and don't plan to re-add it. 

https://github.com/gatling/gatling/issues/4591

ahus1:
After downgrading to the last version that support this, I think it is time to move on to another load testing tool. 
We'll prioritize #829 then.

PR #1012, opened 2024-10-14
Fix error creating EC2 runner instances

mhajas:
This fix comes from @ryanemerson commit: https://github.com/ryanemerson/keycloak/commit/8174bb26932cba2b8ad984d073ae20f34ba038a4 

Tested in my personal GHA: https://github.com/mhajas/keycloak-benchmark/actions/runs/11326943860/job/31496938839

### DIFF
--- a/.github/actions/ec2-create-instances/action.yml
+++ b/.github/actions/ec2-create-instances/action.yml
@@ -13,6 +13,8 @@ runs:
       name: Install the required Ansible AWS collections.
       shell: bash
       run: |
+        python3 -m venv .venv
+        source .venv/bin/activate
         ./aws_ec2.sh requirements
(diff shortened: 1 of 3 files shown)


mhajas:
So the fix doesn't work fully, there are some additional failures

ryanemerson:
> This fix comes from @ryanemerson commit: [ryanemerson/keycloak@8174bb2](https://github.com/ryanemerson/keycloak/commit/8174bb26932cba2b8ad984d073ae20f34ba038a4)

This should be sufficient. The errors on KC that I was encountering were because we were using a RHEL 8 based image in the EC2 image, which is not applicable here.

ahus1:
Problem seen in Michals PR should be resolved after rolling back the Gatling changes, see #945

PR #1019, opened 2024-10-22
Use default machinepool instead of additional autoscaling one

mhajas:


kami619:
@mhajas I think the changes look good to me.

.github/workflows/keycloak-scalability-benchmark.yml
```diff
@@ -181,10 +177,6 @@ jobs:
         with:
           project: ${{ env.PROJECT }}
 
-      - name: Scale down the cluster
```
kami619, line 184: Would we still want this to be there? If the cluster decides to autoscale based on this scaling group?

mhajas: This step is not scaling down anything if the resources are still used. It is allowing the machinepool to scale to 0 if nodes are not needed. After this PR we are never setting min-replicas to > 0 so we don't need this step.

ryanemerson:
@mhajas Do you have a link to a successful action creating clusters?

mhajas:
> @mhajas Do you have a link to a successful action creating clusters?

I don't have any. My idea was to test it in "production" 😈. I can give it a try though.

ryanemerson:
After discussing on slack, we have agree to merge and fix any issues that appear on the daily run tomorrow.

PR #1020, opened 2024-10-22
add Keycloak SLO docs fixes

### LINKED ISSUE #579: Metrics - Most important SLI to determine that Keycloak is healthy
For customers to be able to run Keycloak in a production environment, they should have an overview of the ~10 metrics on both Keycloak an Infinispan so they can be both confident that their system is running ok, and anyone supporting their setup can ask for those metrics (eventually a dashboard).

Example include: 
* Number of cache entries per Keycloak and Infinispan pod
* Memory and CPU usage per pod
* JGroup thread usage
* Cache hit ratio + cache size for local caches 
* ...

kami619:
Fixes #579


ahus1 (changes requested):
Thank you for this PR. It captures the discussions we had internally, still it would need to change its shape a bit to be useful as a future guide in Keycloak main.

* The introduction needs to briefly explain SLI/SLO to our best knowledge
* It should be more specific about which metrics should be used, as readers would otherwise be lost in the metrics and their tags. 
* It needs to be less specific about actual SLOs, as those would require a lot more assumptions to be explained and calculations to be made

See below for the comment.

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,36 @@
+= {project_name} Service Level Indicators
+:description: This document contains details of the SLI's to monitor your {project_name} deployment's performance.
+
+To ensure that {project_name} can be confidently run in a production environment, it is important for customers to have an overview of key metrics from both {project_name} and {jdgserver_name}. This will allow them to assess the health and performance of their system, ensuring smooth operation. Additionally, these metrics will provide critical insight for anyone supporting the deployment, allowing them to request and analyze the necessary data effectively.
```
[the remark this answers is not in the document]
kami619: I was thinking along the lines about the metrics needed further down to troubleshoot. I can restrict the definition to SLI alone, in this page to make it clearer.

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,36 @@
...
+* access and use the admin console.
+* and manage my profile through the account console.
+
+So that I can interact with the {project_name} system effectively and perform the necessary tasks without interruption.
+====
+
+
+[cols="1,1,1,1,2,2,2", options="header"]
+|===
+| SLO         | SLO definition                                                              | Single Site SLO Target  | Multi-Site SLO Target  | SLI Metric                                                                      | Metric Details                                                                                                        | Dashboard
```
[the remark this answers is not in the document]
kami619: Fine, we can leave the targets for single and multi-site to individual deployment types of the Keycloak users, which came up even during a discussion with @mhajas and @pruivo. However, I still think we need to layout the SLOs and the relevant SLIs in the table.

kami619:
@ahus1 I have addressed most of the review feedback, and I still have some doubts on how to implement your input on the PromQL queries. I will connect with you offline and work on it.

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+| {project_name} should be available 99.99% of the time.                 | Availability            | Uptime percentage is the ratio of successful authentication requests to total authentication requests.               |
+The `up` metric which indicates if the Prometheus server is able to scrape metrics from the {project_name} instance.
+This metric will have a value of 1 if the {project_name} service is available
+and responding to Prometheus scrape requests,
+and 0 if the service is down or unreachable.
+
+It could be represented in a pseudo PromQL query as `sum(up{...}) > 0`
+
+| 95% of {project_name} authentication requests should have a latency below 200ms. | Authentication Latency  | {project_name} server-side metrics to track latency for specific endpoints along with Response Time Distribution.            | `http_server_requests_seconds_bucket`, `http_server_requests_seconds_count`.
```
pruivo, line 41: there is no 200ms bucket in the metrics, should we change this to 250ms?

ahus1: I'm currently reworking this, and will update it 250 ms

pruivo: ack, the doc looks good to me.

kami619:
@ahus1 

Thanks, Alex for the detailed review.

- The SRE book-styled definitions are a standard approach, and I agree with those changes.

- SLO interval definitions also create a good starting point for discussions outside of our team; it's a good idea to do it that way.
 
- Thanks for making the PromQL query section more presentable and practical, I owe you one this :) 
 
- Coming to the 4 golden signals, we are not referencing the other two Throughput(Traffic) and Saturation, do we want to mention them or want to keep them out of this page for now?

ahus1:
Hi Kamesh, when looking at SLIs that measure user facing behaviour, I don't see the need for capturing throughput and saturation separately.  If something saturates within Keycloak this would affect the SLIs we already have (slower responses, more errors).
There will be other dashboards for trouble shooting and capacity planning where we will come back to them.

I think we reached the "as simple as possible" here. 

If you agree we can merge it and as the community for feedback once it has been published on GitHub pages.

kami619:
I am very happy with the final result. Thanks again @ahus1. I think we can merge it now.

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,160 @@
...
+      uri=~"/realms/{realm}/protocol/{protocol}/.*|/realms/{realm}/login-actions/.*", # <1>
+      container="keycloak",
+      namespace="$namespace"}
+    [5m] # <3>
+  )
+) without (le,uri,status,outcome,method,pod,instance) # <5>
+----
+<1> URLs related to logging in
+<2> Response time as defined by SLO
+<3> Filter by additional tags like
```
pruivo, line 118: Does the sentence look unfinished to you? @ahus1

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,160 @@
...
+[source,plaintext]
+----
+sum(
+  rate(
+    http_server_requests_seconds_count{
+      uri=~"/realms/{realm}/protocol/{protocol}/.*|/realms/{realm}/login-actions/.*", # <1>
+      outcome="SERVER_ERROR", # <2>
+      container="keycloak", # <3>
+      namespace="$namespace"}
+    [2m] # <4>
```
pruivo, line 137: In the table, we say 5 minutes intervals. Do you think we should use that value in this example? @ahus1

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
```diff
@@ -0,0 +1,160 @@
...
+sum(
+  rate(
+    http_server_requests_seconds_bucket{
+      uri=~"/realms/{realm}/protocol/{protocol}/.*|/realms/{realm}/login-actions/.*", # <1>
+      le="0.25", # <2>
+      container="keycloak", # <3>
+      namespace="$namespace"}
+    [5m] # <4>
+  )
+) without (le,uri,status,outcome,method,pod,instance) # <5>
```
pruivo, line 104: Should we remove the `URI` from the list (assuming it gives a per-URI metric)? 
Would the fast endpoints skew the results in some way?

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
ahus1: fixed in next commit

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
ahus1: fixed in next commit

doc/kubernetes/modules/ROOT/pages/running/metrics/keycloak_service_level_indicators.adoc
ahus1: > Should we remove the URI from the list (assuming it gives a per-URI metric)?

no, as to a user using the login screen it is either fast all the time or sometimes slow. So it is better to have one number IMHO

> Would the fast endpoints skew the results in some way?

It would skew a lot when you would go for the median. As we're going for a percentile, one slow endpoint would show up in the percentile, so we're fine here. It would show up even more if we would go to 99% instead of 95%.

pruivo: Thanks!

PR #1021, opened 2024-10-23
Reduce ClientSecrets scenario load

### LINKED ISSUE #1018: Why 20k clients is so slow
Cache usage: 2x in users cache, 3x in realms cache per client. 

-> 10k clients

-> reduce RPS to a number where the DB is no longer overload -> reduce to 50% to not overload the database

-> Write some docs on how to size users and client cache based on the number of clients

Follow-up task: update the sizing guide for RHBK26

pruivo:
* Current performance run is (over)loading the DB too much (98% CPU)
* Reduce the total number of clients to 10k (75% cache hit ratio)
* Reduce the requests per second to 750 (25% reduction)

Closes #1018

### DIFF
--- a/.github/workflows/rosa-scaling-benchmark.yml
+++ b/.github/workflows/rosa-scaling-benchmark.yml
@@ -30,7 +30,7 @@ on:
       numberOfClientsPerSecond:
         description: 'Client credential grants per second'
         type: number
-        default: 1000
+        default: 750
       measurement:
         description: 'Measurement period (seconds)'
         type: number
@@ -73,7 +73,7 @@ on:
       numberOfClientsPerSecond:
         description: 'Client credential grants per second'
         type: number
-        default: 1000
+        default: 750
       measurement:
         description: 'Measurement period (seconds)'
         type: number
@@ -93,7 +93,7 @@ env:
   PROJECT_PREFIX: runner- # same as default
   PROJECT: runner-keycloak
   ANSIBLE_CUSTOM_VARS_ARG: '-e @env_rosa_benchmark.yml'
-  CLIENTS_PER_REALM: 20000
+  CLIENTS_PER_REALM: 10000
 
 jobs:
   run:
--- a/provision/rosa-cross-dc/Taskfile.yaml
+++ b/provision/rosa-cross-dc/Taskfile.yaml
@@ -792,6 +792,7 @@ tasks:
     vars:
       SCENARIO: '{{.SCENARIO | default "keycloak.scenario.authentication.AuthorizationCode"}}'
       USERS: "{{.USERS | default 20000}}"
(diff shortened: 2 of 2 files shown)


.github/workflows/rosa-scaling-benchmark.yml
```diff
@@ -94,6 +94,7 @@ env:
   PROJECT: runner-keycloak
   ANSIBLE_CUSTOM_VARS_ARG: '-e @env_rosa_benchmark.yml'
   CLIENTS_PER_REALM: 20000
+  GATLING_CLIENTS_PER_REALM: 10000
```
mhajas, line 97: Why are we distinguishing between clients created and clients used in load running? Should we just change the variable CLIENTS_PER_REALM to 10k?

provision/rosa-cross-dc/Taskfile.yaml
```diff
@@ -792,6 +792,7 @@ tasks:
     vars:
       SCENARIO: '{{.SCENARIO | default "keycloak.scenario.authentication.AuthorizationCode"}}'
       USERS: "{{.USERS | default 20000}}"
+      CLIENTS: "{{.CLIENTS | default 10000}}"
```
mhajas, line 795: Is this used somewhere? Should we remove from the Taskfile if not?

.github/workflows/rosa-scaling-benchmark.yml
```diff
@@ -378,7 +379,7 @@ jobs:
           --ramp-up=20
           --measurement=${{ inputs.measurement }}
           --users-per-realm=${{ inputs.usersPerRealm }}
-          --clients-per-realm=${{ env.CLIENTS_PER_REALM }}
+          --clients-per-realm=${{ env.GATLING_CLIENTS_PER_REALM }}
```
mhajas, line 382: Should we change this for all test runs not only for the ClientSecret scenario?

.github/workflows/rosa-scaling-benchmark.yml
pruivo: I though we still wanted to create 20k but only use 10k. If I'm wrong, then I can remove this variable.

provision/rosa-cross-dc/Taskfile.yaml
pruivo: I use the task file to trigger the runs when I need it. I want to keep it in sync with the nightly defaults.

.github/workflows/rosa-scaling-benchmark.yml
mhajas: I don't think there is much difference between creating 20k and 10k so I would just use 10k everywhere to save time creating the additional 10k. Also, we previously used 1 client-per-realm everywhere, so 10k is still better.

PR #1023, opened 2024-10-30
Add eviction metric

pruivo:
Describe the eviction metric and define how it can be used to tune the cache sizes


pruivo:
@ahus1 if you wanna do your magic :smile:

ahus1:
Thank you for this documentation update, looks great! Around the eviction rate, I added an extra bit of cause-and-effect. 

I saw that you added an `ifndef::cross_site[]` - IMHO the `users` and `realms` cache is present in all setup we currently have, as they will locally cache data. 

Please merge if you think this is now ready to be merged, or continue the discussion when I missed a point.

pruivo:
The `ifndef::cross_site[]` hides these docs updates from the external Infinispan metrics page. The `cross_site` is only set by the external Infinispan cluster metric page: 

https://github.com/keycloak/keycloak-benchmark/blob/4545887ae84d2ca6bdd8c48f96c2fe5559c8a186/doc/kubernetes/modules/ROOT/pages/running/metrics/external_infinispan.adoc?plain=1#L5

I'll change the attribute name to make it more clearer.

pruivo:
@ahus1 I made a mistake! I force-pushed the branch to rename the attribute and it deleted your changes by mistake. Can you add them again? I'm sorry.

ahus1:
Everything is in Git, so we are safe.

PR #1025, opened 2024-10-31
Update to ARM worker node types

### LINKED ISSUE #1022: Support ARM based Keycloak deployments on ROSA
### Description

- Identify remaining services on Keycloak Benchmark that don't support ARM, possibly leading to a mixed ARM and non-ARM cluster

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

kami619:
Fixes #1022

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -14,7 +14,7 @@
 2. Click on Run workflow button
 3. Fill in the form and click on Run workflow button
    1. Name of the cluster - the name of the cluster that will be later used for other workflows. Default value is `gh-${{ github.repository_owner }}`, this results in `gh-<owner of fork>`.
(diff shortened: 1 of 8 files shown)


ahus1:
Thank you - I think it will be green when you roll back the changes in `provision/minikube/keycloak/templates/keycloak.yaml`, the build will be green. Still, we might want those changes to be able to test in a mixed cluster, so then they would need to be conditional on an architecture selector.

mhajas:
We can use this in `keycloak.yaml`
```
{{ if eq .Values.environment "openshift" }}
```

ahus1:
Hm, with the change `{{ if eq .Values.environment "openshift" }}` we now enforce ARM on OpenShift all the time. We would need to do this only on mixed clusters, as arm would be the default anyway in a pure ARM cluster. I'll comment in the parent issue about what is in scope and what not.

mhajas (approved):
I am merging this so we have the results on Monday

PR #1026, opened 2024-11-01
Fix jvm_gc_overhead metric name

mhajas:
It seems `jvm_gc_overhead_percent` was renamed to `jvm_gc_overhead`.


ahus1:
Thanks, it seems it was fixed here: https://github.com/micrometer-metrics/micrometer/pull/4765/files

PR #1030, opened 2024-11-05
Add Grafana dashboard based on latest metrics docs

mhajas:


mhajas:
This PR contains changes that it should not. Also, I found out error responses graph does not work. I need to fix.

PR #1036, opened 2024-11-11
Change DB worker type to db.r6g.xlarge

### LINKED ISSUE #1035: Current nightly run is hitting 90% CPUUtilization of the database
### Describe the bug

In current nightly run (AuthorizationCode) we observed sudden drops in latency SLO metric (percentage of requests with response time lower than 250ms). 

image

As suggested in the screenshot above the drops correspond to higher amount of threads in WAITING state. After looking at the thread dump it is visible the threads are waiting for persistent session updates in the database.

Based in the CPUUTilization the database is the bottleneck here: 
image

I tested with bigger instance (2cpu -> 4cpu, `db.t4g.large` -> `db.r6g.xlarge` [t4g.xlarge [is not supported](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.DBInstanceClass.SupportAurora.html) for aurora]) and the results look much better:
image

Therefore I suggest to change the database comput for  `db.r6g.xlarge`.

I did also some additional experiment for changes in Keycloak settings for user sesssions update batching and the results are visible in this document:
https://docs.google.com/spreadsheets/d/1OnS8PPjwYo4i-APdyFd_A1w2dCw9NRyK/edit?usp=sharing&ouid=103250532662258091059&rtpof=true&sd=true
 

### Version

main

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

mhajas:
Closes #1035


ahus1:
When I was testing in June, the `db.t4g.large` was still available. Interesting...

PR #1051, opened 2024-11-21
enable the event metrics for users by default

### LINKED ISSUE #1050: enable event-metrics-user-enabled in Keycloak benchmark deployments
```[tasklist]
### Tasks
- [x] update the keycloak.yaml
- [x] run performance tests to see if there an overhead with and without the user event metrics
```

kami619:
Fixes #1050 partially.

### DIFF
--- a/doc/kubernetes/modules/ROOT/partials/running/metrics/kc_enable_metrics.adoc
+++ b/doc/kubernetes/modules/ROOT/partials/running/metrics/kc_enable_metrics.adoc
@@ -1,7 +1,7 @@
 = Enable {project_name} metrics
 
 {project_name} exposes metrics on the management interface endpoint `/metrics`.
-To enable, use the build time option `--metrics-enabled=true`.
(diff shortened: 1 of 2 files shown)


kami619:
Once we add these flags in the deployment, we are able to enable the user event metrics in the Keycloak nightly runs.

ahus1:
> Once we add these flags in the deployment, we are able to enable the user event metrics in the Keycloak nightly runs.

@kami619 - reading your comment, it seems to differ from the changes in this PR: Once this PR is merged, the user event metrics are enabled. 

Did you run a smoke test to see any performance changes before or after, or do you plan to merge it straight away and then try in out in our nightly run?

kami619:
@ahus1 thanks Alexander, we agreed to merge this and utilize the nightly run to observe the performance overhead, if any. I will work on the docs and the prototype dashboard in the meantime and raise another PR.

PR #1053, opened 2024-11-22
Update shade maven plugin add minimizeJar properties

lvassa88:
Update shade maven plugin add minimizeJar properties to reduce the size of the generated jar
Add the minimizedJar because need to upload the jar into blazmeter but it's support upload with max size 50MB.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -91,7 +91,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.4.0</version>
+                <version>3.6.0</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
@@ -122,6 +122,7 @@
                                     <exclude>org.scala-lang.modules:scala-java8-compat_2.13:*</exclude>
                                 </excludes>
                             </artifactSet>
+                            <minimizeJar>${minimize}</minimizeJar>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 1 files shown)


benchmark/pom.xml
```diff
@@ -151,4 +152,23 @@
...
+            <id>default</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <minimize>false</minimize>
+            </properties>
+        </profile>
+        <profile>
+            <id>blazemeter</id>
```
kami619, line 166: We do not utilize Blaze-meter tooling in our Keycloak Benchmark, @lvassa88 can you please elaborate how would this be useful for the folks who do not use the Blaze-meter in their frameworks ?

lvassa88: I try to use the keycloak benchmark jar on blameter to run test in different situation/timezone, but the jar that you generate is bigger for BlazeMeter tool, so I use the minimizeJar to solve the issue and upload the jar generated with this option and I think other guys can have the same problem.

lvassa88:
Hi @kami619, I have modified the PR to remove the profile and add only the minimizedJar configuration to reduce the sizing of the jar generated.

ahus1 (changes requested):
I tried this branch locally, and I found that once I packaged the JAR file and used it with the `kcb.sh` shell script we provide, I get an error

```
java.lang.ClassNotFoundException: io.gatling.app.Gatling
```

This indicates that the `minimizeJar` option removes the entry class that we depend on. 

Looking at Blazemeter, maybe they require only the scenarios and not the full Scala/Gatling bundle. If you build it locally, there is a file `original-keycloak-benchmark-0.15-SNAPSHOT.jar` in the target folder that contains only the scenarios. Maybe that is just enough for what you need for Blazemeter.

PR #1085, opened 2024-12-20
Fix tofu not removing accelerators

### LINKED ISSUE #1084: Nigthly reaper job is not removing Global accelerator
### Describe the bug

It seems to remove accelerators it is necessary for Openshift clusters and aws load balancers used in LoadBalancer service needs to still exist, however, we don't enforce this in reaper job and therefore it is likely this is not done in correct order.

https://github.com/keycloak/keycloak-benchmark/actions/runs/12420696525/job/34678788792

### Version

main

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

mhajas:
Closes #1084

### DIFF
--- a/.github/workflows/auto-delete-on-schedule.yml
+++ b/.github/workflows/auto-delete-on-schedule.yml
@@ -6,8 +6,29 @@ on:
     - cron: '0 21 * * *' # Runs every day at 9 PM UTC.
 
 jobs:
+  accelerator-delete:
+    if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Configure AWS CLI
+        run: |
+          aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws configure set aws_secret_access_key ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws configure set region ${{ vars.AWS_DEFAULT_REGION }}
+
+      - name: Setup OpenTofu
+        uses: opentofu/setup-opentofu@v1
+        with:
+          tofu_wrapper: false
+
+      - name: Delete Global Accelerators
+        run: ./provision/aws/global-accelerator/accelerator_reaper.sh
 
   rosa-delete:
+    needs: accelerator-delete
     if: github.event_name != 'schedule' || github.repository == 'keycloak/keycloak-benchmark'
     uses: ./.github/workflows/rosa-cluster-delete.yml
     secrets: inherit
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-delete-on-schedule.yml
```diff
@@ -56,26 +77,6 @@ jobs:
...
-          aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws configure set aws_secret_access_key ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws configure set region ${{ vars.AWS_DEFAULT_REGION }}
-
-      - shell: bash
-        run: ./provision/aws/aws_check_vpcs.sh
-
-      - shell: bash
-        if: ${{ always() }}
-        run: ./provision/aws/aws_check_subnet_groups.sh
```
pruivo, line 74: aws_check scripts are not invoked anymore.

.github/workflows/auto-delete-on-schedule.yml
```diff
@@ -56,26 +77,6 @@ jobs:
   delete-check:
     needs: [rosa-delete, aurora-delete, route53-delete]
```
pruivo, line 78: Wouldn't adding `accelerator-delete` fix the issue?

.github/workflows/auto-delete-on-schedule.yml
mhajas: It seems moving up and down of yaml blocks does not work very well in Intellij.

.github/workflows/auto-delete-on-schedule.yml
mhajas: I don't think so. The setup we currently have is the following: `[kc1,kc2,kc3] <- LB-site-1 <- GA -> LB-site2 -> [kc1,kc2,kc3]`. Removal of GA is failing when LBs are not available. My assumption is that the LBs are removed as part of the keycloak deployment removal/ROSA clusters removal. Therefore we need to make sure GA removal is done before ROSA clusters are removed.

pruivo: gotcha. How about Route53? Does it have the same problem?

mhajas: No because we are not using tofu for removing it. Since this: https://github.com/keycloak/keycloak-benchmark/blob/main/provision/opentofu/modules/aws/accelerator/main.tf#L37 is an input for creating GA it needs to be available also when destroying the state even though it should not be necessary for removing it. Maybe we should change this to have more flexible way for removing GAs. Let's see how this fix works and if we will have some unremoved GAs in future we can rethink how we do it.

pruivo (approved):

pruivo:
merged! thanks @mhajas

PR #1089, opened 2025-01-09
Move away from DockerHub for used images

### LINKED ISSUE #1088: Move to non-docker-hub images
Motivation: docker-hub has rate limitations and expensive enterprise plans. 
Let's move to non-docker-hub images. 
 
TODO:
* Move Grafana / Jaeger / PostgreSQL image etc. to either quay or Google cache

ahus1:
Closes #1088


mhajas (approved):
I went through all images we use in Minikube and all seems to be downloaded either from quay or gcr. Thanks @ahus1!

PR #1118, opened 2025-03-07
monitoring upgrades

ryanemerson:
- **Upgrade to kube-prometheus-stack 69.7.3**
- **Upgrade Jaeger to 3.4.0**
- **Upgrade Loki to 6.27.0**
- **Upgrade promtail to 6.16.6**
- **Don't track keycloak-grafana-dashboard**

All appears to be working as expected on my local `minikube` cluster. I wanted to utilise Jaeger v2, but it seems that's not currently possible with the helm chart https://github.com/jaegertracing/helm-charts/issues/610.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,6 @@ provision/environment_data.json
 **/*.terraform*
 !**/*.terraform.lock.hcl
 provision/opentofu/modules/aws/accelerator/builds/*
+
+# Grafana dashboard
+provision/minikube/monitoring/keycloak-grafana-dashboard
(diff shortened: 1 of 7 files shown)


ahus1:
Thank you for this update! I did some more testing, and to make Loki work I needed to apply some of the documented [single replica config](https://grafana.com/docs/loki/latest/setup/install/helm/install-monolithic/#single-replica). 

I then started to modify the services so that linking logs, traces and metrics (via Exemplars) work. It seems that they broken once we moved away from the OpenTelemetry instrumentation to Quarkus' own OpenTelementry. I tested it with my local Minikube and it worked for me.

ryanemerson:
Thanks for the fixes @ahus1. I've added a couple of more commits fixing SQLPad issues. I have everything I need for my demo working here, so I'm fine for this to be merged once you're happy.

PR #1120, opened 2025-03-10
Extend config for custom proxy and username / password patterns

### LINKED ISSUE #1119: Extend configuration options for custom proxy and username / password patterns
### Description

Let's extend the configuration options for a wider range of use-cases.

**Proxy**

Currently there is a config option to enable an http proxy (`--http-proxy`), but only for a static host and port. For environments where it is a prerequisite to use a specific proxy, it would be awesome to be able to configure this with e.g.:

- `http-proxy-host`
- `http-proxy-port`

**Users and passwords**

The current user and password patterns are `user-<index>` and `user-<index>-password`, with an index starting at 0. That does not give us a lot of room for the user base we could use (and I guess we can reach the borders of these patterns by only using the `user-prefix` config of the dataset module when we generate things). For more freedom in the creation and usage of a user base it would be awesome to be able to adjust this by e.g.:

- `username-prefix`
- `user-password-prefix`
- `user-password-suffix`
- `user-index-offset`

Having that it would be possible to not only adjust the patterns of usernames and passwords, but also to use the user base from let's say index 1-100, or 1000-2000 when we have differenty configured users for some cases.

### Discussion

_No response_

### Motivation

We are running keycloak benchmark in an environment that gives us 2 prerequisits.
1. a proxy
2. a pre-provisioned user base

To be able to meet our requirements, we have to extend the available configuration options by the ones mentioned in the description.

### Details

_No response_

dmnkdmnt:
Closes #1119

### DIFF
--- a/benchmark/src/main/java/org/keycloak/benchmark/Config.java
+++ b/benchmark/src/main/java/org/keycloak/benchmark/Config.java
@@ -42,9 +42,29 @@ public class Config {
     public static final String userName = System.getProperty("username");
 
     /**
+     * Sets the prefix for usernames. If not set, usernames use the {@code user-} prefix. Eg.: user-0, user-1.
+     */
+    public static final String userNamePrefix = System.getProperty("username-prefix", "user-");
+
+    /**
+     * Sets the index offset for users and user passwords. If not set, index starts at {@code 0}. Eg.: user-0, user-1.
+     */
+    public static final int userIndexOffset = Integer.getInteger("user-index-offset", 0);
+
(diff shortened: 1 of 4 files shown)


ahus1 (approved):
Thank you for this enhancements and for also updating the docs!

ahus1:
Thank you for the PR. The GitHub actions will build a new snapshot release within the next minutes. If you need this change also for the 26.2 snapshot releases, create a backport with a pull request against that branch with the same changes.

PR #1130, opened 2025-04-03
Provisioning and GHA updates related to Keycloak and Infinispan OLM installation on ROSA OCP - `main`

tkyjovsk:


mhajas (approved):
Thanks @tkyjovsk

PR #1140, opened 2025-05-21
Re-introducing the embedded Keycloak server

pedroigor:


ryanemerson:
Thanks @pedroigor. Can you share an example benchmarking use-case for this?

PR #1159, opened 2025-07-09
Add `clusterPrefix` and `region` params to the multi-az tests workflow

tkyjovsk:
- Added `clusterPrefix` and `region` params to the multi-az tests workflow. This will allow launching the `-run-tests` workflow independently on the scheduled job, and possibly against a different set of clusters.
- Renamed the `rosa-multi-az-clusters-create` workflow to reflect the fact that it also deploys keycloak afterwards.

Note that the `region` is not being passed to all sub-flows, only to `rosa-multi-az-clusters-create-and-deploy-keycloak` and `rosa-scaling-benchmark` because these two need to create stuff in a specific region.

The remaining subflows use the default region (only to initiate the `rosa` CLI tool in order to get details about clusters that already exist).


ryanemerson:
Thanks @tkyjovsk

PR #1161, opened 2025-07-15
Deploy keycloak in a stretch cluster

### LINKED ISSUE #1160: Deploy keycloak in a stretch cluster
### Description

Add options to deploy Keycloak in a multi-az Kubernetes cluster (also known as stretch cluster)

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #1160

Draft PR for now, as I haven't tested all the variables. 

Environment variables example for stretched clusters:
```
AVAILABILITY_ZONES=eu-west-1a,eu-west-1b,eu-west-1c
REGION=eu-west-1
REPLICAS=6 # must be a multiple of the number of availability zones
USE_SINGLE_AZ=false
```

### DIFF
--- a/provision/minikube/keycloak/templates/keycloak.yaml
+++ b/provision/minikube/keycloak/templates/keycloak.yaml
@@ -10,6 +10,31 @@ metadata:
   namespace: {{ .Values.namespace }}
 spec:
 # end::keycloak-ispn[]
+  {{- if .Values.stretchCluster }}
+  # evenly spread pods between zones and nodes.
+  scheduling:
+    affinity:
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app: keycloak
+                  app.kubernetes.io/managed-by: keycloak-operator
+                  app.kubernetes.io/component: server
+                  app.kubernetes.io/instance: keycloak
+              topologyKey: topology.kubernetes.io/zone
+          - weight: 90
+            podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app: keycloak
+                  app.kubernetes.io/managed-by: keycloak-operator
+                  app.kubernetes.io/component: server
+                  app.kubernetes.io/instance: keycloak
+              topologyKey: kubernetes.io/hostname
+  {{- end}}
   hostname:
     {{- if .Values.keycloakHostname }}
     hostname: {{ .Values.keycloakHostname }}
--- a/provision/minikube/keycloak/values.yaml
+++ b/provision/minikube/keycloak/values.yaml
@@ -50,3 +50,7 @@ infinispan:
     port: 11222
     username: developer
 tracingEndpoint: http://jaeger-collector.monitoring.svc:4317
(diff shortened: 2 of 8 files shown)


provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -10,6 +10,19 @@ metadata:
...
 # end::keycloak-ispn[]
+  {{- if .Values.stretchCluster }}
+  # evenly spread pods between zones and nodes.
+  scheduling:
+    affinity:
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 50
+            podAffinityTerm:
+              topologyKey: topology.kubernetes.io/zone
```
pruivo, line 21: Should I add label selectors to match the Keycloak pods? It doesn't seem to be mandatory, but it sounds like a best practice.

provision/minikube/keycloak/values.yaml
```diff
@@ -50,3 +50,5 @@ infinispan:
     port: 11222
     username: developer
 tracingEndpoint: http://jaeger-collector.monitoring.svc:4317
+# Stretched cluster?
```
ryanemerson, line 53: :+1:

provision/minikube/keycloak/templates/keycloak.yaml
ryanemerson: I think so, otherwise I believe the policy will also take into account non-keycloak pods which is not desirable.

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -10,6 +10,19 @@ metadata:
   namespace: {{ .Values.namespace }}
 spec:
 # end::keycloak-ispn[]
+  {{- if .Values.stretchCluster }}
+  # evenly spread pods between zones and nodes.
+  scheduling:
+    affinity:
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 50
```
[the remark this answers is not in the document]
ryanemerson: The weights combined don't need to equal 100, instead we want to ensure that different zones takes priority, and when that's not possible we prioritize different nodes. In the Infinispan operator docs we recommend users have a weight of 100 for the zone and 90 for the hostname so that a distinct zone takes preference over a distinct node.

Gemini summary of the behaviour:

- Filtering Phase: All nodes are initially considered viable candidates.

- Scoring Phase: The scheduler scores each node based on the weight assigned to the preferred anti-affinity rule. For each node that has a pod matching the anti-affinity rule, its score is reduced by the specified weight (a value between 1 and 100).

    Nodes that do not have a matching pod will have a score of 0.

    Nodes that do have a matching pod will receive a negative score (e.g., -50 if the weight is 50).

- Final Selection: The scheduler then selects the node with the highest score. In this scenario, a node with a score of 0 will always be chosen over a node with a negative score. If all nodes have a matching pod and thus all have the same negative score, the scheduler will choose one of them. The pod will still be scheduled, but on the least "undesirable" option according to the preferences.

ryanemerson (changes requested):

pruivo:
Fixed!

PR #1183, opened 2025-07-31
Remove Aurora global database scripts

### LINKED ISSUE #1167: Remove Aurora global database scripts from repository
### Description

The Aurora global database scripts should be removed from the repository as we don't use them any more, they are no longer regularly tested, and they add clutter to the taskfiles.

### Discussion

_No response_

### Motivation

With a lot content that is not well maintained, maintenance costs increase. 

The data is still available in the old tags if we should need them. 

### Details

@pruivo  / @ryanemerson - please comment on this idea. Thanks!

pruivo:
Closes #1167


ryanemerson:
Thanks @pruivo

Issue #996, opened 2024-10-02
dataset-import.sh isn't included in the binary release of keycloak-benchmark


paul-civitas:
### Describe the bug

keycloak-benchmark-0.12-SNAPSHOT.tar.gz includes the following:

```
keycloak-benchmark-0.12-SNAPSHOT/bin/kcb.sh
keycloak-benchmark-0.12-SNAPSHOT/bin/kc-failover.sh
keycloak-benchmark-0.12-SNAPSHOT/bin/initialize-benchmark-entities.sh
keycloak-benchmark-0.12-SNAPSHOT/bin/kc-chaos.sh
keycloak-benchmark-0.12-SNAPSHOT/conf/README.adoc
keycloak-benchmark-0.12-SNAPSHOT/lib/keycloak-benchmark-0.12-SNAPSHOT.jar
```

But it doesn't include: dataset-import.sh which is a crucial piece of the puzzle. To get that script to run it you need to download the src.

### Version

Keycloak 24, keycloak-benchmark 0.12-SNAPSHOT

### Expected behavior

Scripts and binaries expected to be run by the user, are included with the binary distribution.

### Actual behavior

Only some portion of the scripts are included.

### How to Reproduce?

Try and run benchmarks without cloning down the src.

### Anything else?

This is a minor issue

mhajas:
Thank you for the report @paul-civitas. You are right, the script is missing from the release. The simplest fix would be to the same as we are doing in `benchmark` module and package a zip file also for `dataset` module that would contain both, the shell script and the jar file.

PR #1194, opened 2025-08-22
dataset: Include dataset-import.sh in assembly

TobyTheHutt:
Adding the assembly plugin to the dataset module to pack the `dataset-import.sh` script.

It's my first commit in this repo. Feel free to roast me if I forgot or oversaw anything.

PR Relates to #996


ryanemerson:
Thanks for the PR @TobyTheHutt

Your commit results in `dataset/target/keycloak-benchmark-dataset-999.0.0-SNAPSHOT.tar.gz` being created with the dataset jar and script, however this still isn't packaged as part of `benchmark/target/keycloak-benchmark-999.0.0-SNAPSHOT.tar.gz ` which is what we release. Looking at the original issue, this seems to be desired.

I have pushed a commit to change the behaviour so the jar and script are included in `keycloak-benchmark-999.0.0-SNAPSHOT.tar.gz ` and `keycloak-benchmark-999.0.0-SNAPSHOT.jar`. Let me know what you think.

TobyTheHutt:
Thank you for your commit @ryanemerson ! I tested it locally; It looks good to me and makes sense, in that the collection of scripts is complete that way.

Just so we align in our understanding:  
The initial request certainly requested the script to be contained in the release package, as you say. However, in the comment of @mhajas , I interpreted that a second package would be the fix:

> The simplest fix would be to the same as we are doing in benchmark module and package a zip file also for dataset module that would contain both, the shell script and the jar file.

Did I misinterpret, or set the wrong focus in my approach? Thanks for your feedback.  
In any case, I'm content with both solutions and trust your seniority in this.

ryanemerson:
> Did I misinterpret, or set the wrong focus in my approach? Thanks for your feedback.

No I think you understood Michal's suggestion correctly :+1:

We can go down that route, but then we also need to make changes to our Git Hub release actions and users will need to download multiple `.zip|.tar.gz` files. IMO releasing a single archive is simpler for us to maintain and easier for users to consume.

I'm happy to hear opinions to the contrary though :slightly_smiling_face:

TobyTheHutt:
Seems like good thinking to me. Thanks for the clarification!
To me it seems, we're good to go.

PR #1199, opened 2025-09-03
Aurora PostgresSQL 17.5

ryanemerson:
Closes keycloak/keycloak-benchmark#1200

Scheduled run on my fork for multi-site: https://github.com/ryanemerson/keycloak-benchmark/actions/runs/17427780394

I didn't run the scheduled workflow on my fork as the cluster name defaults to `gh-ryanemerson-az` which is too long and fails. Currently there's no way AFAICT to modify this default with the scheduled run. Example failed run: https://github.com/ryanemerson/keycloak-benchmark/actions/runs/17427232221/job/49477049875


ahus1:
@ryanemerson - a workaround might be creating the feature branch in the main repository which comes which the usual caveats or cleaning it up afterwards, etc. 

If you are ok to babysit this change after it is merged, I'm ok if you go ahead and merge it to the repository.

PR #1213, opened 2025-10-01
Secure Aurora PosgreSQL connection

### LINKED ISSUE #1208: Secure Aurora PosgreSQL connection
### Description

Official docs: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.Security.html

We deploy Aurora with TLS support enabled, as shown below.

Image

We are missing the configuration in the Keycloak image. By default:
```
The client first tries to connect to the database with SSL/TLS by default. If the client can't connect with SSL/TLS, it reverts to connecting without SSL/TLS
```

We have to configure `sslrootcert` and `sslmode=verify-ca|verify-full`

The CA certificate can be found on this page: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL.html

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #1208


provision/minikube/keycloak/certs/eu-west-1-bundle.pem
```diff
@@ -0,0 +1,76 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
ryanemerson: No, but I figured we would probably want a unified approach.

provision/minikube/keycloak/templates/keycloak.yaml
```diff
@@ -55,7 +55,7 @@ spec:
     vendor: mysql
 {{ end }}
 {{- if .Values.useAWSJDBCWrapper}}
-    url: jdbc:aws-wrapper:postgresql://{{ .Values.dbUrl }}:5432/keycloak
+    url: jdbc:aws-wrapper:postgresql://{{ .Values.dbUrl }}:5432/keycloak?sslmode=verify-full&sslrootcert=/opt/keycloak/certs/aurora.pem
```
ahus1, line 58: When we do it this way, I expect it would also show up in the docs eventually. In order for that to work, we would need to have a `podTemplate` configured, which is unsupported. 

This is then a block this issue, and it is soon do be deprecated - see [#40183](https://github.com/keycloak/keycloak/issues/40183)

There is an alternative way to configure this in https://github.com/keycloak/keycloak/issues/40184#issuecomment-3306911756 that is using `truststores` to pass in the name of the secret. 

With that change, I assume the `slrootcert=...` should no longer be necessary, and only `sslmode=verify-full` would be needed.

ahus1 (changes requested):
Thank you for the PR. See below for a required change. 

In addition to that, can we configure our AuroraDB so that it only accepts TLS connections, and rejects a non-TLS connection?

provision/minikube/keycloak/templates/keycloak.yaml
ahus1: Having said that, baking it in the image is not ideal, but for those (long lived) certificates, it might be an option. We are already adding the JDBC driver, and we do that in the same step.

pruivo:
@ahus1 @ryanemerson I've updated and removed the changes from the `unsupported` section. Unfortunately, I still have to use `sslrootcert` in the URL.

ahus1 (approved):

ahus1:
@pruivo - thank you for the change!

As a follow-up: 

* Verify and potentialls pdate our scripts in KCB to enfore TLS connections - since PostgreSQL 17 this should be the case, but I am surprised that it worked before https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.Security.html with option rds.force_ssl 
* Update https://github.com/keycloak/keycloak/issues/40185 with your solution, and also that it didn't work by just adding the truststore. We should make the CND team aware that we with that documentation, we would depend on the trust store configmaps to appear always in the same place, and a change of place would be a breaking change for us.
* Update https://www.keycloak.org/operator/advanced-configuration with the solution on how to add a PostgreSQL certificate to the database when using the Operator
* Update the for the single and multi-cluster docs: https://www.keycloak.org/high-availability/single-cluster/deploy-keycloak

Thanks!

PR #1221, opened 2025-10-16
Fix terraform AWS cross-region secret handling.

### LINKED ISSUE #1220: ROSA provisioning fails in regions that are not eu-central-1
### Describe the bug

The provisioning fails with: 

```
│ Error: reading Secrets Manager Secret Version (arn:aws:secretsmanager:eu-central-1:606671647913:secret:keycloak-master-password-9eDdyX|AWSCURRENT): couldn't find resource
│ 
│   with data.aws_secretsmanager_secret_version.current,
│   on main.tf line 9, in data "aws_secretsmanager_secret_version" "current":
│    9: data "aws_secretsmanager_secret_version" "current" {
```

See https://github.com/keycloak/keycloak-benchmark/actions/runs/18522823053/job/52786503236

### Version

main

### Expected behavior

It should succeed.

### Actual behavior

It fails with the error message shown above.

### How to Reproduce?

Run the GitHub action

### Anything else?

It seems that an AWS provider in Terraform can only work with a single region, see https://github.com/hashicorp/terraform-provider-aws/issues/7775

Maybe the admin password can be passed in as a variable instead.

tkyjovsk:
After investigating a terraform-native solution (which involves importing the password resource across regions via additional aws provider using the secret's ARN as a input param), I ended up passing the password to tofu from the env variable.

Closes #1220

### DIFF
--- a/.github/workflows/rosa-cluster-create.yml
+++ b/.github/workflows/rosa-cluster-create.yml
@@ -86,6 +86,7 @@ jobs:
           AVAILABILITY_ZONES: ${{ inputs.availabilityZones }}
           CIDR: ${{ inputs.cidr }}
           CLUSTER_NAME: ${{ inputs.clusterName || format('gh-{0}', github.repository_owner) }}
+          CLUSTER_ADMIN_PASSWORD: ${{ env.KEYCLOAK_ADMIN_PASSWORD }}
           COMPUTE_MACHINE_TYPE: ${{ inputs.computeMachineType }}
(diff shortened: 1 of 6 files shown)


tkyjovsk:
Successfull ROSA creation: https://github.com/tkyjovsk/keycloak-benchmark/actions/runs/18535906815
Successful ROSA deletiion: https://github.com/tkyjovsk/keycloak-benchmark/actions/runs/18566364528/

PR #1223, opened 2025-10-21
Fix tofu `destroy.sh` script.

tkyjovsk:
- Remove rosa-specific `CLUSTER_ADMIN_PASSWORD` handling from the script, and dependent GHA workflow.
- Add a dummy password value to `rosa-hcp/output.tf`.


ryanemerson:
Thanks @tkyjovsk

PR #1238, opened 2025-11-26
Allow the dataset to generate users with a limited number of unique passwords

### LINKED ISSUE #1237: Allow the dataset to generate users with a limited number of unique passwords
### Description

Limiting the number of unique passwords to a configurable upper bound should significantly reduce dataset creation time when a large number of users are required, as we restrict the number of hash calculations to the upper bound.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #1237

I have kept it simple for now and only reused credentials per `create-users` invocation. The reason for this is that the `unique-credential-count` ultimately determines the password, so storing and reloading the credentials in the DB means that we then need to handle `unique-credential-count` changing.

Passwords can be calcuated using: `password-$(i % unique-credential-count)` where `i` is the user number.

Quick local benchmark:

4s `./dataset/dataset-import.sh -a create-users -u 10000 -U 100 -l http://localhost:8080/realms/master/dataset`

vs

45s  `./dataset/dataset-import.sh -a create-users -u 10000 -l http://localhost:8080/realms/master/dataset`

I haven't added any docs yet, as I wasn't sure if this was functionality we want to advertise or are happy keeping it hidden for our own needs. Let me know if we require docs.

### DIFF
--- a/doc/dataset/modules/ROOT/pages/using-provider.adoc
+++ b/doc/dataset/modules/ROOT/pages/using-provider.adoc
@@ -68,7 +68,7 @@ To learn more about the tool, see xref:kubernetes-guide::util/task.adoc[] for de
 
 You need to call this HTTP REST requests.
 This request is useful for create 10 realms.
-Each realm will contain specified amount of roles, clients, groups and users:
+Each realm will contain a set of roles, clients, groups and users:
 
 ----
 .../realms/master/dataset/create-realms?count=10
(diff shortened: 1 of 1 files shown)


dataset/src/main/java/org/keycloak/benchmark/dataset/DatasetResourceProvider.java
```diff
@@ -457,6 +462,17 @@ private void addUserCreationTasks(RealmContext context, Task task, DatasetConfig
         }
     }
 
+    private List<PasswordCredentialModel> initializeCredentials(DatasetConfig config, RealmContext context) {
+        PasswordPolicy policy = context.getRealm().getPasswordPolicy();
+        PasswordHashProvider hash = policy.getHashAlgorithm() != null ?
+              baseSession.getProvider(PasswordHashProvider.class, policy.getHashAlgorithm()) :
+              baseSession.getProvider(PasswordHashProvider.class);
+
+        return IntStream.range(0, config.getUniqueCredentialCount())
```
pruivo: Probably should compute the minimum between the number of users and the number of credentials. 
```suggestion
        return IntStream.range(0, Math.min(config.getUsersPerRealm(), config.getUniqueCredentialCount()))
```

pruivo (approved):
LGTM. Tested the `create-realm` and benchmark run with 1 unique password, and it worked. 
I'll let @ahus1 review the docs :)

PR #1264, opened 2026-01-28
Replace Aurora in stretched cluster setup with CNPG.

tkyjovsk:
Closes https://github.com/keycloak/keycloak/issues/45798

### DIFF
--- a/.github/workflows/rosa-stretched-cluster.yml
+++ b/.github/workflows/rosa-stretched-cluster.yml
@@ -7,10 +7,12 @@ on:
         description: 'Check to Create Cluster.'
         type: boolean
         default: true
-      createAurora:
-        description: 'Check to Create Aurora Database.'
-        type: boolean
-        default: true
+      database:
+        description: 'Chose a database'
+        type: choice
+        options:
+          - cnpg
+          - aurora-postgres
       createKeycloak:
         description: 'Check to Create Keycloak.'
         type: boolean
@@ -31,10 +33,10 @@ on:
         description: 'Check to Create Cluster.'
         type: boolean
         default: true
-      createAurora:
-        description: 'Check to Create Aurora Database.'
-        type: boolean
-        default: true
+      database:
+        description: 'Database. Supported values: cnpg, aurora-postgres'
+        type: string
+        default: cnpg
       createKeycloak:
         description: 'Check to Create Keycloak.'
         type: boolean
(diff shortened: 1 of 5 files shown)


tkyjovsk:
The second benchmark run with the full dataset brought down the CNPG writer instance during dataset creation, possibly due to low disk space. I'll have to look into the database configs a bit.
```
Defaulted container "postgres" out of: postgres, bootstrap-controller (init)
{"level":"info","ts":"2026-01-28T17:50:33.255720195Z","msg":"OS distribution is supported","logger":"instance-manager","logging_pod":"cnpg-keycloak-1","entry":{"version":"13 (trixie)","deprecatedFrom":"2028-08-09T00:00:00Z","supportedUntil":"2030-06-30T00:00:00Z"}}
{"level":"info","ts":"2026-01-28T17:50:33.255822237Z","msg":"Starting CloudNativePG Instance Manager","logger":"instance-manager","logging_pod":"cnpg-keycloak-1","version":"1.28.0","build":{"Version":"1.28.0","Commit":"a9696201f","Date":"2025-12-09"},"skipNameValidation":false}
{"level":"info","ts":"2026-01-28T17:50:33.255836222Z","msg":"Checking for free disk space for WALs before starting PostgreSQL","logger":"instance-manager","logging_pod":"cnpg-keycloak-1"}
{"level":"info","ts":"2026-01-28T17:50:33.26196918Z","msg":"Detected low-disk space condition, avoid starting the instance","logger":"instance-manager","logging_pod":"cnpg-keycloak-1"}
...
```

tkyjovsk:
I increased the CloudNativePG storage to 4Gi and ran another benchmark with full dataset and load (same as for the stretched cluster setup with Aurora) [and it has PASSED](https://github.com/tkyjovsk/keycloak-benchmark/actions/runs/21485594678).

I noticed that the connection pool on the writer pod are at 93% of capacity so I'm adding an option to customize this value. 

Currently the CNPG pods don't have any CPU or memory limits set which means they are allowed to use as much as there is available on the worker nodes. We may want to consider setting some limits here.

Also it seems that even though the CNPG operator schedules the pods on different nodes it may still happen that there are 2 pods in the same AZ. I'll investigate setting affinity rules.

tkyjovsk:
The 1M-user dataset unfortunatelly again brought down the writer instance even with storage increased to 4GB. I'll have to look again at storage and WAL-archiving configuration and probably increase again.

ryanemerson:
> Latency seems to be slightly better with CNPG.

Looking at the breakdown of the benchmark it seems like the individual testing steps are actually slower for CNPG, but only by a couple of seconds.

I noticed that the dataset creation step is skipped on the CNPG run, how long does that typically take compared to Aurora?

PR #1265, opened 2026-01-29
Add `.env` and `.task` to global `.gitignore` file.

tkyjovsk:
Move `.env` and `.task` to the global `.gitignore` file and remove redunadnt entries from subfolders.


tkyjovsk:
Removed redundant entries from subfolders.

